PR #67, opened 2014-10-17
Feature54 shortcut newversion

yodalee:
This is the new custom shortcut setting I write. It's a modification of the previous version, which is now in feature54shortcut branch. However, there are many differences between these two, so I suggest simple delete the previous branch.
The shortcut setting dialog is added in Application Setting->setting->"custom shortcut".

*****

Here I will explain what I did in each commits:

1. I use the object keysequenceedit implemented in previous version.
It can generate QString represent the keysequence like hit Ctrl+Alt+T will show Ctrl+Alt+T directly.

2. modify layout in qucssettingdialog

3-5. Implement default shortcut setting, load/save.
Here add a new container in QucsSettings, which record all shortcut key sequence.
It's a Vector of map<QString, QString>
The map inside is directly map QAction name to its shortcut key sequence.
Different QActions that stores in different menu will be stored in different map.

6-7. Implement the layout of qucsshortcutdialog, and fill them with data in shortcut 

8-9. Implement the set/remove/default button function.
Every time you input a key sequence, the dialog will check is there a conflict to other shortcut setting or not.
It you press "setShortcut" when there is a conflict, the conflict one will be wipe out to no shortcut.

10. Implement shortcut setting in qucs.
Collect setShortcut of each QAction into one function.

11. Implement invalidKey.
I notice that some key sequence like "ESC" should be preserved to specific action. So the shortcut dialog will store a list of invalid Key sequence, every input will be checked.

******

I think this version is a perfect version to me. Hope everyone will like my implementation.
Look forward to your comments.

### DIFF
--- a/qucs/qucs/dialogs/CMakeLists.txt
+++ b/qucs/qucs/dialogs/CMakeLists.txt
@@ -13,7 +13,7 @@ labeldialog.cpp			searchdialog.cpp vtabbutton.cpp
 librarydialog.cpp		settingsdialog.cpp vtabwidget.cpp
 matchdialog.cpp			simmessage.cpp newprojdialog.cpp
 sweepdialog.cpp			exportdialog.cpp loaddialog.cpp
-keysequenceedit.cpp
+qucsshortcutdialog.cpp keysequenceedit.cpp
 )
 
 SET(DIALOGS_MOC_HDRS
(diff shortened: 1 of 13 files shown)


yodalee:
I will send new version.

PR #86, opened 2014-11-01
Feature search component

yodalee:
I finally got the solution to search function. The solution is adding a search category, and the search text input only show when user select this category.
When user drag component into schematic, qucs will search all the components to find which component is right. I have to say this solution is not very good, quite waste resource.

### DIFF
--- a/qucs/qucs/qucs.cpp
+++ b/qucs/qucs/qucs.cpp
@@ -323,13 +323,13 @@ void QucsApp::initView()
   CompComps = new QListWidget(this);
   CompComps->setViewMode(QListView::IconMode);
   CompComps->setGridSize(QSize(110,90));
-  CompSearchLabel = new QLabel(this, tr("Search Components:"));
   CompSearch = new QLineEdit(this);
+  CompSearch->setPlaceholderText(tr("Search Components"));
   CompSearchClear = new QPushButton(tr("Clear"));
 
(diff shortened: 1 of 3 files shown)


yodalee:
I just rebase this to most current (though no conflict actually)
@guitorri can you help me review this code? I have tested it and no problem encounter.

yodalee:
Rebase done, I modify the layout excep -20 spacing.
`addSpacing(-20);`
since it makes qucs looks like this:
desktop2

qucs/qucs/qucs.cpp
```diff
@@ -317,14 +317,22 @@ void QucsApp::initView()
...
   QWidget *CompGroup  = new QWidget();
   QVBoxLayout *CompGroupLayout = new QVBoxLayout();
+  QHBoxLayout *CompSearchLayout = new QHBoxLayout();
 
   CompChoose = new QComboBox(this);
   CompComps = new QListWidget(this);
   CompComps->setViewMode(QListView::IconMode);
   CompComps->setGridSize(QSize(110,90));
+  CompSearch = new QLineEdit(this);
+  CompSearch->setPlaceholderText(tr("Search Components:"));
```
guitorri: No ':' needed

yodalee: fix

guitorri:
Merged in 6688617. Thank you!

PR #145, opened 2014-12-03
ElementCache

global667:
* Made ElementCache Qt4 conform. Exchange Q3PtrList to QList and QMutableListIterator.
* Was a bad boy: Add mutualx.h to cmake file in this PR.


guitorri:
Should be fixed in #241

Issue #118, opened 2014-11-14
improve team work on corporate projects


scordilis:
need to have a lock mechanism on the project when opening a project to prevent some modification by others ... I've wrote a small tool for that, but where to place it ?

Thierry.

PR #153, opened 2014-12-22
Project manager

yodalee:
I am back haha, recently having work on my master thesis.
This PR and branch are gonna clean up qucs main program, there will be some further work on this branch. However, I think that too large PR will be more difficult to review and merge.
The PR now is refactoring project viewer at dock window with listview and QFileSystemModel. The model is always sync with file system state, so we don't need call "refresh" every time.
I also examine the code in qucs.cpp, and giving it some clean up.

### DIFF
--- a/qucs/qucs/components/subcircuit.cpp
+++ b/qucs/qucs/components/subcircuit.cpp
@@ -93,7 +93,7 @@ void Subcircuit::createSymbol()
     }
   }
   else {
-    No = QucsApp::testFile(FileName);
+    No = Schematic::testFile(FileName);
     if(No < 0)  No = 0;
 
     Ports.clear();
--- a/qucs/qucs/qucs.cpp
+++ b/qucs/qucs/qucs.cpp
@@ -1104,14 +1104,21 @@ void QucsApp::slotProjNewButt()
   NewProjDialog *d = new NewProjDialog(this);
   if(d->exec() != QDialog::Accepted) return;
 
-  //QDir projDir(QucsSettings.QucsHomeDir.path());
-  //if(projDir.mkdir(d->ProjName->text()+"_prj")) {
-  //  Projects->insertItem(0, d->ProjName->text());  // at first position
-  //  if(d->OpenProj->isChecked())
-  //    slotOpenProject(Projects->item(0));
-  //}
-  //else QMessageBox::information(this, tr("Info"),
-  //                  tr("Cannot create project directory !"));
+  QDir projDir(QucsSettings.QucsHomeDir.path());
+  QString name = d->ProjName->text();
+  bool open = d->OpenProj->isChecked();
+
+  if (!name.endsWith("_prj")) {
+    name += "_prj";
+  }
+
+  if(!projDir.mkdir(name)) {
+    QMessageBox::information(this, tr("Info"),
+        tr("Cannot create project directory !"));
+  }
+  if(open) {
+    //slotOpenProject(Projects->item(0));
+  }
 }
 
 // ----------------------------------------------------------
--- a/qucs/qucs/qucs.cpp
+++ b/qucs/qucs/qucs.cpp
@@ -1279,8 +1279,10 @@ void QucsApp::openProject(const QString& Path)
   editText->setHidden(true); // disable text edit of component property
 
   QString Name = Path;
-  Name = Name.left(Name.length()-1);   // cut off trailing '/'
-  int i = Name.findRev('/');
+  if (Name.endsWith(QDir::separator())) {
+    Name = Name.left(Name.length()-1);  // cut off trailing '/'
+  }
+  int i = Name.lastIndexOf(QDir::separator());
   if(i > 0) Name = Name.mid(i+1);   // cut out the last subdirectory
   Name.remove("_prj");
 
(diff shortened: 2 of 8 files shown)


guitorri:
This is very nice! Can you do a quick rebase to resolve the conflict?

I love that it updates as the file system changes.

Should we hide the `_prj` suffix?

For the record, this is the before an after on OSX Mavericks:

snap 2015-01-05 at 18 50 46
snap 2015-01-05 at 18 50 23

yodalee:
Done, currently run CI test.
I suggest that we keep the prj suffix, since it's easy for user to remember the what the folder is.
However, in the other place we may call the project with its full name. For example the `Content` tab in the dock window, the header is "Content of `projectname`". The `projectname` has no `_prj` suffix. This takes additional effort to process the project name. We can get rid of it, and simply display it "Content of `projectname_prj`"

guitorri:
You are right and it is nice the way it is now.
We either refer to the  _project name_ `ProjectABC` or  _project directory_ `ProjectABC_prj`. 
Just taking another look before merging.

yodalee:
This is just a beginning implementation, there is 10 commits after this. I just PR this because send too many commits will make review work more difficult, and actually I have to prepare my master thesis so no time for qucs recently :P.

guitorri:
It is definitely better to keep PR to a manageable size. This looks good to me. Merging!
Good luck with your thesis.

in3otd:
Hello,
maybe I am missing something, but there a few things which are not clear to me. Perhaps these will be taken care of in the next commits, but I didn't see a description of the planned changes in the functionality from the users point of view, if any.:
- why in the `Projects` tab I see other directories other than the ones ending in `_prj` ?
- what's the purpose of showing the `_prj` suffix? Will the `Projects` tab become a sort of file manager and show also the projects directories content and/or other files/directories ?? Otherwise this suffix will *always* be there and will not add any more information. Similarly, do we plan to show the 'folder' icon for every project? Again, if the plan is to continue to show only the projects directories, this is redundant and just taking up some space. In general I am quite puzzled when I see changes in a GUI which are not due to a change/addition of features offered to the user. 
- Could the usage of `QFileSystemModel` support also the multiple projects directories mentioned in #118 at some time in the future ? Multiple projects directories are useful also for a single user, when projects start to become too many to fit in a single page list... A multiple projects directories view like the one we have in the `Libraries` tab could be nice to have.

...and thanks for all the nice work! :wink:

yodalee:
@in3otd Here is my response:
* this is.... errr, my fault. I'm curious about this too. There is actually code to deal with this:
https://github.com/Qucs/qucs/blob/master/qucs/qucs/qucs.cpp#L397
I will look into this. 
* My thought is only show the project directory only. The modification of this PR is split out a file system model, it will track project directory automatically. We don't need to fetch directory by ourselves, calling the update function by ourselves. Though this way lose some elastic such that `_prj` become more difficult to remove, but I think the advantage is greater then the disadvantage.
* I still cannot get the point of issue 118 wanna to achieve, and I cannot get the meaning of "multiple project view", either. Is that open multiple project in the same time? otherwise there is no difference with current project viewer.

yodalee:
OK I see the problem, in filter_enum I use QDir::AllDirs which cause directory not passing name filter.
It is fixed in #158

in3otd:
Thanks for the explanations, yodalee; I was wearing the (picky) user hat, which thinks "why do I see now the project tab differently if it has the same functionality as before". He is of course not aware of all the work to rationalize and improve the code behind, and should not (necessarily) be.
For the time being, I am fine with showing the project directory only (same GUI behavior as before); but sometimes I move back and forth between projects to copy/check some details, so having a tree view of the projects contents might be helpful to open faster a schematic from another project, but this is just a nice to have feature for the next long winter days :smile: 
Regarding issue 118, as I understand it, it is about having several `QucsHomeDir`, containing each several projects. Imagine for example to have several directories, e.g. `AnalogProjects`, `WorkProjects`, `InProgressStuff`, etc., possibly not having a common parent, each containing different `Qucs` projects. A tree view could show each main directory and list the projects contained in each, more or less in the same way we show the different libraries and their content in the `Libraries` tab.
Then issue 118 raised the need of having a lock mechanism; if the projects collection directories are on a shared disk, potentially different users could try to work on the same project at the same time.

guitorri:
Issue 118 is about one shared `QucsHomeDir` among multiple users. Shared in a network drive maybe. If a user opens one project a lock should be created in order to let other users know that the project is being modified. It can be as simple as a 'lock' file checked every time a project is open.

About the _multiple project view_.  It could be more like having a hierarchy of projects. Projects become sort of  libraries that can be combined to be used in another project.
Some  tools allow you create a list of include paths. Doing so you can for instance build your `audioamp_prj` based on other projects like `op_amp_prj`, `bandfilter_prj`without moving stuff around. 

@crobarcro created the 'Subcircuit Search Path' into Qucs Properties > Locations. It sort of handles the situation above. I haven't used it in a while not sure how it works anymore.

in3otd:
If I recall correctly, the `qucsMgr` @scordilis wrote allowed to choose between several directories, the one chosen was linked to `.qucs', becoming the new working dir (with lock file inside).
I saw there is some code for the 'Subcircuit Search Path' in the sources but I did not understand what was its purpose; is there any doc somewhere?
I think there should be a separation between a project and a library (in the sense of a common repository of user defined (sub)circuits. Maybe the  'Subcircuit Search Path' addresses this already?
Just two level of project directories ("project collections" directories plus actual projects directories) "ought to be enough for everybody" :smile:

crobarcro:
The subcircuit search path works a bit like a compiler search path, so subcircuits on the path can be found just by their name, rather than requiring the full file path. It's very similar in concept to the PYTHONPATH, or the Matlab path of Octave path if you're familar with these.

I will write up some info about it on the wiki to explain fully its usage.

scordilis:
for the qucsMgr, the purpose is really different and complimentary to what has been described herein.

My concerns was : how ensure the fact that working as a team on the same design is protected in some ways ( using the same concept as what we do have with Mentor or Cadence, then the versioning is here for other purposes ... a link to a versioning tool could also be nice, but will need a graphical tool to check the differences ) ==> this is the role of QucsMgr. 

The reason why it is a separate tool, is due to the fact that it is independent of the qucs version ( could be retrofitted also) and the fact that we are using a qucs version 15/debian squeeze on the machine in production (the migration to xubuntu is planned for 2015.

Then the search path is a very good option to share common librairies ( company dependant, debugged and cleaned ). it could help also to point to such or such version of the libs ... which I think is great.

I have always push for a more professional environment for qucs ( like the template, etc ... )

in any case, the search path is a good thing ... it could be good also to be able to configure for absolute path, relative path, search path in the preference file by default ...

crobarcro:
Yes, I think it's sort of separate from what you are describing, note 
also that it currently doesn't work for library files (just individual 
.sch files and .sp, spice files). That would be a useful addition.

in3otd:
thanks for the detailed description @crobarcro , I think it will be good to add it to the 'Working with Subcircuits' user document, if that is ok for you

guitorri:
Speaking about documentation, please take a look fist at #179.

PR #157, opened 2015-01-05
Tune coverage

guitorri:
A few changes to ease the use of gcov on OSX.

There is more things to come, please don't merge it yet.


guitorri:
Bastien handled/merged these already.

PR #173, opened 2015-01-10
Clear qucs app

yodalee:
This is a small (compare to work to do) clean up to qucs.cpp.
I removed some long comment out lines, do some deletion of unnecessary function.


yodalee:
Funny, error in Qucs-doc.

PR #174, opened 2015-01-11
Refactor update recent file function

yodalee:
Refactor open recent file function in qucsApp. Make theses function more clear.
actually mostly just copy and paste from "programming with qt4" handbook.


ra3xdh:
New code looks more compact. What was wrong with old QSignalMapper?

yodalee:
There is a problem using signal mapper. Every time you call the updateRecentFiles, you create a new list of QAction to store the QAction. Though they will be deleted with QObject link, but during Qucs running, these memory never released. Change to static object avoid this problem
The name of recent file is now store in QAction's data field, so we don't need signal mapper, which is more complex than data field.

guitorri:
Functionality seems unchanged. Please merge.

PR #175, opened 2015-01-11
Fixed path to start external tools, added QucsApp::launchTool() function

in3otd:
I had installed Qucs in a local directory, in parallel with the main system-wide install, and noticed that the external tools launched (`qucsfilter`, `qucstrans`, etc.)  were always the ones from the system wide installation.
Looking at the code I think there is a copy-and-paste error, in that almost all the tools were launched with code like

```` C++
QucsFilter->setWorkingDirectory(QucsSettings.BinDir);
QucsFilter->start(prog);
prog = QDir::toNativeSeparators(QucsSettings.BinDir+prog);
qDebug() << "Command :" << prog;
````

so the actual tool launched was the one in the environment path, but the debug message printed contained the right path !!
I took the opportunity to clean up this part of the code, eliminating the copy/paste and adding a new `launchTool()` function.
Note that the code could probably be simplified even more, also in `qucs_init.cpp`, by having a Tool class handling the available external tools registration, where each registered tool will provide its executable name and its description to use in menus and messages, so the Tools menu built starting at https://github.com/Qucs/qucs/blob/master/qucs/qucs/qucs_init.cpp#L525 and https://github.com/Qucs/qucs/blob/master/qucs/qucs/qucs_init.cpp#L766 could "just" be built iterating over the available tools... but honestly it's still not clear to me if and how this could be done, so for the moment  I am stopping here :wink: 

Edit: just fixed syntax error in #elif __APPLE__ part of last modification.

### DIFF
--- a/qucs/qucs/qucs.h
+++ b/qucs/qucs/qucs.h
@@ -368,6 +368,7 @@ private slots:
 private:
   void showHTML(const QString&);
   bool performToggleAction(bool, QAction*, pToggleFunc, pMouseFunc, pMouseFunc2);
+  void launchTool(const QString&, const QString&, const QString& = ""); // tool, description and args
   friend class SaveDialog;
   QString lastExportFilename;
 };
--- a/qucs/qucs/qucs_actions.cpp
+++ b/qucs/qucs/qucs_actions.cpp
@@ -1,5 +1,5 @@
 /***************************************************************************
-                               qucsactions.cpp
+                               qucs_actions.cpp
                               -----------------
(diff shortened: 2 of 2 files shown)


yodalee:
It seems `launchTool` with two QString arguments are calling `launchTool` with three QString and leaving the third argument an empty string.
Can you change it to only launchTool with three QString and using default argument?

in3otd:
oops! ...I am more of a C programmer... Fixed now.

yodalee:
Merged

Issue #171, opened 2015-01-10
Build fails on Windows


in3otd:
Trying to build `Qucs` on Windows 7 following the [instruction on the Wiki] (https://github.com/Qucs/qucs/wiki/Build-Windows) (this is the first time I try), I get an error on `qucs-filter` : 

````
g++ -DHAVE_CONFIG_H -I. -I..   -I/c/Qt/4.8.6//include -I/c/Qt/4.8.6//include/Qt
-I/c/Qt/4.8.6//include/QtGui -I/c/Qt/4.8.6//include/QtCore -I/c/Qt/4.8.6//includ
e/QtSvg -I/c/Qt/4.8.6//include/QtXml -I/c/Qt/4.8.6//include/QtScript -I/c/Qt/4.8
.6//include/Qt3Support -DQT_DEBUG -DQT3_SUPPORT -DQT_THREAD_SUPPORT -D_REENTRANT
 -mthreads -DQT_DLL -DUNICODE   -g -O2 -pipe -fno-exceptions -W -Wall -std=c++0x
 -O0 -MT eqn_filter.o -MD -MP -MF .deps/eqn_filter.Tpo -c -o eqn_filter.o eqn_fi
lter.cpp
eqn_filter.cpp: In static member function 'static QString* Equation_Filter::crea
teSchematic(tFilter*)':
eqn_filter.cpp:71:16: error: 'M_PI' was not declared in this scope
   Omega *= 2.0*M_PI;   // angular frequency
                ^
make[2]: *** [eqn_filter.o] Error 1
make[2]: Leaving directory `/c/git/qucs/qucs/qucs-filter'
make[1]: *** [all-recursive] Error 1
make[1]: Leaving directory `/c/git/qucs/qucs'
make: *** [all] Error 2
````
did someone else try to build on Windows recently?

in3otd:
oh, seems to be related to this http://ubuntuforums.org/showthread.php?t=583094 ; if I remove the `-std=c++0x` it compiles ok...  interesting "improvement" in C99 :astonished:

ra3xdh:
What is MinGW version? We moved project from `math.h` to `<cmath>`. See https://github.com/Qucs/qucs/issues/154 for additional information. It builds OK on Linux and `gcc-4.8.2`. Maybe your MinGW version supports <cmath> only partially. Unfortunatelly, I have no Windows and no MinGW and I cannot test and fix it. Maybe add `#ifdef` directive for `M_PI` for Windows?

[...]

PR #190, opened 2015-01-18
Fix windows build

guitorri:
Closes issue #171


guitorri:
TODO:
- [ ] should be `static const`
- [ ] rename M_PI to pi, avoid collision... looks nicer

guitorri:
@in3otd would you like to comment the above changes?

guitorri:
Merged in 6c1fb1e964c3a9199e7ba433fc27aa8fabb16025

in3otd:
...you are going too fast :smile:
sorry no time to closely look at this, only just tried to build on windows also here and it works as expected.
Had a couple of questions on the steps in the wiki, will discuss later.

guitorri:
Hehe, I gave a little push to fix the Windows build. Soon I will add an AppVeyor instance (Travis CI for Windows) to test the build on the Windows platform. Some day it would be interesting to give it a try with the Visual Studio compiler... 
I did some small updates on the wiki... feel free to ask and modify.

in3otd:
regarding the instructions on the Wiki (but it's  the same on Linux), if I clone the git repository and follow the instructions to build `Qucs Core` when I do `sh bootstrap.sh` I get `Could not locate adms autogen script in ./adms, you may use configure with --disable-adms to use installed version` which seems a bit of a chicken and egg problem since `configure` is not available because `bootstrap.sh` did not run... If I clone the submodules then of course it works, but should be not needed, as we use the pre-built binary for adms...

guitorri:
@in3otd the ADMS bootstrap should be fixed in 13ae936ee3e87b20dbd5611e902dc8ed2a6ef440

Issue #178, opened 2015-01-13
`define` inside `qucs` namespace makes no sense


guitorri:
This does not make much sense:
https://github.com/Qucs/qucs/blob/master/qucs-core/src/consts.h#L37-L48

A `#define` is handled by the preprocessor and end up having global scope. 
 
We should replace the `#define` by typed `const` instead.

The fix looks easy, but it has implication on the Verilog models and ADMS generator.

As a mater of fact we had to work around the collision of `T0` from qucs with a `T0` variable on the BSIM6 model.

guitorri:
Fix merged in 1a856890df45b8dd0d547f7cdeffb20d9bfaf352

PR #191, opened 2015-01-19
Fix core namespace

guitorri:
Put constants into qucs namespace.
* Remove defines fro `const.h and `constants.h`, use `static const` instead
* use std:: for min and max functions
* rename kelvin, degree, deg, rad to celcius2kelvin, kelvin2celsius, rad2deg, deg2rad
* move some functions to real.h and complex.h, create overloads in vector, matrix, matvec

The above closes issue #178

I went for a major cleanup of warnings on qucs-core.
Now there are only warning coming from the verilog models, due to a variety of uninitialized variables and other various small issues.

I will cleanup a bit the reverted commits before merging.

I am drafting a proposal for an API to libqucs. This will require us move things around and structure the namespace a bit better.


guitorri:
@bastien-roucaries and @crobarcro would you mind taking a look at these changes? 

Please read also the PR description, there is more work to come after this... Just checking if you have any comments.

Bastien, on cb31903, do we really need to test for the erase iterator type?

bastien-roucaries:
Le 20 janv. 2015 00:01, "Guilherme Brondani Torri" <notifications@github.com>
a écrit :
>
> @bastien-roucaries and @crobarcro would you mind taking a look at these
changes?
>
> Please read also the PR description, there is more work to come after
this... Just checking if you have any comments.
>
> Bastien, on cb31903, do we really need to test for the erase iterator
type?

Yes apple has fixed this bug . So it should be tested.

Bastien

guitorri:
Merged in 1a856890df45b8dd0d547f7cdeffb20d9bfaf352

PR #194, opened 2015-01-24
id_dialog: (Edit Subckt) cannot add new parameters

guitorri:
* The operation is locked in a mutual update:
  - selection on the table update the edit fields
  - change on the edit fiels update/select the table
  - user never get clean input fields for a new set of parameters.

* Fix: Remove the mutual update, add Apply button to copy
  data from edit fields to table and clear the input fields.


guitorri:
@MikeBrinson the issue with the "Edit Subcircuit Properties" should be fixed now.

PR #195, opened 2015-01-24
Refactor replace app to signal slot

yodalee:
This is a refactor work to Schematic and Textdoc
Current design use a lot of "App" to modify the mainwindow, which is not a good design.
This PR replace some of them into Signal-Slot.

### DIFF
--- a/qucs/qucs/qucs.cpp
+++ b/qucs/qucs/qucs.cpp
@@ -1733,6 +1733,7 @@ bool QucsApp::saveAs()
     break;
   }
   Doc->setName(s);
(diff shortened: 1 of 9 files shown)


in3otd:
@yodalee, there is something strange here... if I open several schematics and then close one of the firsts by clicking the "Close Tab" button on the tab actually **two** schematics are closed, not just the one on which I clicked...

yodalee:
I will look into this.

yodalee:
Btw, @in3otd you can just open a new issue and reference to this. Finding a closed PR is a little more difficult than an open issue.

PR #197, opened 2015-01-28
Refactor project view object 

yodalee:
This is a refactor work to main window:
The origin main window manage all the project tree view in the tab, which contaminates the view with the model logic. I create a object ProjectView which can set the project path to it. It will automatically create the TreeView of the directory. If we are going to add something special or modifying the TreeView, just modifying this class.
You can see the number of difference by 
```
git diff master --numstat
191     460     qucs/qucs/qucs.cpp
```
Actually number is just a reference, the importance is the separation of main window view with the file logic.
*******
This is not the complete refactor though. The function that addes, remove, copy files are still in main window. In the end a object of `Project Manager` may be needed, which will manage, that is locked, added, removed, modified, files in the open project.

### DIFF
--- a/qucs/qucs/CMakeLists.txt
+++ b/qucs/qucs/CMakeLists.txt
@@ -112,7 +112,7 @@ SET(QUCS_SRCS
   mouseactions.cpp qucs_actions.cpp	schematic_file.cpp
   wirelabel.cpp node.cpp qucs_init.cpp
   syntax.cpp misc.cpp messagedock.cpp
-  imagewriter.cpp printerwriter.cpp
+  imagewriter.cpp printerwriter.cpp projectView.cpp
 )
 
 SET(QUCS_HDRS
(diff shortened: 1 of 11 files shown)


yodalee:
@guitorri How do you think? I didn't modify the qucs-core but the converge drop OAO

guitorri:
Sorry @yodalee, it looks good but I won't have time to test it this week. 
The Travis CI is raising issues on tests done on qucs-core. I will try to fix that on the weekend.

If you are looking for some more aggressive refactoring, please take a look at my branch https://github.com/guitorri/qucs/compare/QGraphicView-port-exploration. This should be our priority really.
It is rough, but it can already load and save components. A lot more work needs to be done.
The focus is to finish the port of the graphics. Some refactor to methods an to introduce namespaces are acceptable. The most important is get rid of Qt3Support while keeping the same behavior. On a second pass we straighten up everything as we see fit. I have some further notes if you are interested in helping out. Cheers .o/

yodalee:
I'm doing refactor to QGraphis-View right now actually. But I clean some code during the transition.

PR #203, opened 2015-02-11
Fixed errors due to numerical precision.

### LINKED ISSUE #34: Simple LCR-circuit transient simulation fails
I am trying to simulate simple LCR-circuit. For Qucs0.0.17 all works fine. Qucs0.0.18 has very strange behavior. Schematic under simulation is here:
sch
If DC voltage is 12 V, simualtion results are incorrect. Pr1.Vt voltage always is zero. You cannot obtain voltage waveform. If transient simulation stop time is 20ms, simulation process hangs. Here is voltage waveform plot:
volatge-1
If we set less V1 voltage value (for example 5V), simulation works properly. Voltage waveform is damped oscillations. Here is circuit and waveform plot.
sch-2
volt-2
Qucs schematic file is attached:
https://gist.github.com/ra3xdh/9df9d0694d9a4ca6a954/download

So, simulation results are dependent from voltage value even for simple LCR-circuit. What reasons can cause this strange behavior?

in3otd:
...this fixes #34 , hopefully :wink:

The current `tswitch` model had some issues with numerical precision, in the calculation of time differences and of the current switch resistance.

Below is a long explanation (and some questions), for those interested; you need to read it while looking at the [code of `calcTR` of `tswitch.cpp`] (https://github.com/Qucs/qucs/blob/master/qucs-core/src/components/tswitch.cpp#L102)

 In the code, `t` is the current simulation time, `duration` is the switch transition time, then there is a variable `ts=t-duration`and the calculation of the current switch resistance is  based on `ts` and another variable, `tdiff`. When a switching point has not yet been reached, this latter is defined as `tdiff=t-ts` ; from the definitions above it would seem that `tdiff=t-ts=t-t+duration==duration` but due to numerical errors this is not always the case...

As an example, for the circuit in #34, `duration=9.9999999999999995e-07` and when `t=6.5536000000000002e-06` (this is **not** a switching point for the switch), `ts=t-duration=5.5536000000000005e-06` and `tdiff=t-ts=9.9999999999999974e-07` that is `2.1e-22` less than the theoretical value. How could two `double` values differ by so little, when the theoretical machine epsilon for this type is around `2E-16` I do not know. Maybe these variables are kept in registers with higher precision?
The numbers above means that now `tdiff/duration = 0.99999999999999978` instead of `1.0` and 
when computing the switch resistance with a linear approximation (see below why I am not using the cubical spline approximation in this example) `r = roff - ( (roff - ron) * (tdiff / duration) )` with `roff=1e12` and `ron=0` the approx above results in `r=0.000244140625` instead of zero.
At this point `qucsator` realizes that the time point needs to be recomputed (a component suddenly changed value and the current solution is no longer valid) and starts backtracking the solution until `t=3.4844950090878202e-06`, where `ts=2.4844950090878201e-06` and `tdiff=9.9999999999999995e-07` so that `tdiff/duration=1.0` and `r=0`.
At this point `qucsator` is happy again and the simulation time starts going forward again but with smaller steps... but when reaching `t = 3.89988502726346e-06` we have `ts = 2.8998850272634602e-06`, so `tdiff = 9.9999999999999974e-07` and again `r=0.000244140625`. Here `qucsator` starts backtracking again, finds again a good solution, steps forward a little, finds a bad solution and continues like this for a while until the time step becomes too small and the results become `nan` everywhere.
To avoid all this I have simply set the initial value of `ts` well in the past, as I think this was also the original idea.

Another numerical issue was present in the calculation of the current switch resistance using the cubic spline formula; there were two terms computed using higher order powers of small time differences multiplied with very high resistance values which were then subtracted and this made the results sometimes incorrect. For example this made the switch `r` equal to `-2.441406e-04` (negative!) instead of zero when `tdiff==duration`. I have now rearranged the terms to try to minimize the errors and the computed resistance results seems ok.

Out of curiosity, here are the simulation details when using the cubic spline resistance model :

````
NOTIFY: TR1: average time-step 2.88392e-06, 79 rejections
NOTIFY: TR1: average NR-iterations 2.03037, 0 non-convergences
````
run time:

````
real	0m0.219s
user	0m0.214s
sys	0m0.003s
````

And when using the linear resistance model :

````
NOTIFY: TR1: average time-step 2.69407e-06, 73 rejections
NOTIFY: TR1: average NR-iterations 2.02622, 0 non-convergences
````
run time:

````
real	0m0.234s
user	0m0.226s
sys	0m0.004s
````

The cubic spline model seems slightly better, but when stepping thru the code of the actual simulation, for both cases you can see `qucsator` having to backtrack significantly and reducing a lot the time step when the switch changes state, I am even  more convinced we should try to implement the *breakpoints* so `qucsator` will not be caught by surprise when a planned discontinuity appears...

### DIFF
--- a/qucs-core/src/components/tswitch.cpp
+++ b/qucs-core/src/components/tswitch.cpp
@@ -22,6 +22,15 @@
  *
  */
 
+/*!
+ * \file tswitch.cpp
+ * \brief time controlled switch class implementation
+ *
+ * \todo add a property to allow choosing the resistance profile
+ * (cubic spline, linear, abrupt)
+ */
+
+
 #if HAVE_CONFIG_H
 # include <config.h>
 #endif
@@ -86,12 +95,12 @@ void tswitch::initTR (void) {
   qucs::vector * values = getPropertyVector ("time");
   // Find the total time of the switching periods
   T = real (sum (*values));
-  // if the user enters an even number of switchng times
+  // if the user enters an even number of switching times
   // the pattern is repeated continuously
   repeat = (values->getSize () % 2) == 0 ? true : false;
   // make the time taken to go from fully on to fully off
   // the smallest switching time / 100, or the smallest possible
-  // number, but no bogger than the max specified duration
+  // number, but no bigger than the max specified duration
   nr_double_t maxduration = getPropertyDouble("MaxDuration");
   duration = std::min ( std::max (10*NR_TINY, values->minimum() / 100),
                         maxduration );
@@ -112,16 +121,17 @@ void tswitch::calcTR (nr_double_t t) {
   nr_double_t ti = 0;
 
   if (repeat) {
-    // if the user enters an even number of switchng times
-    // the pattern is repeated continuously. This is acieved by
+    // if the user enters an even number of switching times
+    // the pattern is repeated continuously. This is achieved by
     // subtracting an integer number of total switching periods
     // from the real time
     t = t - T * qucs::floor (t / T);
   }
 
-  // Initialise the last switching time to be a full
-  // switching duration
-  nr_double_t ts = t - duration;
+  // Initialise the last switching time to be well in the past
+  // to avoid having the switch even partially in a transition
+  // state (due to inaccurately computed time differences)
+  nr_double_t ts = -2*duration;
 
   // here we determine whether a switching event should occur
   // by looping through the list of switching times and comparing
@@ -145,7 +155,7 @@ void tswitch::calcTR (nr_double_t t) {
       break;
     }
   }
-  // calculate the time since the last switch occured
+  // calculate the time since the last switch occurred
   nr_double_t tdiff = std::max(NR_TINY, t - ts);
 
   // set the time difference to be no more than the max switch
@@ -164,7 +174,7 @@ void tswitch::calcTR (nr_double_t t) {
 
     rdiff = ron - roff;
 
-    s_i = (rdiff) / (duration);
+    //    s_i = (rdiff) / (duration);
   }
   else
   {
@@ -172,12 +182,18 @@ void tswitch::calcTR (nr_double_t t) {
 
     rdiff = roff - ron;
 
-    s_i = (rdiff) / (duration);
+    //  s_i = (rdiff) / (duration);
   }
 
   // perform the interpolation of the constrained cubic spline
-  r = r_0 + ((3. * s_i * qucs::pow (tdiff,2.0)) / (duration))
-          + ((-2. * s_i * qucs::pow (tdiff,3.0)) / qucs::pow (duration, 2.0));
+  //r = r_0 + ((3. * s_i * qucs::pow (tdiff,2.0)) / (duration))
+  //        + ((-2. * s_i * qucs::pow (tdiff,3.0)) / qucs::pow (duration, 2.0));
+  // use Horner's rule to reduce the numerical errors
+  r = r_0 + (((-2. * rdiff * tdiff / duration) + 3. * rdiff) * qucs::pow(tdiff/duration, 2.0));
+
+  // check for (numerical) errors
+  assert(r >= ron);
+  assert(r <= roff);
 
   setD (VSRC_1, VSRC_1, -r);
 }
--- a/qucs-core/src/components/tswitch.cpp
+++ b/qucs-core/src/components/tswitch.cpp
@@ -104,7 +104,6 @@ void tswitch::initTR (void) {
   nr_double_t maxduration = getPropertyDouble("MaxDuration");
   duration = std::min ( std::max (10*NR_TINY, values->minimum() / 100),
                         maxduration );
-
   initDC ();
 }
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by guitorri** [comment]
Thank you for the richness in details!
I will revert the qucs-test to linux baseline so we can continue testing...

**Comment by in3otd** [comment]
BTW, I will change the assert to check against the actual `Ron` and `Roff`, since it is more meaningful than to check just for `r>=0`. Then maybe one day I will also add a property to allow choosing the resistance profile (cubic spline as now, linear variation as it was sometimes in the past, abrupt change as it was originally).

**Comment by guitorri** [comment]
Done, qucs-test is now based on Linux 64bit, qucsator 0.0.17.
I restarted Travis on this pull request, TR_Puls3b_prj is still in NUM_FAIL. I will try to build with your patch on Linux to have a closer look at the numerical differences.
Travis is always pulling the tip of qucs-test:master. Code and tests are decoupled...

**Comment by in3otd** [comment]
I have changed the `assert()` to check against `Ron` and `Roff` and also added the correction for the rectangular voltage source, while I was at it; this latter does not fix completely #41 but at least allows the simulation to start...

**Comment by ra3xdh** [comment]
@in3otd I have tested your PR and found that #34 indeed seems to be fixed. I didn't find any side-effects. I also tested my electrostatic discharge (ESD) models that are susceptible to such sort of bugs. They also still working. I think this PR could  be merged, because it fixes one of the most annoying bugs in simulator.

**Comment by guitorri** [comment]
I restarted Travis, so it picks up the latest qucs-test.

**Comment by in3otd** [comment]
thanks for testing @ra3xdh ; @guitorri did you also update the reference results for `qucs-test` ? I cannot remember now, but I think the Travis  originally failed only due to a small numerical difference w.r.t. on of the reference results, but I am not sure. 
The latest Travis seems to fail due to a problem with doxygen ??

**Comment by guitorri** [comment]
@in3otd the first Travis instance is passing, apart from the `TR_Puls3b_prj` test that fails with numerical difference. If I remember correctly, with your patch the switch is activating one step before/after than the older reference... The second instance is failing on Matplotlib while trying to plot the above error... The third instance (OSX) is timing out for some unknown reason. I will try to get the plots for the `TR_Puls3b_prj`. Perhaps we  just need to update its reference. Doxygen only works after building the master branch.

**Comment by guitorri** [comment]
These are the plots I had in mind, I will rerun with your patch applied: 
https://github.com/Qucs/qucs-test/commit/8941f2b8e93b78bdb21d2529a21133594f89277a
Look for the `TR_Puls3b_prj` to me it looks like one time step difference.

@ra3xdh can you compare ESD circuits you mentioned between qucsator with this patch and ngspice? I think it is safe to merge, just be a bit more certain.

**Comment by in3otd** [comment]
ah, now I remember :grin: 
I think the apparent "one time step" difference is due to the different variation of the switch resistance over time during the switching event. In 0.0.17 it was instantaneous, while later this was changed to be a cubic spline between `Ron` and `Roff` during a non-zero switching time. It will be a good thing to be able to choose the resistance transition shape to compare; I am not sure about what the `ngspice` switch does...

**Comment by ra3xdh** [comment]
@guitorri There are results and schematic from Ngspice and patched Qucsator As you can see ESD current (Pr2.It) is identical for both Qucsator and Ngspice. Test methods are described here http://ieeexplore.ieee.org/xpl/articleDetails.jsp?arnumber=7098391 
cdm-mosfet
iesd

**Comment by guitorri** [comment]
Great! Thank you for looking it up. I will update the faulty test later. 
@in3otd, @ra3xdh general question: How to we do this? Shall we try to always rebase (and merge) or merge directly? I think that rebases makes for a more readable history... What do you think?

**Comment by in3otd** [comment]
@guitorri , I usually always rebase or at least I do before doing a pull request. 
I agree the history will be more readable this way; the current Qucs history looks already messy enough, sometimes...

This is another thing to add to the always-in-progress `git` guide for the Wiki :wink: 

BTW, **do not merge this PR yet**, I have just finished implementing the possibility to choose the resistance transition shape for the switch component, just testing it now, will update the PR (with rebase) later, if you think it could be of interest.

**Comment by guitorri** [comment]
OK, let us know when this PR is ready.

 I wrote some stuff on the Wiki. I forgot to tell. Please add/modify as you see fit.

**Comment by in3otd** [comment]
I have run `qucs-test` on `TR_Puls3b_prj` with different resistance transitions shapes and only the current cubic spline gives the "one-step late" result, the linear and abrupt transitions give the same results as the old reference result. Of course, when forcing the transition time to be very small the results of the *spline* transition are the same as the old reference.

**Comment by in3otd** [comment]
Added the resistance transitions shape selection and rebased to master. Should be ok to merge, please check.

FYI, the resistance transition shapes available look like this:
tswitch
The *abrupt* transition always completes on the first computed point after the specified time, the other just start transitioning on that point.

File: qucs-core/src/components/tswitch.cpp [hunk 1]
```diff
@@ -22,6 +22,15 @@
  *
  */
 
+/*!
+ * \file tswitch.cpp
+ * \brief time controlled switch class implementation
+ *
+ * \todo add a property to allow choosing the resistance profile
```
guitorri: This was done already, no?

in3otd: yes, the `\todo` should have been removed... the latest commit implemented exactly that...

There are others `\todo` around that should be checked, I think. Will address this when doing the Big Formatting Cleanup, one day :grin:

PR #206, opened 2015-03-02
Another step towards solving issue 41...

in3otd:
**EDIT: may continue to work on this, one day...**
needed a lot of time to find a nasty bug that caused almost all transient simulations with BJT having a non-zero  `PTF` (excess phase) parameter to fail.
Now the circuit in #41 runs until `2 ms` then it fails due to the well-known issue with PWL sources; at the beginning the simulator uses small time steps, so it can cope with the source fast transitions but then it increases the time step size and cannot handle the source fast variations anymore.
It's high time to implement the breakpoints now...

While looking at the code saw also a (maybe small) issue with the initial solution (DC) propagation in the transient solver, now fixed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -75,6 +75,7 @@ matrix:
             - texlive-font-utils
             - ps2eps
             - pgf
+            - python-tk
 
     - os: osx
       # Qt 4.8.7 official installer works up to OSX 10.10 or xcode6.4
--- a/qucs-core/src/interface/e_trsolver.cpp
+++ b/qucs-core/src/interface/e_trsolver.cpp
@@ -430,7 +430,7 @@ int e_trsolver::stepsolve_sync(nr_double_t synctime)
     error += predictor ();
 
     // restart Newton iteration
-    restart (); // restart non-linear devices
+    restartNR (); // restart non-linear devices
 
     // Attempt to solve the circuit with the given delta
     try_running () // #defined as:    do {
@@ -677,7 +677,7 @@ int e_trsolver::stepsolve_async(nr_double_t steptime)
         // restart Newton iteration
         if (rejected)
         {
-            restart ();      // restart non-linear devices
+            restartNR ();      // restart non-linear devices
             rejected = 0;
         }
 
--- a/qucs-core/src/nasolver.cpp
+++ b/qucs-core/src/nasolver.cpp
@@ -268,6 +268,8 @@ void nasolver<nr_type_t>::applyNodeset (bool nokeep)
     }
     if (xprev != NULL) *xprev = *x;
     saveSolution ();
+    // propagate the solution to the non-linear circuits
+    restartNR ();
 }
 
 /* The following function uses the gMin-stepping algorithm in order to
--- a/qucs-core/src/trsolver.cpp
+++ b/qucs-core/src/trsolver.cpp
@@ -146,7 +146,6 @@ int trsolver::dcAnalysis (void)
         logprint (LOG_ERROR, "WARNING: %s: %s analysis failed, using line search "
                   "fallback\n", getName (), getDescription ().c_str());
         applyNodeset ();
-        restart ();
         error = solve_nonlinear ();
         break;
     default:
@@ -276,7 +275,7 @@ int trsolver::solve (void)
             // restart Newton iteration
             if (rejected)
             {
-                restart ();      // restart non-linear devices
+                restartNR ();      // restart non-linear devices
                 rejected = 0;
             }
 
@@ -790,17 +789,6 @@ void trsolver::calcTR (trsolver * self)
     for (circuit * c = root; c != NULL; c = (circuit *) c->getNext ())
     {
         c->calcTR (self->current);
-    }
-}
-
-/* Goes through the list of non-linear circuit objects and runs its
-   restartDC() function. */
-void trsolver::restart (void)
-{
-    circuit * root = subnet->getRoot ();
-    for (circuit * c = root; c != NULL; c = (circuit *) c->getNext ())
-    {
-        if (c->isNonLinear ()) c->restartDC ();
     }
 }
 
--- a/qucs-core/src/trsolver.h
+++ b/qucs-core/src/trsolver.h
@@ -53,7 +53,6 @@ public:
     void initTR (void);
     void deinitTR (void);
     static void calcTR (trsolver *);
-    void restart (void);
     void initDC (void);
     static void calcDC (trsolver *);
     void initSteps (void);


in3otd:
...Travis fails... as usual, several issues at once:
- at the first failing circuit, the qucs test python script `run.py` [exits with an error] (https://travis-ci.org/Qucs/qucs/jobs/52813151#L7688) ; might be a different `matplotlib` version ?
- there are several circuits failing (also locally, after updating `qucs-test`...) : most of them are oscillators which works almost as before, but changed phase, that is, they start up a little earlier or later than before. This is due to the DC solution propagation fix, as now the DC starting point used for starting the transient simulation is correctly taken into account and this causes different initial conditions than before. Here below an example:

Here is a plot of the output node for the now-failing `TR_sawtooth-1_prj`, with and without this patch :
tr_sawtooth-1_out
note the initial point (`t=0`) for the original results: there is a strange transient there...
More in detail:
tr_sawtooth-1_out_detail
(with markers on the actual data points, for clarity): the `t=0` point for the original data is different than the DC solution! This apparently is enough to make the oscillator start with a different timing, so the subsequent comparison with the new simulation fails.

In my opinion it will be better, for oscillators, to use the `initialDC=no` option, so that they will start up from zero. Doing this the results from the current and patched `qucsator` are practically identical:
tr_sawtooth-1_out_nodc
(graphs are on top of each other)

Most of the failing circuits have similar issues, will investigate in detail as time permits...

guitorri:
You are probably right, Matplolibt 'framealpha' option is rather new. I any case, the plot and save by default is not necessary. I will remove that (when I find the time...)

The change in phase is also an issue in Windows 32bit (compared to Linux 64bit) in several examples.

Your results are encouraging. Perhaps it is time we figure out a way to cross check against other simulators. We are not even sure that our baseline (golden reference) is correct.

in3otd:
uhm, for DC coupled oscillators the DC bias point used as a starting point for the transient simulation should be an unstable equilibrium point... so I think there is no guarantee that the oscillations will start with a predetermined phase, actually there is no guarantee that it will start at all, AFAIU.
I quickly tried with `ngspice` and I cannot get the circuit to oscillate at all, to make it start I have to nudge it a bit with a brief current pulse. In LTspice instead is starts by itself, with more or less the same phase as the patched `qucsator`.  So I think it will be better to review the oscillator circuits used for testing to make sure they are forced to start, either by skipping the initial DC solution (I think should be enough in most cases) or by the usual small current pulse at the beginning. If this is ok for you, @guitorri, I will prepare a patch for `qucs-test` for this.

guitorri:
It makes sense to enforce known initial conditions, specially on oscillators. Perhaps a nodeset might help. Feel free to update/change the examples. Note that the examples results are now based on qucsator 0.0.17, Linux 64. If you change the source schematic, please try to update also the result file with version 0.0.17.

guitorri:
Can you please rebase this PR?

in3otd:
rebased...

guitorri:
@in3otd have you tried changing the examples? I did in one. Shall I do the others? I am trying to figure out who to push the qucs-test submodule and test with your patch on it. I guess I will have to use your patch an push another branch...

in3otd:
no, sorry. I need a 64-b OS for doing this, as I know that with my 32-b Linux the simulation results will not be good enough, due to the usual numerical differences (qucs-test has always failed here on some test, even on an unmodified `qucsator`).
So I need to update my windows Qucs install, which will be 64 b there, but due to the recent changes the build process on Windows needs to be updated (with mingw/msys I get the same issue as in #408 ) and I'm still trying to figure out all the steps...
So feel free to do the update, you'll likely finish before I'll be able to compile again on Windows :grin:

Not sure I understood the last part of your message, you mean you would like to push both this PR and the updated `qucs-test` reference results at the same time to see Travis passing?

guitorri:
No problem. I changed a bit the 3 failing circuits. In 2 cases, just adding an initial `v=0` to a capacitor did it. In fact this device was instantiated in the netlist with `v=""` (no value i guess). I wonder how the simulator handling that? 
In the oscillator there are lots of zero crossings, somewhere in the comparison it is performing a division by "almost zero" going on and the error explodes... Something is wrong on the `qucs-test` routines.

Now `qucs-test` is a submodule and Travis pulls it recursively. Before we had a line to tell Travis to  pull the head of `qucs-test` on each build. When we do changes on `qucs-test we` have to commit, push somewhere accessible and update `qucs` so it knows which version of `qucs-test` is should use. So, if we update the submodule we also have to commit to `qucs` to reflect that. I hope it is clear... :smile:

guitorri:
In #429 I tried to reset the golden results generated by OSX, but it will fail on Linux. I need to look at more closely. I will postpone this to the next release cycle...
I also messed up qucs-test. I should have pushed a branch, instead of writing to master withoug --no-ff as I did. Anyway, learning this stuff...

in3otd:
ok, we will see for the next release... hoping to have a shorter cycle :wink: 

regarding `qucs-test`, you mean it would have been easier to revert if you merged a branch instead?

guitorri:
The idea is to have shorter cycles. If we manage to get nightly builds with CI services... then we don't have to worry so much anymore... 

Concerning the easier qucs-test reverting a merge. I thing so. I always do --no-ff, I was not thinking straight this time and pushed... when I realized it was too late. That is why PRs are such a good thing. One has time to thing it over. I don't actually know how the history looks like when a merge is reverted.

It would have been "easier" to push a branch to qucs-test, update the submodule in qucs, commit, and push to a branch and pull request. So, qucs would find the commit on the feature branch of qucs-test and tests would run as expected. If the tests go fine we merge the branch in qucs-test, update the submodule in qucs, push to qucs branch and merge this as PR.... :confused: 
Someone said that submodules are more complicated... maybe I am doing it wrong. :)

guitorri:
It seems that most patches from this PR are already merged. No?
We need to enable qucs-test on AppVeyor to check it on 32 and 64 bit Win.

in3otd:
IIRC, only 4dd65ccf3cce8968567c6075432a6abdf13d959d is not yet in. It was not merged since it changed the startup phase of most oscillator circuits (which is usually ill-defined) so required updating `qucs-test` and gave different results on different platforms.
It's ok if I rebase and force-push?

guitorri:
I think a rebase and force-push is fine. I would rather do that link PR to new PR.

guitorri:
I rebased locally. Can I push force to your branch?
Please enable the "Allow edits from maintainers" on the right.

in3otd:
enabled edits from maintainers, feel free to force-push

guitorri:
thank you. 
I want to see the how it runs against the current qucs-test. I will update it later.

in3otd:
um,some `NUM_FAILS` as expected.

But the `clang` build also needs `python-tk`, there is an error `ImportError: No module named _tkinter, please install the python-tk package` which stops the build early (but is not reported as failed...)

And BTW I have the impression that with the new container-based infrastructure and/or `ccache` the OS X build takes too much time, compared to the others. There are also some environment warnings there, not sure if they are relevant.

guitorri:
Previous builds took some 15min on OSX. Sometimes they have a backlog of OSX builds. I am not sure if the timer starts when the instance is launched or when the PR lands on their system. It is also fetching Qt on each build. If the download is slow we see that on the build time. I also think the package manager is updating lots of stuff on each run.... Yeah, things could be improved.

I will check the python-tk.

On my machine indeed the only difference is the starting phase. In any case the results look right.

I guess we just update the projects with `NUM_FAILS` based on `qucs-test/align_issue_41`.

guitorri:
Just to mention that there small NUM_FAIL on OSX 64bit compare to the current golden reference.

@in3otd which platform did you run to create the dat files?

TR_sawtooth-3_prj:
tr_sawtooth-3_prj_darwin__net0 vt

TR_sym_osci_prj:
tr_sym_osci_prj_darwin__net1 vt

in3otd:
I think the `.dat` files were created on a 64-b Linux system or at least I see no NUM_FAILs on this platform here.

Probably, at least for the cases shown above, the issue is still the undefined start phase, the start-up transient has apparently not settled there. I think we discussed forcing the start-up e.g. using a short current pulse but I didn't have time to try if this actually works.

PR #209, opened 2015-03-28
QUCSATOR override

felix-salfelder:
i'd like to play with a different qucsator implementation... 
(i can't think of a less intrusive patch that makes this work.)

also, this fixes a presumed bug in error reporting. SimProcess.Running did not work as potentially intended.

### DIFF
--- a/qucs/qucs/dialogs/simmessage.cpp
+++ b/qucs/qucs/dialogs/simmessage.cpp
@@ -719,7 +719,7 @@ void SimMessage::slotStateChanged(QProcess::ProcessState newState)
       }
     break;
     case QProcess::Starting:
-          ProgText->insertPlainText(tr("Starting ") + Program + ")\n");
+          ProgText->insertPlainText(tr("Starting ") + Program + "\n");
     break;
     case QProcess::Running:
     break;
--- a/qucs/qucs/dialogs/simmessage.cpp
+++ b/qucs/qucs/dialogs/simmessage.cpp
@@ -540,12 +540,12 @@ void SimMessage::startSimulator()
                   << "-o" << "asco_out";
       }
       else {
-	QString x = env.value("QUCSATOR");
-	if(""!=x){
-	  Program = x;
-	}else{
-	  Program = QucsSettings.BinDir + "qucsator" + executableSuffix;
-	}
+        QString x = env.value("QUCSATOR");
+        if(""!=x){
+          Program = x;
+        }else{
+          Program = QucsSettings.BinDir + "qucsator" + executableSuffix;
+        }
         Arguments << "-b" << "-g" << "-i"
                   << QucsSettings.QucsHomeDir.filePath("netlist.txt")
                   << "-o" << DataSet;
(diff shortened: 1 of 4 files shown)


guitorri:
The variable QUCSDIR can be used for relocation. I use that to switch among different qucsator versions. 

If it helps debug and experimentation we can add a QUCSATOR variable as you propose (or QUCSATORBINDIR for consistency with other env variables...).

felix-salfelder:
> The variable QUCSDIR can be used for relocation. I use that to switch among
> different qucsator versions.
QUCSDIR will not cut it. i'd like to specify a command. this seems most natural to me.

> If it helps debug and experimentation we can add a QUCSATOR variable as you propose
that would be great. please consider my second commit. as it removes the file.exists hack.

thank you!

ra3xdh:
I think that new $QUCSATOR environment variable is redundant. These functions could be performed with existing $QUCSDIR. It's not need to introduce this env variable.

felix-salfelder:
I can't seem to set the qucsator using QUCSDIR.
without the proposed change, "qucsator" is hardcoded as part of QString Program in simmessage.cpp...

thanks
felix

ra3xdh:
`QString Program` is derived from `QucsSettings.BinDir` that is derived from `QDir QucsDir` that is derived from `$QUCSDIR`. See main.cpp lines 641--660.

felix-salfelder:
exactly. "Program" is what i meant to override with another program. QUCSATOR appears to be a canonical name, as anything that does not point to a qucsator implementation does not seem to make sense as a value for "Program".

thanks.

ra3xdh:
The only case of usage of $QUCSATOR variable may take place if Qucsator executable is renamed (for example qucsator --> qucsator-0.0.19 ). It is very exotic case.

felix-salfelder:
there's a second use case (see my first message). i'd like to use a different qucsator implementation. neither is its executable name 'qucsator' nor does it have anything to do with QucsBinDir.

this is not a debug problem. also nothing is installed on my side, so an install prefix won't help.

thanks

in3otd:
...`QUCSATORBINDIR` might be a better name for the environment variable, as Guilherme said.

Using `waitForStarted()` should be the right way to be sure the process has started; elsewhere we use a shorter timeout than the default, since 30 s seems a bit long, IMHO. There is no need to explicitly check `state()` after, just use the return value of `waitForStarted()` as done, e.g., [here] (https://github.com/Qucs/qucs/blob/master/qucs/qucs/qucs_actions.cpp#L817)

Good to have a more expressive error message if the process starts to fail, you might want to align the error messages in the other files where we use `QProcess` :grin:

felix-salfelder:
> ...QUCSATORBINDIR might be a better name for the environment variable, as Guilherme said.

for the bindir, sure. but i don't need the bindir, i wouldn't even know what to set it to.
i just want qucs to call a custom command instead of /whichever/path/qucsator.

(i'm not even sure why qucsbindir is not redundant in view of $PATH, but i do not intend to touch anything related to that).

> Using waitForStarted() should be the right way to be sure the process has started;

i put waitForStarted as a workaround, because i do not know how/where to place the callback. the  timeout, however long, does not make any sense to me.

> Good to have a more expressive error message if the process starts to fail, you might want to align the error messages in the other files where we use QProcess

if i had time to refactor the code, i might even do so. i hope this is not a requirement to get the tiny enhancement merged.

thank you

guitorri:
I see, @felix-salfelder you are indeed trying to override the simulator engine, is that right?
My guess is that you are trying test a gnucap+qucsator plug-in... if so, I believe your proposal seems indeed the least intrusive. 
It might be possible to achieve what you want with QUCSDIR and/or a symlink to your custom simulator engine, but that might not work on Windows later on.
I think we can find some use for the QUCSATOR override variable. Do you mind commenting on how to intend to use it?

felix-salfelder:
Hi Guilherme.

> I see, @felix-salfelder you are indeed trying to override the simulator engine, is that right?

yes. i'd like to run something that behaves similar to qucsator but actually is not, rather it is supposed to be a gnucap derivative (something that links to libgnucap + some plugins).

> It might be possible to achieve what you want with QUCSDIR and/or a symlink to your custom simulator engine, but that might not work on Windows later on.

this is no better than compiling a patched custom qucs (or overwrite the qucsator binary). hence the patch. i'm not sure about windows though.

> Do you mind commenting on how to intend to use it?

my intention is to set QUCSATOR to the name of the qucsator replacement then start qucs and see how far i get with this. missing lots of components still...

i have now cleaned up the waitState hack that i introduced earlier (sorry for the confusion).

cheers
felix

PR #210, opened 2015-03-31
Fix in qucs-activefilter

ra3xdh:
This  PR contains fix to avoid infinite loop at some active filter input parameters.

### DIFF
--- a/qucs/qucs-activefilter/filter.cpp
+++ b/qucs/qucs-activefilter/filter.cpp
@@ -416,7 +416,7 @@ bool Filter::calcButterworth()
 
     order = Poles.count();
 
-    return false;
+    return true;
 }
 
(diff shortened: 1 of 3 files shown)


yodalee:
@ra3xdh can you squash these two commits?
and btw @guitorri the coverage test is still under testing right, so we can merge this PR even coverage test fail?

ra3xdh:
Megred two commits. See #212

guitorri:
@yodalee there is no coverage for the Qt part, only for qucs-core. Anyway, last time I set Coveralls not to comment on the PR. They must have changed something. I will take a look.

yodalee:
Oh, actually you don't need open another PR. Just squash the commits then push the branch with `-f` (forced update) flag.

Issue #213, opened 2015-04-05
Copy/paste (and save/load) with 6 coupled inductors does not work


in3otd:
There is a problem with the `N Mutual Inductors` component; using 6 coupled inductors the component is not pasted correctly after being copied (and not loaded correctly after being saved).
An example in [this schematic] (https://gist.github.com/in3otd/1f768b1cbd75efbb7f38); to "see" the bug, delete the mutual inductors component, instantiate a new one, change the number of inductors to 6 and then simulate: everything will work fine.
Now either
- "cut" the component (`Ctrl+X`) and then paste it
- or save the schematic, close it and then reopen it again

now the simulation will give wrong results.

If you make visible all the components properties you will see that after pasting the component (or reopening the file), the last inductance values are mixed up with the coupling coefficients...

ra3xdh:
See #214 for fix.

PR #214, opened 2015-04-06
MutualX fix

ra3xdh:
There is a fix for #213

### DIFF
--- a/qucs/qucs/components/component.cpp
+++ b/qucs/qucs/components/component.cpp
@@ -817,7 +817,7 @@ bool Component::load(const QString& _s)
   else if(Model == "VHDL")
     tmp = 2;
   else if(Model == "MUTX")
-    tmp = 12; // number of properties for the default MUTX (4 inductors)
+    tmp = 5; // number of properties for the default MUTX (2 inductors)
   else tmp = counts + 1;    // "+1" because "counts" could be zero
 
   for(; tmp<=(int)counts/2; tmp++)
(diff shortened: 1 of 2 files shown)


ra3xdh:
Don't merge it. This PR is not ready.

ra3xdh:
Now it is ready for merge

in3otd:
LGTM, but I think there is no need to keep the intermediate corrections to the fixes in the global history, you can probably squash everything into just one or two commits...

ra3xdh:
@in3otd There are two points of view on commits amount: to unite all in a big commit and to store every change in a separate commit. I prefer a second way. It allows to store full history of changes and extract diffs from commit history when it is needed. So, my opinion is to leave commits in this PR as is.

in3otd:
well, I understand it's a matter of opinion and I am not saying we should pull just a single commit for all the changes every time, but, for example, in this particular case I do not see the value in keeping both 7f56fe71a155a0c79e977ef56cea749e6f73d66d and 9885b017d25851bebcfbe3720c7bff91129e9f9e separate in the `Qucs` history. Same for d748874b8eb804c3d093e3fd5404e1d38c66559f and bd2353cea89803ca01bedfb4e78be573e3834ec2 . The first commit 344bcc2c3812b5bfc892ed5b940c9835dc2a4ff8 is also largely rewritten by the following commits. There are 6 commits modifying practically always the same lines in 2 files. 

If we will ever need to check this code some time in the future (hopefully not...) the details of *how* we got to the final solution won't really be useful, again IMHO.

Other points of view are welcome... :smile:

ra3xdh:
@in3otd Reduced number of commits, please check

in3otd:
merged, but for some reason Github did not close the PR automatically (maybe because I closed from from the command line...)

Issue #161, opened 2015-01-07
Shall we remove `qucs-filter-v2`?


guitorri:
The `qucs-filter-v2` seems to have models that are not in `qucs-filter`. However in '-v2' some tabs are missing (#75)  and it seems to be unfinished. 

Furthermore it is not being packaged, it only builds with `-stdlib=libstc++` on OSX (default to `libc++`). 

See also here for issues on BSD: https://sourceforge.net/p/qucs/mailman/message/32825591/

Is someone interested in having a look at it and salvage what can be reused in `qucs-filter`?

ra3xdh:
I think qucs-filter-v2 should not be romoved right now. It contains some interesting filter transfer function approximations (Legendre and Blinchikoff). We should backport them to qucs-filter. After that we can remove qucs-filter-v2. Qucs-filter and qucs-filter-v2 should be merged.

yodalee:
Maybe we can first remove it from compile rule, but leave the data there so we can keep moving our building system forward. qucs-filter-v2 seems containing some dated function declare that give out warning every time I compile.

yodalee:
Remove filter-v2 from build system, wait extracting useful part.

guitorri:
Portions of filter-v2 are not portable. The code is no being build, it will bit rot further.
Let us see if we can merge/refactor the useful parts by 0.0.20.

PR #215, opened 2015-04-13
remove filter-v2 from build system

yodalee:
Issue #161
Remove all build command on filter-v2, the source will stay there until @ra3xdh examine and extract all the useful part inside.


ra3xdh:
You can merge it.

PR #232, opened 2015-04-19
Usability stuff

in3otd:
Here is some random stuff that I modified over the past few months, when I saw some small bugs or details that IMHO needed to be improved. As I didn't want to open several issues/pull requests for lot of small details I kept it all together, but it has now grown too much, sorry.

- for several dialogs (Component Properties, Qucs Settings, etc.) changed the tables re-size mode to show the full content, if possible; adjusted dialogs widgets margins, alignment, spacings; corrected behavior when resizing.
In the `Component Dialog` added the possibility to toggle the visibility for all properties by double clicking on the 'display' header in the properties table. Also improved the handling of file-based components filenames: the location of the currently used file will be used as starting directory when browsing for a new file. If the file is local to the project the relative path will be used.

- in the `Edit Optimization Properties` dialog added a button to copy to clipboard an equation with the current optimization values. Added context menu when right-clicking on the 'initial' column header to change the shown precision of the variables values.
Updated the optimization tutorial describing the new features. Updated all the figures with Qucs 0.0.19 and added text and figures describing the "Copy optimized values to clipboard" and "Set precision" features recently added to the optimization dialog.
Removed usage of Qt3 support functions from `optimizedialog.cpp`.

- Corrected another `QFontMetrics` not using the screen metric in `schematic.cpp`...

- in the `Simulation Messages` window avoid to show an error messages if the simulator is killed by the user pressing the 'Abort Simulation' button.
Also changed `QTextEdit` with `QPlainTextEdit` and adapted code accordingly since `QPlainTextEdit` is better suited to handle large log files, as it can result from large simulations, when debug is enabled or when using the optimizer. It has also the advantage to automatically scroll to the last messages and to handle user scrolling of text nicely.

- in `qucs-edit` set 'modified' to false when loading a file, so it won't ask to save when closing the editor with an unmodified file.

- changed `configure.ac` for `qucs-core` to show `$prefix` in the configure information output (so I know where I specified it to be installed...), also because this is already shown in the `qucs` (GUI) configure output
- in the `configure.ac` for `qucs`added a flag to suppress qDebug() output when debug is disabled.
Also added QT3_WARNINGS flag. **WARNING**: this will flood the build output with warnings... but we need this to remind to remove all the Qt3 stuff...


guitorri:
I will take a look at these now...

guitorri:
You might want to push the updated docs also to https://github.com/Qucs/qucs-help, or open an issue there so we don't forget...

guitorri:
This looks very good to me! Liked the optimizer values as equation... :+1: 
Resolved a small conflict manually... closing via commit.

in3otd:
actually the equation for the optimization values was suggested by `siliconcarne` on the Discussion Help forum on SF; it seemed a good idea so I implemented it...

PR #235, opened 2015-04-21
fix saveAs not update filename in tabwidget

yodalee:
The cause:
DocumentTab->setTabText(DocumentTab->indexOf(This), misc::properFileName(s));

setTabText should apply to TabWidget which contain widget 'w'
However here is apply to widget contain This, that is, main app
*******************

The fix:
indexOf should search for 'w'
DocumentTab->setTabText(DocumentTab->indexOf(w), misc::properFileName(s));

### DIFF
--- a/qucs/qucs/qucs.cpp
+++ b/qucs/qucs/qucs.cpp
@@ -1530,7 +1530,7 @@ bool QucsApp::saveAs()
     break;
   }
   Doc->setName(s);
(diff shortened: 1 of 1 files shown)


yodalee:
I think to bug we can apply simple rule: who found the bug review the solution
review= ＠in3otd

in3otd:
yes, seems a sensible thing to do in general.

PR #236, opened 2015-04-21
fix data display show wrong extension

yodalee:
The repeat steps:
1. Simulate a schematic.
2. Press goto display page at the end of simulation.
3. The corresponding Data Display tab shows extension
   .sch instead of .dpl

The cause:
it's cause by qucs.cpp:L2123
i = DocumentTab->addTab((Schematic *)d, QPixmap(empty_xpm),
    Info.fileName());
where the Info is linked to origin document "*.sch"
so the tab will show "*.sch" rather than "*.dpl"

The fix:
Change Info.fileName() into DataDisplay, which is the filename of *.dpl


yodalee:
review= @in3otd

Issue #231, opened 2015-04-19
File with dot in name are not shown in the Project Content view


in3otd:
File names with dots are not shown in the project content: create a Project, create a file named `MyFile_v0.1.sch` and you will see that it is not shown in the project content...

yodalee:
the projectView.cpp:111, it use completeSuffix() rather than Suffix(), so new file appear in Other category but Schematic.

PR #237, opened 2015-04-21
fix projectView shows dot filename wrong place

yodalee:
The repeat steps:
1. Create a filename with dot like: MyFile_v0.1.sch
2. The file will shows in category "Others" in project content
   But it should be in "Schematics"

The cause:
It's caused by projectView.cpp, when it categorize files by extension
it get extension name with following line:
extName = QFileInfo(workPath.filePath(fileName)).completeSuffix();
which get the complete suffix:
complete suffic of `MyFile_v.0.1.sch` will be `0.1.sch`
which is not a '*.sch" file.

The fix:
Change completeSuffix into suffix
extName = QFileInfo(workPath.filePath(fileName)).suffix();
so filename with dot still get right extensions

### DIFF
--- a/qucs/qucs/projectView.cpp
+++ b/qucs/qucs/projectView.cpp
@@ -108,7 +108,7 @@ ProjectView::refresh()
 
   for(it = files.begin(); it != files.end(); ++it) {
     fileName = (*it).toAscii();
-    extName = QFileInfo(workPath.filePath(fileName)).completeSuffix();
+    extName = QFileInfo(workPath.filePath(fileName)).suffix();
 
     columnData.clear();
     columnData.append(new QStandardItem(fileName));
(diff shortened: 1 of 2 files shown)


yodalee:
review = @in3otd 
solved issue = #231

in3otd:
file *renaming* works well now, but when I *open* a file with a dot in its name, like `MyFile_v.0.1.sch` by double clicking on it in the Project Content, it is opened as a *text document*. Can you take a look at this @yodalee ?

yodalee:
It's also cause be `completeSuffix()`, I add a commit to fix it.
This branch may need squash before merge.

in3otd:
ok, but since `completeSuffix()` has been (correctly) changed to `suffix()` also the complementary `baseName()` calls needed to be changed to `completeBaseName()` in `slotCMenuCopy()` and `slotCMenuRename()`, otherwise these operations will lost part of the file name, if it contains dots...

yodalee:
make sense =w=, fixed.

PR #239, opened 2015-04-23
Vs2010 build

guitorri:
Do not merge! Only using Travis CI.

Fixes to get Qucs Schematic only to build with Visual Studio 2010.


in3otd:
...will you prepare detailed build instructions also for VS in [the wiki] (https://github.com/Qucs/qucs/wiki/Build-Windows) ? Will we move away from MinGW ??

guitorri:
There is no reason to drop MinGW. I will be using a Windows 7 box for a while, it has vs2010 in it... 
No need to merge this yet. I just want to use Travis to see if things are not breaking down. I will update the wiki when the qucs directory gets ready.

Issue #230, opened 2015-04-19
Random crashes when closing a Tab


in3otd:
I get "random" crashes, when closing a Tab. Did not yet find a repeatable sequence to reproduce the bug, but after opening/simulating/closing a few tab Qucs crashes.
Here is a gdb backtrace of a crash happened when closing a tab, don't know how valid it is

````
0  0x00000003 in ?? ()
#1  0x08071331 in QWidget::setHidden (this=0x91ecb30, hidden=true)
    at /usr/lib/qt/include/QtGui/qwidget.h:495
#2  0x0807b9ea in QucsApp::slotHideEdit (this=0x91dce20) at qucs.cpp:2630
#3  0x080839cd in QucsApp::closeFile (this=0x91dce20, index=1) at qucs.cpp:1623
#4  0x08083ea4 in QucsApp::slotFileClose (this=0x91dce20, index=1) at qucs.cpp:1613
#5  0x0810a99e in QucsApp::qt_static_metacall (_o=0x91dce20, _c=QMetaObject::InvokeMetaMethod, _id=12, 
    _a=0xbf958938) at qucs.moc.cpp:269
#6  0xb74eedd5 in QMetaObject::activate(QObject*, QMetaObject const*, int, void**) ()
   from /usr/lib/libQtCore.so.4
#7  0xb6eac665 in QTabWidget::tabCloseRequested(int) () from /usr/lib/libQtGui.so.4
#8  0xb6eac8d9 in ?? () from /usr/lib/libQtGui.so.4
#9  0xb74eedd5 in QMetaObject::activate(QObject*, QMetaObject const*, int, void**) ()
   from /usr/lib/libQtCore.so.4
#10 0xb6ea0325 in QTabBar::tabCloseRequested(int) () from /usr/lib/libQtGui.so.4
#11 0xb6ea03c9 in ?? () from /usr/lib/libQtGui.so.4
#12 0xb74eedd5 in QMetaObject::activate(QObject*, QMetaObject const*, int, void**) ()
   from /usr/lib/libQtCore.so.4
#13 0xb70d7f6d in QAbstractButton::clicked(bool) () from /usr/lib/libQtGui.so.4
#14 0xb6df0881 in ?? () from /usr/lib/libQtGui.so.4
#15 0xb6df1c17 in ?? () from /usr/lib/libQtGui.so.4
[...]

PR #243, opened 2015-04-26
fix crash when close file with editText focus

yodalee:
Solved issue: #230 
Review = @in3otd 

The repeat steps:
1. Open a schematic file.
2. Focus on a component property, show the property edit
3. Close the file
4. The program will crash

The source:
This bug is appeared since commit eadd7b

The cause:
When click on property edit, this line
qucs_action.cpp:1221, editText->reparent(...
will change the editText to child of the open document
At this time, if the document is closed, the editText will
be deleted by Qt
After that, qucs try to access to editText will access to invalid
memory.

The fix:
To be general, now editText will child of main app every time the
slotHideEdit() called, it's a little overhead, but guarantee safe.
function slotHideEdit() {
  ...
  editText->setParent(this, 0);
}

### DIFF
--- a/qucs/qucs/qucs.cpp
+++ b/qucs/qucs/qucs.cpp
@@ -1620,6 +1620,7 @@ void QucsApp::slotFileClose(int index)
 void QucsApp::closeFile(int index)
 {
     statusBar()->message(tr("Closing file..."));
+
     slotHideEdit(); // disable text edit of component property
 
     QucsDoc *Doc = getDoc(index);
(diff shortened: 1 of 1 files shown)


in3otd:
this works, but is there any particular reason not to do the `editText` reparent to the main `App` in the schematic dtor, as it was originally. Just curious...
(your solution is maybe "cleaner" as when the `editText` is not used this way it belongs always to the main `App` and not to the last schematic that used it...)

yodalee:
It creates a strong bond between schematic and main app. The schematic will still have to 'care' about what `editText` status. The `editText` is belong `App` and it's better to preserve the `editText` in App

yodalee:
You can assume that, if in some day, we decide to remove `editText`, even just rename it, we still have to modify code in `Schematic`. This will be a additional work and error-prone.
The purpose of software design is make software modification easier and not so error-prone.

in3otd:
makes sense, thanks for the explanation.

PR #260, opened 2015-05-31
Started fixing text.

nvdl:
First pull request.

### DIFF
--- a/qucs/qucs-filter-v2/qf_dialog.cpp
+++ b/qucs/qucs-filter-v2/qf_dialog.cpp
@@ -159,7 +159,7 @@ FilterDialog::FilterDialog (QWidget * parent) : QDialog (parent)
   dBLabel = new QLabel(tr("dB"),Tab1);
   gp1->addWidget(dBLabel,6,2);
 
-  DualBox = new QCheckBox(tr("dual"),Tab1);
+  DualBox = new QCheckBox(tr("Dual"),Tab1);
   gp1->addMultiCellWidget(DualBox,8,8,0,2);
   CauerPoleBox = new QCheckBox(tr("Stopband is first pole"),Tab1);
   CauerPoleBox->setEnabled(FALSE);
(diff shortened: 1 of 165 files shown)


nvdl:
Fun thing happened at "More More fixes."

nvdl:
Looks like fixing is complete.

guitorri:
Thanks looking into this! We now have to find the right time to merge it... Otherwise most translations will be invalidated.

in3otd:
@nvdl, can you write down the rules you followed when changing text? I see you capitalized the first letter of every string, ok, might be a good thing; then, which other rules you had in mind? It will be a good thing to have the rules to be followed for the user text strings in the Wiki, so we can all have a common reference.
I am asking this also because I saw you still left some `Ac` or `ac` instead of `AC` (for *alternating current*); if the rule is use `AC` for *alternating current* then we should have `AC` everywhere.
Also, in some places you changed `ohms` to `Ohms` which is wrong: unit names, when spelled out in full,  are **not** capitalized (http://physics.nist.gov/Pubs/SP811/sec09.html) . And, BTW, `ohm` is a bit special in Qucs since in several places the unit name is used instead of the unit symbol (maybe because when the project was started more than 10 years ago, the ohm symbol was always not properly handled ??); could we correctly use the unit symbol now ?
Similarly, `Farad` and `Henry` should have been corrected to start with a lower-case letter, when used as units names.
Then other details, like `File based current source` which I think should be `File-based current source`.
I think I still have some notes somewhere about common issues in the user strings, I will try to find it later...

nvdl:
I used a script to do these changes, having a dry run in the beginning, looking into what is to be changed in which manner and doing the wet run.

Regarding Ohms, Henry, I was wrong so I will revert the changes and run the script again.

The symbols idea is a good one; the degress symbol has been fixed in other of my branches.
We can either use symbols or text.
With symbols we have to make sure that if the user changes the application font, the symbol is present.

If interested, I can upload the script and add/modify rules before doing the final wet run.

nvdl:
Added "parse-text-for-translations.py".

guitorri:
@nvdl your Python script is idented by tabs... The PEP8 strongly recommends spaces... :grin:

nvdl:
:smile:
I will fix it.

nvdl:
@in3otd 
Please give https://github.com/nvdl/qucs/blob/nvdl-textual-fixes/scripts/parse-text-for-translations.py a look. Need to take this forward.

in3otd:
will need some time to review this.
There are other things, like 
- "Coul" instead of "C" for the charge unit in `hicumL2V2p31n.cpp`
- "Diac" in several more or less capitalized variants; same for "Triac" (and maybe "SCR"). Should be all caps, IMHO.

PR #261, opened 2015-05-31
GUI Fixes, Changes and SPICE4QUCS Merger

nvdl:


guitorri:
You should not be sending the spice4qucs part... This is still a work in progress being cleaned up and modified by the authors.

If you want to send only your changes with respect to master it makes it easier for us to review and discuss.

This is why 'one branch per feature/issue' is strongly advised...

Now, you can either create new branches and cherry pick your related commits into them, or make a interactive rebase to select just your commits from the list of commits (among other options). Lets us know if you need assistance...

nvdl:
I agree. I will send a new pull request.

Issue #246, opened 2015-05-01
Opening a subcircuit proposes its symbol for pasting


in3otd:
To reproduce the issue:
- create a new project
- create a subcircuit in the project (just insert a couple of ports), save it.
- close the subcircuit schematic Tab
- open the subcircuit by double-clicking on its name in the project view
- the subcircuit schematic will open and **its symbol will be proposed for pasting, sticking to the mouse pointer**. This does not make sense, as one never needs to use the subcircuit symbol in its schematic.

@yodalee can you take a look at this? This behavior appeared after merging #197, before that, opening a subcircuit did not automatically propose its symbol for pasting.

yodalee:
I cannot reproduce this. When I double click on its symbol, it open the `edit component properties` dialog.

yodalee:
Oh, I got it. This issue is more complex than I thought.
Here is the story: in #197 I replace the QTreeWidget with custom QTreeView, and I think that, QTreeView handle double click event less smart then QTreeWidget, which will also fire the single click event. 
Here is some discussion about this, I will try to fix this on weekend:
http://www.qtcentre.org/threads/7858-Double-Click-Capturing
http://stackoverflow.com/questions/4627347/distinguish-between-single-and-double-click-events-in-qt
I think Qucs may face similar problem when switching to `QGraphicsView`

ra3xdh:
There may be another way to fix this issue. We can simple forbid pasting subcircuits from TreeView into itself! You can look at my concept in my test branch: https://github.com/ra3xdh/qucs/tree/246_fix I have tested it and this bug seems to be fixed. Please test it to find undiscovered side-effects.

ra3xdh:
@in3otd Thank you for testing! I have just prepared PR #269 for this fix.

in3otd:
Fixed in #269

PR #269, opened 2015-06-04
Forbidden to paste subcircuit in itself by double click in TreeView

ra3xdh:
This is fix for #246 . See discussion thread at #246 for details.

### DIFF
--- a/qucs/qucs/qucs.cpp
+++ b/qucs/qucs/qucs.cpp
@@ -2284,6 +2284,10 @@ void QucsApp::slotSelectSubcircuit(const QModelIndex &idx)
 
   QString filename = idx.sibling(idx.row(), 0).data().toString();
   QString note = idx.sibling(idx.row(), 1).data().toString();
+  int idx_pag = DocumentTab->currentIndex();
+  QString tab_titl = "";
+  if (idx_pag>=0) tab_titl = DocumentTab->tabText(idx_pag);
+  if (filename == tab_titl ) return; // Forbid to paste subcircuit into itself.
 
   // delete previously selected elements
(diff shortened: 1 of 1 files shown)


qucs/qucs/qucs.cpp
```diff
@@ -2284,6 +2284,10 @@ void QucsApp::slotSelectSubcircuit(const QModelIndex &idx)
 
   QString filename = idx.sibling(idx.row(), 0).data().toString();
   QString note = idx.sibling(idx.row(), 1).data().toString();
+  int idx_pag = DocumentTab->currentIndex();
+  QString tab_titl = "";
+  if (idx_pag>=0) tab_titl = DocumentTab->tabText(idx_pag);
+  if (filename == tab_titl ) return; // Forbid to paste subcircuit into itself.
```
nvdl: Will it handle schematic files with same names but different absolute paths?

ra3xdh:
Absolute paths are never stored in projects.

qucs/qucs/qucs.cpp
in3otd: I thought about this possibility, but since you need to have a Project open to insert a subcircuit symbol by clicking on its filename in the Project View tree it won't be possible to have files with the same names but different absolute paths. All the files for a project reside in the same subdirectory.
Sure, the solution is a bit of a hack...

nvdl: I have not looked so much into hashes. But there is something related to hash function in:
https://github.com/Qucs/qucs/blob/master/qucs/qucs/components/subcircuit.cpp
Line: 331

Doesn't that uniquely identify a schematic?

PR #273, opened 2015-06-09
Add most use bar containing European symbol

smile5:
I Have created european symbol for ground ( different from earth-US gnd symbol ), perfect voltage source and perfect current source.
You can find bitmaps in "png" format for these symbol. I have suffixed them with _eu.png, and for the code I have suffixed them by _eu.c and eu.h.

Thanks


nvdl:
:clap:

in3otd:
Thanks for the pull request; I think there are a few things that could be improved:
- there is no need to duplicate the code for the new components symbols; you can handle them in the same way [as done for the resistor] (https://github.com/Qucs/qucs/blob/master/qucs/qucs/components/resistor.cpp) or [the diode] (https://github.com/Qucs/qucs/blob/master/qucs/qucs/components/diode.cpp) , but this latter does not handle multiple icons, just multiple symbols. 
- changing the default GND symbol to the EU one is not a good idea. The other half (or more) of the users won't be happy with this change. Better will be to have an application setting to select EU or US symbols by default.
- you did the changes in the master branch of your cloned repository. It will be "cleaner" if you did the changes to a separate branch, as [explained here] (https://github.com/Qucs/qucs/wiki/Contribution) . There are a couple of commits that could simply be removed and or squashed with the others. Ideally your branch should be able to be merged *as is* and that means that the commit history of your branch will be merged in the main Qucs history, so it should be as easy to read and straightforward as possible.

Could you please modify your code to take into account the above points?

Then, for the long term, ideally one should be able to select a different symbol by clicking on the generic symbol in the Components tab and e.g. right clicking to open a context menu. And the first item of the context menu might be 'next symbol' so by right-clicking one will cycle thru all the available symbols. This will be useful if in the future someone would like to have more choices for symbols (different GNDs, capacitors, etc.) as it will avoid to fill the Components tab with different symbols for the same function. Might not be so easy to do, though.

Feel free to ask, here or on the [`qucs-devel`] (https://sourceforge.net/p/qucs/mailman/qucs-devel/) mailing list if you have any question.

smile5:
Hello

Thanks for the idea...

For the GND Symbol, I have add a check_box in application settings, so it can be reload at the next restart.

For all other component I have modify or add an european icon, it can be selected in the component property, it also can be done for the ground.

in3otd : i can modify or create other icon, if  "qucs" community want ?  

Sorry for my poor english

Best regards

in3otd:
...did not yet have time to look at the code. Anyway, when you rework some modifications you just did in the previous commit, it will be better if you *squash* everything, so only the final desired results appear.
This is explained a bit more here:
https://github.com/Qucs/qucs/wiki/Contribution#rebase
http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html
Sometimes you see around tutorials warning about "rewriting history", IMHO it's perfectly fine to rewrite the history of topic branches, even after a pull request. On the other end, the official Qucs repository history will never be rewritten (since it is shared with a large number of users), so it's important that the commits going there are as clean as possible.

Regarding icons creation/modification: personally I do not encourage to work on this. It might end up in endless and useless discussions about personal preferences, as it is a rather subjective area.

Your English is fine. And I think most of the contributors here are not native English speakers anyway.
We need also help with the documentation translation: some of the tutorials need to be translated to French or might need an update https://github.com/Qucs/qucs/tree/master/qucs/qucs-help/docs Note that there is a work in progress to improve the translation process here https://github.com/Qucs/qucs-help . Ask on [qucs-devel] (https://sourceforge.net/p/qucs/mailman/qucs-devel/) in case of doubts.
I know that writing documentation might seem less interesting than writing code, but it is at least as useful for the users.

nvdl:
I would add something here regarding icons creation.

I think we need a GUI based symbol creator. Some of the components are off-grid or become off-grid when rotated. Thus, users can create their own symbols while keeping them grid compatible.
We can also share symbols that way.

in3otd:
@nvdl I not sure I understood what you mean. As 'icons' I meant the various GUI icons that you can see e.g. on the toolbar or the components icon in the Components tab on the left. And 'symbols' for me are the ones you place on the schematic.
Symbols by default should be aligned with the grid (i.e. their ports are on the grid), but I did not check them one by one. It is true that some symbols become off-grid when rotated, I think these are mostly symbols whose bounding box is not symmetric around the center, so it's not that the symbols are wrongly defined, but that the rotation algorithm should be improved.
Currently you can already create symbols graphically, for subcircuits. One thing that is missing is exporting/importing a symbol so you could reuse it as a starting point for a new symbol, for a subcircuit.
I think there was some work in the past on having symbols defined using `xml` but I was not involved; you might want to check the [qucs-devel] (https://sourceforge.net/p/qucs/mailman/qucs-devel/) archives.

nvdl:
Yes, I meant the same with icons but that is of low priority as they do not change once a user is happy with them.

**so it's not that the symbols are wrongly defined, but that the rotation algorithm should be improved**
Are you sure that it is the rotation that is causing the issue?
Properly designed symbols do not come off-grid when rotated.

For example:
Libraries->Regulators->LM317K-1

Rotate it, it is off-grid.
Rotate it three more time, it is back on-grid.

If grid compatibility is not needed, it is just dots in the background with no good use or one can argue that it is good to align many but not all components. If wire is to be on-grid, there is no way to snap to off-grid nodes. I think that is the reason, Ctrl+drag was invented to handle this issue where one has to move an off-grid component while it is back on-grid "hopefully".

in3otd:
sorry @smile5 , we hijacked your PR thread...

@nvdl, we probably need an official Qucs glossary . With off-grid I mean a component which has ports not aligned on the grid used for wires, which is always half of the drawn grid (if I understood correctly). The center of rotation is always the origin of the component coordinates and the rotation algorithm currently forces this origin on the grid (IIRC). I think it is not possible, in general, to have the center fixed and rotate a component so that its ports still end on grid points. What should be done is that the component center is slightly moved as needed to have the ports still on grid.
BTW, if a component gets off-grid, `Set on grid`, `Ctrl-U`, will align it. I was tempted to add an automatic `Set on grid` step after every rotation, but some "components", like Paintings, do not necessarily need to be aligned...

nvdl:
**BTW, if a component gets off-grid, Set on grid, Ctrl-U, will align it. I was tempted to add an automatic Set on grid step after every rotation, but some "components", like Paintings, do not necessarily need to be aligned...**

That's it. It works. I will add it to todo list; at least for components that have nodes.

in3otd:
IIRC, one reason why I didn't do it was that even with that fix, the component will not come back to the original position, after 4 90 degrees rotations. It will still have the ports on the grid, but not on the original position, it will be shifted one grid position away. Not a big deal, but not completely right. It is somewhat funny to see a component wandering around when you keep `Ctrl-R` pressed...

nvdl:
I had the same issue with text box about which I was nagging. Looks like an artificial lifeform.
Sorry for hijacking the thread @smile5. :smile:

nvdl:
@smile5 
After closing the pull request, please keep this stuff in a branch.
Will give it a look later and we can discuss how to "themize" these things.

guitorri:
Continued in #304.

PR #274, opened 2015-06-11
Fixup afl

nplanel:
Here a bunch of fixup, mainly segfault occurred during fuzzed input with AFL "american fuzzy lop"

### DIFF
--- a/qucs-core/src/check_netlist.cpp
+++ b/qucs-core/src/check_netlist.cpp
@@ -888,7 +888,7 @@ static int checker_validate_strips (struct definition_t * root)
         if (!def->action)
         {
             /* find components with substrate property */
-            if (checker_is_property (def->define, "Subst") == PROP_STR)
+            if (def->define && (checker_is_property (def->define, "Subst") == PROP_STR))
             {
                 /* check validity of 'Subst' property */
                 if ((val = checker_validate_reference (def, "Subst")) == NULL)
--- a/qucs-core/src/check_netlist.cpp
+++ b/qucs-core/src/check_netlist.cpp
@@ -715,7 +715,14 @@ static int checker_validate_lists (struct definition_t * root)
                                  !strcmp (def->type, "AC") ||
                                  !strcmp (def->type, "SP")))
         {
-            struct value_t * val = checker_find_reference (def, "Type");
+            struct value_t * val;
+            if((val = checker_find_reference (def, "Type")) == NULL) {
+                logprint (LOG_ERROR, "line %d: checker error, required property "
+                          "`%s' is invalid in `%s:%s'\n", def->line, "Type",
+                          def->type, def->instance);
+                errors++;
+                continue;
+            }
             char * type = val->ident;
             // list of constant values and constant values
             if (type && (!strcmp (type, "const") || !strcmp (type, "list")))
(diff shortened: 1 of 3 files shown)


in3otd:
Thanks for running AFL on `qucsator` and taking the time to fix the issues. 
I have just read a bit about AFL, seems quite an interesting tool...

nplanel:
You are warmly welcome, It's more I would like to Thanks all the team for a such good software.
I always found a pity, that's a good software can crash on bad parsing
AFL is the way to find theses, I will do further testing based on dictionaries attack.

in3otd:
I was in fact wondering how did you use AFL exactly, in the default syntax-blind mode or with keywords input, since with this latter mode it should be even more effective, AFAIU.
Could it be used to fuzzy the GUI schematic input (it can load a schematic and output the corresponding netlist `-i`,  `-o` and `-n` options) ?

Should I leave this PR open for the moment in case you find further issues?

nplanel:
Basically AFL inject fault in the input (random byte, change a bit, flip ...) and look by this code instrumentation (each branch) if the execution path change, it's how he learn, which input motif do an  un-explorated branch further. Dictionary (vocabulary) help to found these guess, as it's the hardest part.
This process need to be fast as possible, or you will need a lot of times to tests mostly all your branch (might be impossible, depend on the speed and the code complexity) but you will found crash or stall of your process.

If I remember well, I done the qucsator AFL's test was done only with the netlist checker (and it's a quite slow)
Injecting error on the GUI should be done another way and AFL will be useless, but can got the idea.
Basically you need to :
 1/ records some scenario : (Qt events) and replay them  (I'm sure there are plenty of libs/API)
 2/ inject some fake event based on vocabulary test (pess key 27  ...) when you replay them
 3/ check is the program failed / crash / stall
Running these tests under valgrind or compiling with ASAN (Address Sanitizer) will be warmly helpful

Feel free to contact me directly if you want to talk further.


I think you can close this pull request as if I have more bugs fixies, I will do another one.

BTW I done a PR for ADMS module, but nobody seems come back to me.

in3otd:
Thanks, will contact you for further details on how to best use AFL on `qucsator` and the GUI.

I didn't notice the PR on ADMS, I was not watching the repository, will take a look at it later...

in3otd:
@nplanel , I have just noticed that this PR broke the build on Windows, since in MinGW `strndup()` is not defined. Since it is only used once in `scan_netlist.l` [here] (https://github.com/nplanel/qucs/commit/dc93970d70e25491ed01d7c2abe8bdb48bfcfbb3#diff-8babe943e4195090147e50bdc9571546R184) , I think I'll do as in the original code, modifying the input and using `strdup()`, if you do not see any issue with that (will of course leave the null pointer check)

nplanel:
I think it will be more cleaver to define _GNU_SOURCE on the makefile as MINGW32 sound to need it even if glibc introduce it officially from 2.10 release.

Or may do a trick like this :

#ifdef __MINGW32__
#define _GNU_SOURCE
#include <string.h>
#endif

Side question, do there are a valid reason to not define _GNU_SOURCE during compilation as all the code should compile on all platforms (win, mac, linux) ?

PR #279, opened 2015-06-23
silence irrelevant compiler warnings

felix-salfelder:
i'm getting warnings that absolutely don't help. maybe this is work in progress by someone else. hopefully the fix works for the build bot...

### DIFF
--- a/qucs/qucs/qucs.cpp
+++ b/qucs/qucs/qucs.cpp
@@ -672,7 +672,7 @@ void QucsApp::slotSetCompView (int index)
     for (it = Comps.constBegin(); it != Comps.constEnd(); it++) {
       if ((*it)->info) {
         /// \todo warning: expression result unused, can we rewrite this?
-        *((*it)->info) (Name, File, false);
+        (void) *((*it)->info) (Name, File, false);
         QListWidgetItem *icon = new QListWidgetItem(QPixmap(":/bitmaps/" + QString (File) + ".png"), Name);
         icon->setToolTip(Name);
         CompComps->addItem(icon);
@@ -705,7 +705,7 @@ void QucsApp::slotSearchComponent(const QString &searchText)
       for (modit = Comps.constBegin(); modit != Comps.constEnd(); modit++) {
         if ((*modit)->info) {
           /// \todo warning: expression result unused, can we rewrite this?
-          *((*modit)->info) (Name, File, false);
+          (void) *((*modit)->info) (Name, File, false);
 
           if((Name.indexOf(searchText, 0, Qt::CaseInsensitive)) != -1) {
             //match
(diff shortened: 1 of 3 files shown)


in3otd:
ideally we should have no warnings anywhere... I know we are quite far from this, especially for the Qt3support stuff, but not only. On the master branch there should be no WIP, all the work is done in topic branches and merged when deemed ready.

felix-salfelder:
what do you suggest to do about the warnings in master?

in3otd:
What I try to do is to silence the warnings in the files I have to modify anyway for other reasons; since I  had to go thru the code to understand how it works, this usually helps to figure out if the warnings are "meaningful" or not. So my recommendation will be to do the same :grin:
Of course I am not against, someone systematically going through the code to silence all the warnings.

The remaining Qt3Support ones are, I think, 70% just mere names substitutions, 20% a slightly different behavior of the Qt4 functions to be adapted and the remaining 10% somewhat hard to do.

felix-salfelder:
the warning this PR amends are the non-Qt related warnings. the changes are not hard to figure out, removal of unused variables, and stuff like that. these warnings are a nuisance, as my editor moves the cursor to warnings (and i will certainly not change that behaviour).

I added -Wno-deprecated-declarations to CXXFLAGS, masking  the remaining Qt warnings (which i do not consider irrelevant, but i cannot work on right now).

qucs/qucs/qucs.cpp
```diff
@@ -1793,7 +1793,7 @@ void QucsApp::slotRefreshSchPath()
   this->updateSchNameHash();
   this->updateSpiceNameHash();
 
-  int exit = QMessageBox::information(this, tr("Update schematic path"),
+  QMessageBox::information(this, tr("Update schematic path"),
       tr("The schematic file path has been refreshed."), tr("OK"));
```
nvdl: I do not see a reason for this message box. Of all the other software I have used, "Save As" is supposed to change to a different filename. Or am I missing a point?

guitorri:
Warnings are annoying. I had them fixed somewhere in a branch I forgot to merge... I don't see them as often as often reuse the out of tree build and QtCreator only show warning on recently modified/compiled sources.

The message box is related to a search path feature... there is an issue where Richard explains how it works.

in3otd:
@guitorri , BTW, IMHO the message box about the refreshed search path is just annoying and as useful as a "The file has been saved message". Every time I see that dialog popping up I wonder why we should inform the user about this and force him/her/it to press "Ok" to close the message box. If it is really important to inform the user about this we might just show a message on the status bar. I always wanted to open an issue to propose to remove it but forgotten to do so till now, so I have just opened #281. 

@nvdl, the search path feature is documented [here] (https://github.com/Qucs/qucs/wiki/Subcircuit-Search-Path-%28Experimental%29)

PR #276, opened 2015-06-14
Marker

felix-salfelder:
cleanup related to marker code

PR #282, opened 2015-06-23
Started with the new DC bias points crash fixes. (WIP)

nvdl:
And the dragon rider embarked on a new journey to fix the DC bias point issue with the help of Felix.

### DIFF
--- a/qucs/qucs/diagrams/diagram.cpp
+++ b/qucs/qucs/diagrams/diagram.cpp
@@ -350,13 +350,13 @@ void Diagram::rectClip(Graph::iterator &p) const
 
   if(code1 != 0) if((p-5)->Scr >= 0) { // is there already a line end flag ?
     p++;
-    (p-5)->Scr = STROKEEND;
+    (p-5)->setStrokeEnd();
   }
   if(code1 & code2)   // line not visible at all ?
     goto endWithHidden;
 
   if(code2 != 0) {
-    p->Scr = STROKEEND;
+    p->setStrokeEnd();
     (p+1)->Scr = x_2;
     (p+2)->Scr = y_2;
     z += 3;
@@ -432,7 +432,7 @@ void Diagram::clip(Graph::iterator &p) const
 
   if(dt1 < 0.0) if((p-5)->Scr >= 0.0) { // is there already a line end flag ?
     p++;
-    (p-5)->Scr = STROKEEND;
+    (p-5)->setStrokeEnd();
   }
 
   float x = x_1-x_2;
@@ -469,7 +469,7 @@ void Diagram::clip(Graph::iterator &p) const
   if((dt2 > 0.0) && (dt2 < D)) { // intersection outside start/end point ?
     (p-2)->Scr = x_1 - x*dt2 / D;
     (p-1)->Scr = y_1 - y*dt2 / D;
-    p->Scr = STROKEEND;
+    p->setStrokeEnd();
     p += 3;
     code |= 2;
   }
@@ -509,7 +509,7 @@ void Diagram::calcData(Graph *g)
   auto p = g->begin();
   auto p_end = g->begin();
   p_end += Size - 9;   // limit of buffer
-  p->Scr = STROKEEND;
+  p->setStrokeEnd();
   ++p;
   assert(p!=g->end());
 
@@ -535,17 +535,16 @@ void Diagram::calcData(Graph *g)
 	  if(Counter >= 2)   // clipping only if an axis is manual
 	    clip(p);
 	}
-	if((p-3)->Scr == STROKEEND)
+	if((p-3)->isStrokeEnd() && !(p-3)->isBranchEnd())
 	  p -= 3;  // no single point after "no stroke"
-	else if((p-3)->Scr == BRANCHEND) {
+	else if((p-3)->isBranchEnd() && !(p-1)->isGraphEnd()) {
 	  if(((p-2)->Scr < 0) || ((p-1)->Scr < 0))
 	    p -= 2;  // erase last hidden point
 	}
-	(p++)->Scr = BRANCHEND;
+	(p++)->setBranchEnd();
       }
 
-
-      p->Scr = GRAPHEND;
+      p->setGraphEnd();
 /*z = p-g->Points+1;
 p = g->Points;
 qDebug("\n****** p=%p", p);
@@ -573,10 +572,10 @@ for(int zz=0; zz<z; zz+=2)
           if(insideDiagram(p->Scr, (p+1)->Scr))    // within diagram ?
             p += 2;
         }
-        (p++)->Scr = BRANCHEND;
+	(p++)->setBranchEnd();
 	assert(p!=g->end());
       }
-      p->Scr = GRAPHEND;
+      (p++)->setGraphEnd();
 /*qDebug("\n******");
 for(int zz=0; zz<60; zz+=2)
   qDebug("c: %d/%d", *(g->Points+zz), *(g->Points+zz+1));*/
@@ -640,7 +639,7 @@ for(int zz=0; zz<60; zz+=2)
         else {
 	  dist -= Space;
 	  if((p-3)->Scr < 0)  p -= 2;
-	  else (p++)->Scr = STROKEEND;
+	  else (p++)->setStrokeEnd();
 	  if(Counter < 0)  Counter = -50000;   // if auto-scale
 	  else  Counter = 0;
         }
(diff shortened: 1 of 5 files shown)


nvdl:
@felix-salfelder 
https://github.com/Qucs/qucs/blob/1791ffd4161b1a299fe0c6585feebc80e4b28e69/qucs/qucs/dialogs/sweepdialog.cpp#L166-212

Are you reloading the data file for every node?
I managed to make it crash free but only one bias point (possibly the first) is updated.

In "int Graph::loadDatFile(const QString& fileName)", when you do "return 1", "cPointsY" has been already nullified when "return 2" happened previously. So, it seems that the valid return value is only 2 and not 1 & 2.

Since many things have changed, it may be easier for you to spot the issue and let us know.
I will give it a look later.

nvdl:
The test schematic:

https://gist.github.com/nvdl/57e50125cf172361984e#file-diodes-sweep

felix-salfelder:
> Are you reloading the data file for every node?
the current implementation reads data files as a whole (FileContent = file.readAll(); in Graph::loadDatFile). currently this is hard to fix, but i am thinking about it. (i don't yet believe the comment about the one piece in diagram.cpp).

@nvdl i have never seen this actually working. do you know of an older commit where the code does what is intended?

in3otd:
@felix-salfelder this code seems to have been broken since a long time; apparently it's not working even in 0.0.18. Did not yet have the time (and patience) to bisect to see where the problem appeared.
The strange thing is that is is working for some schematics, especially if there are only a few components. It also was apparently working with the previous "fix" from nvdl, before merging one of your PRs.
So it smells like some buffer overflow or uninitialized pointer that sometime happens to have the right value.

felix-salfelder:
thanks.

please consider [felix-salfelder:dc-bias-points-fix](https://github.com/felix-salfelder/qucs/tree/dc-bias-points-fix).

guitorri:
@nvdl your commit fix the crash, but only one bias value is updated from the `Calculate DC` dialog spinbox to the schematic (only me?)

@felix-salfelder, pressing menu `Simulation > Calculate DC` bias there should be a small dialog to choose the sweept value(s) and its respective DC result(s), with your commit the dialog does not show up (?)

felix-salfelder:
@guitorri yes, die dialog is probably created from the graph instance the return value of setBiasPoints points to. the return NULL; i hacked in is not the right thing. maybe we should (for now) return just one of the involved Graphs (and not delete it).

a proper fix does not abuse Graph and needs more work. i will take care of this after the marker code redesign.

thanks

nvdl:
@guitorri 
_your commit fix the crash, but only one bias value is updated from the Calculate DC dialog spinbox to the schematic (only me?)_
This PR is not complete. I started it to let Felix spot the issues as he has most probably done changes here. So, yes, only one point now.

@felix-salfelder 
_i have never seen this actually working. do you know of an older commit where the code does what is intended?_
In the PR, I proposed before the latest merge from your PRs. I did "git reset --hard" this branch to follow up with the changes. The previous fix is not compatible with the new code anyway but I can share the files if needed.

**So how does it go? I can try to fix it and reduce the number of times data files is loaded; it should be only once. Or should I pause here and let @felix-salfelder do the changes.

DC sweep is very important for my experimentation.
When it was broken, I ended up using a transient simulation with a ramping up source using voltage controlled "x" source.

guitorri:
@nvdl, if you do a PR we assume the code is ready to be tested and merged. You can use the a PR to trigger the builds on Travis, that it is also fine, but let leave us a note. Than I won't pull just to find out the issues you already know :smile:... Do you have an older state on which the code works? Can you put up a branch and let us know about it?

felix-salfelder:
i have pushed the described temporary fix to my local branch. it now returns just one of the graphs and the dialog reappears.

> So how does it go? I can try to fix it and reduce the number of times data files is loaded; it should be only once. Or should I pause here and let @felix-salfelder do the changes.

every call to Graph::loadDatFile scans through the whole file content without actually parsing it. what is needed is a complete/proper parser (creating an object corresponding to the dat file, not just a single chunk of it). could you write one? on this level it makes sense to cache file contents...

nvdl:
@guitorri 
_if you do a PR we assume the code is ready to be tested and merged._
The way I will do is to leave a comment **"The PR is ready"** when it is ready to be pulled.
You had suggested previously that a PR can be created while work is still in progress.

_Do you have an older state on which the code works? Can you put up a branch and let us know about it?_

The previous PR was working but you closed that because of the incompatibility that arose from recent merges. I can recreate that branch but to get along recent changes, things have to be modified in other ways.

@felix-salfelder 
_could you write one?_
Yes, I can give that a try. Have to roll in this code space to see how things are connected.

felix-salfelder:
@nvdi. perfect. i wouldn't know where to put the parser in the current code base. maybe it's best to concentrate on the parser and suitable data structures for the file contents.

Theres some work in progress in my local SimOutput branch, but it still contains lot of breakage and transition glue -- there's some tentative interface definition in qucs/sim/data.h and the current dat parser code adaption in qucs/sim/dat.cpp.

in3otd:
@nvdl regarding PRs, the way we are used to work here is that a PR is ready unless the contrary is explicitly stated. It will be nice if you could follow this convention.
As you said, we agree that "a PR can be created while work is still in progress", I wrote that also [here] (
https://github.com/Qucs/qucs/pull/276#issuecomment-113750959),  we just need to know if it is a WIP.

guitorri:
I think it makes sense to refactor the dataset parser. 

At the moment the datased is read by pieces in 3 different locations (are there other places?) with slight different purposes:
- `DiagramDialog::slotReadVars`
- `Diagram::loadGraphData`: which calls:
 - `Diagram::loadVarData`
 - `Diagram::loadIndepVarData`

Just want to mention that there are other 3 _parsers_ available:
- In `qucs-core` there is a `dataset` parser related to `qucsconv`. This is based on Bison and I am not sure we should be linking the GUI with the backend.
- There is the [Matlab/Octave](https://github.com/Qucs/qucs/tree/master/qucs/qucs/octave) parser.
- And the [Python](https://github.com/Qucs/qucs-test/blob/master/qucstest/qucsdata.py) parser.

It quite easy to port one of the last two into C++. 
Perhaps it makes more sense to load the dataset once after a simulation (or cache it on the first load).

I have to think about the data structure and where to put it.

felix-salfelder:
indeed. qucsdata.py looks close to what is needed.

felix-salfelder:
btw: does the second commit of mine fix the dialog for you? i propose to merge it as a temporary fix and take care of efficiency another time.

guitorri:
It seems to work just fine with flat schematics. 
The schematic [with a subcircuit](https://gist.github.com/nvdl/5ebe56ddebba5986ba95) it is not updating the bias values on the schematic. You might need to set the QUCSDIR to an installed location in order for Qucs to find the subcircuit on the library...

nvdl:
"loadDatFile"; the previous name was more descriptive as it does not load the whole dat file.
"diagram.cpp" has functions from "graph.cpp" possibly because of removal of duplicate functions.
We need a data reader calss that returns what is queried. It may keep the loaded data in the memory if it is not too much.

nvdl:
The new commit fixes the issue.

I disabled the timestamp check as it treats all the nodes equally and only the first node manages to make the function read the data file. For the rest of the nodes, the function return nothing.

https://github.com/Qucs/qucs/commit/fd70523c38f2e9253fcdc1f1de88b813d63d298d#L906-909

Need to rewrite this whole stuff.

Current commit is to stop people from pulling out hair.

felix-salfelder:
> Need to rewrite this whole stuff.

have been thinking about this for a while. lets do it right this time.

what would be the right place for the timestamp check? i think the baseclass for simulation data (lets call it `SimOutputBase`) needs a `virtual SimOutputBase const* refresh() const {return this;}`.

refresh() returns a pointer to a fresh object. maybe `this`, if there's nothing to do. the harder part will be the garbage collection, or resolve dependencies during refreshes. for example (base classes omitted):
 
    DatFileContents* mydatfile = DatFileContents("some/path.dat"); // thinking of something close to the qucsdata.py class
    SomeWaveType* mywave1 = mydatfile->somehowgetWave("var1");
    SomeWaveType* mywave2 = mydatfile->somehowgetWave("var2");

so, clearly, both wave objects may use the same data in mydatfile, as a copy would be too expensive (and a lazy loader does not seem to work in general...). what will `mywave->refresh()` do, if `path.dat` has changed? first of all, SomeWaveType needs a pointer to its parent.

> We need a data reader calss that returns what is queried.

yes. this container class needs to somehow own/track the fileContents objects, however it must not unnecessarily reread a file :). maybe a fileContents object (that is still in use) should hold a pointer to its refreshed self (and owns it until the container claims it)?

what is better, a global container fo all data file objects, or a container that also knows about the directory tree? i guess its easier to locate data in a tree, and the nodes could be used for listing purposes...

the fun has just begun.

nvdl:
I agree; we should later create another thread to discuss the new implementation.

**Regarding reloading the data file, I was thinking if OS file caching is already doing the job for us.**

One other issue is when to load the data file and keep the relevant parts of it in memory and when not; the size of the data file can vary largely.
What if I sweep through 1000 nodes for 1 million points? We need some extreme test schematics as well.

guitorri:
For large data files we better take a leap an use a library to handle it, HDF5 for instance. 
Shall we open another thread?

nvdl:
I have been thinking about HDF5 since the beginning.

Any idea if "qucsator" can be configured to write out another data format?
Though easily readable, for larger simulations the current ".dat" format seems inefficient as values are stored as text.

New thread: #284

nvdl:
New pull request: #286 

This pull request closed on me as I did "hard reset" and "push -f" of the branch.
Or maybe I did something else as well. :cocktail:

Issue #281, opened 2015-06-23
Should the "Update schematic path" popup be removed?


in3otd:
When one saves all the open tabs using "Save all" or uses "Save As" a modal message box telling that ""The schematic file path has been refreshed" pops up and forces the user to press the Ok to continue.
I fail to see why the user should be informed about this in a way which seems to me too intrusive. I think the search path could well be updated without informing the user. Or if it is really important we might just show a message on the status bar. 
What do **you** think?
@crobarcro , what was/is the purpose of the modal dialog?

Of course this is about removing just the modal dialog, the search path feature will not be touched.

crobarcro:
To be honest, I think I've left this in by accident while testing. Feel free to remove it!

guitorri:
A message on the status bar seems fair...

nvdl:
Message on the status bar. +1

in3otd:
ok, I will implement the status bar message...

PR #283, opened 2015-06-24
Remove search path popup

in3otd:
As discusses in #281, the popup message "The schematic file path has been refreshed"  was not really useful, so it has been changed to a more discrete Status Bar message.

I have also changed the deprecated Qt3 functions still used in the code with the equivalent Qt4 ones. I tried to test most of the things I changed, but of course additional testing is welcome.


guitorri:
Tested! Thanks!

PR #286, opened 2015-06-25
DC Bias Points Bug Fix (Hackish) (READY)

nvdl:
Loading data file is not skipped based on the previous loaded time.
This a current hack as 'Graph::loadDatFile()' does not support multi-node data loading.

### DIFF
--- a/qucs/qucs/dialogs/sweepdialog.cpp
+++ b/qucs/qucs/dialogs/sweepdialog.cpp
@@ -27,6 +27,7 @@
 #include <QLineEdit>
 #include <QValidator>
 #include <QPushButton>
+#include <QDebug>
 
 // SpinBoxes are used to show the calculated bias points at the given set of sweep points
 mySpinBox::mySpinBox(int Min, int Max, int Step, double *Val, QWidget *Parent)
@@ -46,7 +47,9 @@ mySpinBox::mySpinBox(int Min, int Max, int Step, double *Val, QWidget *Parent)
 using namespace std;
 QString mySpinBox::textFromValue(int Val) const
 {
-  cout<<"Values + Val"<<*(Values+Val)<<endl;
+  if (Values == NULL) return "";
+
+  qDebug() << "Values + Val" << *(Values+Val) << endl;
   return QString::number(*(Values+Val));
 }
 
@@ -62,6 +65,8 @@ QValidator::State mySpinBox::validate ( QString & text, int & pos ) const
 SweepDialog::SweepDialog(Schematic *Doc_)
 			: QDialog(Doc_)
 {
+  qDebug() << "SweepDialog::SweepDialog()";
+
   Doc = Doc_;
 
   pGraph = setBiasPoints();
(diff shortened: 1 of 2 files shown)


nvdl:
Let's check it and pull it so that it is not broken anymore.

felix-salfelder:
thanks

this is working for me.

qucs/qucs/dialogs/sweepdialog.cpp
```diff
@@ -135,8 +143,9 @@ void SweepDialog::slotNewValue(int)
   Index *= 2;  // because of complex values
 
   QList<Node *>::iterator node_it;
-  QList<double *>::const_iterator value_it;
+  QList<double *>::const_iterator value_it = ValueList.begin();
   for(node_it = NodeList.begin(); node_it != NodeList.end(); node_it++) {
+    qDebug() << "SweepDialog::slotNewValue:(*node_it)->Name:" << (*node_it)->Name;
```
nvdl: This line should be after line 150.

guitorri:
Testing...

guitorri:
Works for me as well! Thank you!

in3otd:
it seems now we understand why the DC Bias Points stuff was not working - read data were still considered valid when asking for the data of another variable, IIUC.
Do we understand why it was working in some cases (small circuits, etc.)? I don't; if the above explanation is correct, is should have never worked...

nvdl:
@in3otd 
I have been wondering about that as well but never tested that.
Do you have a circuit where it works without this PR.
I can check that again.
Were there more than one nodes involved?

in3otd:
the schematic in #272 worked, if I removed one resistor. This was before felix's PR, did not check after.

in3otd:
...besides, the commit message is completely lacking some context; "Those who fail to document context in commit messages condemn all the developers to reread the code"...

guitorri:
Sorry, that was my mistake! I should have asked for a commit message...

nvdl:
Please write down the commit message here so that I can follow the logic and format.

in3otd:
I do not pretend to have captured exactly what was done, but I had in mind a commit message like this one:
````
Invalidate old data when looping to update DC Bias values.

When using 'Calculate DC bias' the loops used to compute the voltages and
currents to be shown in the schematic did not invalidate the previously used 
data. Due to the current implementation of `Graph::loadDatFile()` the data 
needed for the new node(s) were then not actually loaded but the previously loaded 
node data were considered valid instead. This caused Qucs to segfault as the
data pointer used was set to null by <etc., ...>
This is intended to be just a quick fix while the Graph and Diagram classes
are refactored and data loading separated from these classes.
````
I think the guidelines [here] (http://chris.beams.io/posts/git-commit/) and [here] (https://github.com/torvalds/subsurface/blob/master/README#L87-L114) sum up the main points nicely. And we know the author of the latter document is always right :grin: .

nvdl:
Good guidelines. I think, we should make it a requirement for a PR to fulfill.

guitorri:
It is now merged. We cannot rewrite history.

in3otd:
will add the commit guidelines to our [Contributing to Qucs] (https://github.com/Qucs/qucs/wiki/Contribution) wiki page.

nvdl:
One another thing regarding "clang-format", when will that be used?
Sometimes, I have to untangle the code but that obscures the review process in PR.
Thus, the effort put into untangling is lost to keep PR with minimal changes.

So I suggest, to make 2 commits; first commit being the changes and the second commit being the "clang-formatted" version of that file.

More ideas?

in3otd:
yes, some part of the code are badly formatted, but I too avoid to touch them if they are not part of the main PR subject.
The general idea was to format all the code in one shot shortly after 0.0.19 is released, which I guess should be Real Soon. We still need to decide the Coding Style to use.

nvdl:
> We still need to decide the Coding Style to use.

Yes, this was on my mind as well but it was going to open a can of worms.

Issue #259, opened 2015-05-31
checker error, no actions defined: nothing to do


leo-from-spb:
Attempt to simulate has produced the following error:

> checker error, no actions defined: nothing to do

(see the attached screenshot). 
![image](https://cloud.githubusercontent.com/assets/1406510/7902133/123e37b4-07ad-11e5-9b10-76fc6339a1b1.png)

No details what kind of error is it, where it occurred and how to resolve it. 
Is a schema wrong? or something else?

So for me it's a show stopper (really, because I have no idea what happened and how to resolve).

leo-from-spb:
P.S. version 0.0.18

in3otd:
...please include or link the schematic you were trying to simulate, otherwise it might be difficult to guess what went wrong; anyway, I guess you did not place a simulation component (Simulations -> Components) to tell Qucs which kind of simulation to do...

leo-from-spb:
> please include or link the schematic you were trying to simulate

Hmm, I found no idea how to attach file to the issue. And I've already uninstalled the software and cannot take a screenshot of my schema.
I can send the file by e-mail, to which address?

> I guess you did not place a simulation component (Simulations -> Components) to tell Qucs which kind of simulation to do...

You're absolutely right! I didn't do it, I didn't know anything about that components.
In such case I'd show a question "There are no required components. Add them to the schema?", or even better adds all necessary components silently.

nvdl:
I agree, the error and warning system needs a bit of improvement especially for the beginners.

in3otd:
@nvdl, surely some error messages can be improved.
But adding a "Welcome" window at startup telling new users "*Please go to Help->Getting started to see how to use this SW*" and "*Under File->Examples you can find several example schematics ready to be used*" (and maybe together with some details on "What's new" for more experienced users) might be an even better idea.

@leo-from-spb , a simulation component is not required in every schematic, as it might be a subcircuit that is not supposed to be simulated alone. `Qucs` will not add any components silently, since it cannot guess what you would like to add and for doing what (even for simulations, there are 7 different types, which one would you like to do?).
To send a schematic, you can create a Gist in GitHub (top of the page, `Gist`link), you can use one of the many file hosting services (e.g. Dropbox) or you can simply paste the schematic file content (it is a text file)

leo-from-spb:
The file on the gist: 
https://gist.github.com/leo-from-spb/9fd1d2b22a4f6559522b#file-multivibrator-sch
[...]

PR #288, opened 2015-06-26
Abort if there is no simulation component

guitorri:
The simulator is always started, regardless if there is no simulation
block on the schematic. The simulator returns a unclear warning about
'actions'.

This commit checks for the presence of simulation components prior to
executing the simulator.

It also opens a information dialog to inform the user about the missing
simulation component.

This closes issue #259.

### DIFF
--- a/qucs/qucs/qucs.cpp
+++ b/qucs/qucs/qucs.cpp
@@ -2026,6 +2026,26 @@ void QucsApp::slotSimulate()
     return;
   }
 
+  // Check if there is at least on simulation block in the schematic
+  /// \todo Filtering components by the Model string is error prone, use enumeration or similar.
+  Schematic *sch = dynamic_cast<Schematic *>(w);
+  if (sch) {
+    bool haveSimulation = false;
+    for(Component *pc = sch->Components->first(); pc != 0; pc = sch->Components->next()) {
+      QString model = pc->Model;
+      // simulation components start with a "."
+      if (model.startsWith(".")) {
+        qDebug() << "  First simulation found:" << model;
+        haveSimulation = true;
+        break;
+      }
+    }
+    if (! haveSimulation) {
+      QMessageBox::information(this, tr("Info"),tr("Please add a simulation component."));
+      return; // abort
+    }
+  }
+
   SimMessage *sim = new SimMessage(w, this);
   // disconnect is automatically performed, if one of the involved objects
   // is destroyed !


guitorri:
Well, it just need to find the first simulation component. Updating the PR.

in3otd:
I know that these are some "extreme" cases but this does not really work when
-  just a simulation component is present and disabled (will get the usual error message from `qucsator`)
- just a Parameter Sweep simulation component is present (and the error message from `qucsator` will be quite cryptic)
- just an Optimization component is present (and the the messages in the Simulation window are pretty useless)

Since we will support multiple simulators, I think handling these "user errors" should be done by the simulators, we will just show the error messages they produce. So I will tend to make (also) the `qucsator` error messages less cryptic.

guitorri:
You are right, I did not think about the simulations you mentioned and the active/inactive status. 
Still I find it a waste to fire up the simulator if we know in advance that there is no work to be done.

in3otd:
I think it's ok to have a basic check like this, just to cover the case "new user trying to learn how to use Qucs trying random stuff" but in end the only the simulator(s) can say if they are happy with the inputs provided. So we will need also to take a look the error messages from `qucsator` and see if they can be made more expressive.
BTW, also the error codes from `qucsator` should be checked, I had the impression that in a few cases with invalid inputs the simulator complains and exits (as expected) but for some reasons the Qucs GUI says that the simulator has crashed.

felix-salfelder:
currently, the qucs build system sets `-fno-exceptions`. can you think of a sane way to tell the user "hey, i (some simulator) can't do harmonic distortion (or whatever) for you", maybe "this is incomplete, please contribute to __FILE__", or just "the component xy does not exist on my end" without using exceptions? (i can not.)

guitorri:
IMHO the GUI should be proactive and do not invoke a simulator if it can know in advance that there is no work (simulation) to be done. That is all. For all the rest we indeed should think about proper error/exception management.

About `-fno-exception`, I am not sure why the previous developers opted for that, till today there are [several others projects](http://programmers.stackexchange.com/questions/113479/are-there-any-real-world-cases-for-c-without-exceptions) not using exceptions.

felix-salfelder:
@guitorri: interesting read. a lot seems to be about game consoles `:D`

but yes, we don't really need exceptions for the GUI. maybe there is some way to permit exceptions on the border to extensions, such as data acquisition or language or simulator-support plugins, while sticking to `-fno-exceptions` otherwise. i can give it a try some time if you agree with the idea.

Issue #253, opened 2015-05-07
Right click menu for toolbar is missing text for the first three entries


nvdl:
"File", "Edit", "View" and "Work" show up properly but the other entries are missing text.
Can someone point me toward the source. I tried searching.

guitorri:
I never tried RMB on the menu and toolbar... cool.

The named ones are coming form here:
```
$ cd ~/git/qucs/qucs/qucs 
$ grep -rn QToolBar .
./qucs.h:44:class QToolBar;
./qucs.h:270:  QToolBar *fileToolbar, *editToolbar, *viewToolbar
./qucs_init.cpp:30:#include <QToolBar>
./qucs_init.cpp:886:  fileToolbar = new QToolBar(tr("File"));
./qucs_init.cpp:898:  editToolbar = new QToolBar(tr("Edit"));
./qucs_init.cpp:909:  viewToolbar = new QToolBar(tr("View"));
./qucs_init.cpp:917:  workToolbar = new QToolBar(tr("Work"));
```

The unnamed ones are the 3 docks: main left dock, octave dock and verilog-a message dock. It should be possible to add a name to these constructors:
```
$ cd ~/git/qucs/qucs/qucs
$ grep -rn 'new QDockWidget' .
./messagedock.cpp:59:    msgDock = new QDockWidget();
./qucs.cpp:225:  dock = new QDockWidget(this);
./qucs.cpp:365:  octDock = new QDockWidget();
```

nvdl:
Fixed by disabling the context menu and adding the entries to "View" menu.

guitorri:
Do you have a patch for this?

nvdl:
MessageDock is now inherited from QWidget so that it does not create a new dock necessarily. It can be inserted into the Octave dock or other common bottom dock.

Now, there is a separate dock for it.
The dock holding it is made visible or invisible through the view menu.

messagesDock = new QDockWidget();
messageDock = new MessageDock(this);

I think, it should work:
git cherry-pick 7a4edccafda533a7d4238b9cb9240d0a8963ce9e

But there may be some conflicts due to other already done changes.

PR #289, opened 2015-06-26
Fix names of docked widgets

guitorri:
Pressing right-clik on the toolbar brings a context menu to control
the visibility of docked widgets (toobar, docks,...).
Some widgets where unamed and the context menu was blank.
This patch add names to the widgets.

Close issue #253

### DIFF
--- a/qucs/qucs/messagedock.cpp
+++ b/qucs/qucs/messagedock.cpp
@@ -56,7 +56,7 @@ MessageDock::MessageDock(QucsApp *App_): QWidget()
 
     builderTabs->insertTab(1,cppOutput,tr("Compiler"));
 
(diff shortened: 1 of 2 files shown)


qucs/qucs/messagedock.cpp
```diff
@@ -56,7 +56,7 @@ MessageDock::MessageDock(QucsApp *App_): QWidget()
 
     builderTabs->insertTab(1,cppOutput,tr("Compiler"));
 
-    msgDock = new QDockWidget();
+    msgDock = new QDockWidget(tr("admsXml dock"));
```
nvdl: Dock with capital "d"; to be consistent. :wink:

PR #275, opened 2015-06-12
Spice4qucs

ra3xdh:
According #226 `spice4qucs` was successfully rebased and ready for merge. At this stage all seems to be working. Please carefully check all again before merge.

Spice4qucs allows you to simulate Qucs schematic with Ngspice and Xyce. See #77 and #181 for more details.

PR #290, opened 2015-06-26
Change 'Export as Image' Dialog to Native

nvdl:
Uses static QDialog for native look for saving the schematic as an image file.

### DIFF
--- a/qucs/qucs/dialogs/exportdialog.cpp
+++ b/qucs/qucs/dialogs/exportdialog.cpp
@@ -43,7 +43,7 @@ ExportDialog::ExportDialog(int w, int h, int wsel, int hsel, QString filename_, 
     filename = filename_;
 
     lblFilename = new QLabel(tr("Save to file (Graphics format by extension)"));
-    lblResolutionX = new QLabel(tr("Width  in pixels"));
+    lblResolutionX = new QLabel(tr("Width in pixels"));
     lblResolutionY = new QLabel(tr("Height in pixels"));
     lblRatio = new QLabel(tr("Scale factor: "));
     lblFormat = new QLabel(tr("Image format:"));
@@ -52,7 +52,7 @@ ExportDialog::ExportDialog(int w, int h, int wsel, int hsel, QString filename_, 
     connect(ExportButt,SIGNAL(clicked()),this,SLOT(accept()));
     CancelButt = new QPushButton(tr("Cancel"));
     connect(CancelButt,SIGNAL(clicked()),this,SLOT(reject()));
-    SaveButt = new QPushButton(tr("File"));
+    SaveButt = new QPushButton(tr("Browse"));
     connect(SaveButt,SIGNAL(clicked()),this,SLOT(setFileName()));
 
     editFilename = new QLineEdit(filename);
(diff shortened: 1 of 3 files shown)


nvdl:
PR is ready.

guitorri:
I cannot tell the difference, the before/after dialogs look the same on my Mac.

nvdl:
On Linux (Debian), it is different. The static function calls open the native dialog.
File->Open is static and has the native feel and importantly the left panel to select special directories.
Qt 4.8.2.

in3otd:
yes, also on Linux here the static QDialog is the native one, while before it was looking different (same as noted before for spice4qucs [here] (https://github.com/Qucs/qucs/pull/275#issuecomment-111599495)).  
But there is something not right, in that when I change the selected file type
- the extension of the entered file name does not change. Not a big deal, but it was changed accordingly to the selected filter in the old dialog
- most important, the files list shown in the dialog is apparently not updated when changing the filter (I suspect the filter is not set correctly). E.g., if I already have `.jpg` files in the selected directory I do **not** see them listed.

A general comment, not really related to this PR, is that the menu text `Export to image` should actually have an ellipsis (the three dots at the end) to indicate that further actions will be needed before the selected command will be executed (see [here] (http://stackoverflow.com/questions/637683/when-to-use-ellipsis-after-menu-items) for further details).
Conversely, other items like `Refresh Search Path...` should not have them...

nvdl:
> the extension of the entered file name does not change. Not a big deal, but it was changed accordingly to the selected filter in the old dialog

Static QDialog call. The extension is updated when the dialog is closed.

> most important, the files list shown in the dialog is apparently not updated when changing the filter (I suspect the filter is not set correctly). E.g., if I already have .jpg files in the selected directory I do not see them listed.

Remnant of the previous dialog. I did not check if the filter was functional in that implementation.

>that the menu text Export to image should actually have an ellipsis

This is coupled to translation fixes PR, IMHO. We can keep on adding suggestions over there.

in3otd:
> The extension is updated when the dialog is closed. 

yeah, it seems also that the native dialog behavior is OS dependent... In any case we cannot do anything about that.

> Remnant of the previous dialog. I did not check if the filter was functional in that implementation.

**I** did check, and the filter was working in the previous implementation. And even if it was not working it would have been good to fix it...

nvdl:
>I did check, and the filter was working in the previous implementation. And even if it was not working it would have been good to fix it...

You can check the new commit. Looks like static QDialog is sensitive to spaces in the filter string.

in3otd:
The filter works now, thanks.

The logic you use to assign the file extension is different than the one used before: previously, if the user specified a file name with a suffix it was kept unchanged while you force the extension the be the same as the active filter. I don't think it is a good idea, since I expect that the file dialog will be often used to navigate the directories but the file name will be typed in (unless the user wants to overwrite an existing file), and in this case the selected filter should not matter.
E.g.: the default file name proposed is `export.png`; if you open the file dialog to navigate to another directory and then click 'Save' the file name is now changed to `export.svg`, since the default filter is `*.svg` (BTW, it will be nicer if the default filter agreed with the default file name...)
To allow the user to use only supported formats suffixes, a good solution could be like the one implemented for the "Save as..." dialog (see [here] (https://github.com/Qucs/qucs/blob/master/qucs/qucs/qucs.cpp#L1480-L1521)): there you have a list of allowed file suffixes (used as filters) and only if the selected file name has an extension which is not in the allowed list, it could be forced to have the extension of the selected filter, otherwise it will be kept unchanged.

nvdl:
Update, rebased, squahsed and force updated.

in3otd:
> Updated, rebased, squashed and force updated. 

Perfect! :grin:

> >  BTW, it will be nicer if the default filter agreed with the default file name...

>  Did it for startup filename but filter does not follow the entered extension; doable but seems "not so needed".

Yes, this is what I meant, the initially selected filter should agree with the default file name extension, but does not have to follow the selected filter if changed, so what you did is perfectly fine here.

I did not actually try the code yet, but I think that the usage of `QString::replace()`for changing the extension is somewhat dangerous, as it replaces everything in the string. Since you correctly used `QFileInfo::suffix()` to retrieve the extension, you should use `QFileInfo::completeBaseName` to get the part before the suffix.
For example, if the `.png`filter is selected and the user enters a file name `A.name.is.just.a.name` this will be changed to something like `A.png.is.just.a.png`, right?  
Then, what should be the final file name in this case?  `A.name.is.just.a.png` or  `A.name.is.just.a.name.png` ? I will prefer this latter, since it will work nicer for more likely names like `results.old` which will become `results.old.png` instead of `results.png`. So I will propose to append the extension instead of changing the entered name if it doesn't have the right extension. What do you think?

nvdl:
>So I will propose to append the extension instead of changing the entered name if it doesn't have the right extension. What do you think?

I agree, we can do it that way. Have to double check, how the export code handles multiple dots in the filename. Should not be a problem though.

nvdl:
Do not merge, I F*ed up using "push -f".
Does anyone have the latest commit?
Do not want to go to another location on weekend and do the fresh "push".

in3otd:
sorry, I did not pull your latest commits. I think with `git reflog` and some patience you should be able to recover...

nvdl:
PR is ready.

in3otd:
*almost* good.

(Hint: there is a reason why [here] (https://github.com/Qucs/qucs/blob/master/qucs/qucs/qucs.cpp#L1480-L1521) a `QStringList` and not a `QString` is used to contain all the allowed extensions)

(Minor stylistic comment: IMHO an expression like [this] (https://github.com/Qucs/qucs/blob/master/qucs/qucs/qucs.cpp#L1500-L1502) is more readable than [this] (https://github.com/nvdl/qucs/blob/71d59768e9573cdbb9f59a9e5ce841a9b100cf0d/qucs/qucs/dialogs/exportdialog.cpp#L183-L191) )

(unrelated, but might fit in this PR: what about changing the button name [here] (https://github.com/Qucs/qucs/blob/master/qucs/qucs/dialogs/exportdialog.cpp#L55) to "Browse", seems clearer to me)

nvdl:
Did some changes. Will cleanup when all is done.

There is yet one issue to be fixed. If the "Browse" button is used, user will be asked before overwriting the file. But, if "Export" is pressed without using "Browse", the file will be overwritten silently.

One way is to check before "Export". But, if "Browse" is used, user will be asked twice.
The fix is to let "Export" know if confirmation was received or not but we can take that later.

Related to that `QucsApp *QucsMain` is declared in "main.cpp".
Should it be in a header or is it OK to go for `extern QucsApp *QucsMain;`?

I needed it for the `QucsMain->statusBar()`.

in3otd:
(to our loyal readers: the problem in the previous version was that keeping all the allowed extensions in a single string caused extensions partially matching the allowed ones to be accepted. E.g., `results.jp` was wrongly accepted, since it matched part of `.jpg`.)

Nice catch on the possible silent overwrite, I'll put a `//FIXME` about this in the related file.

I have changed the `statusBar()` part, as I don't like the `ImageWriter` to mess with the status bar belonging to the main application. Now `ImageWriter::print()` returns a status which is used by `QucsApp` to decide if the status bar messages need to be shown. 

Looking closely at all the image export code it seems not too "clean"; no parent information is passed to `ImageWriter` (which itself does not really need to know it), but then it has to create the `ExportDialog`, and it cannot pass it the parent it needs, so the dialog does not show the typical application icon (on Linux, at least) on the main dialog window and on any MessageBox that it might show.
I would have expected it to be the other way round, `QucsApp` instantiates an `ExportDialog` which in turn uses the `ImageWriter` methods (some of which could likely be static). It will be refactored, one day.

nvdl:
@in3otd 
So, to be correct, I can assume that you have taken the code into another branch and I can forget about it.

in3otd:
I understood you didn't plan to do further modifications in the near future, so I took your code, added the changes mentioned in my previous message and merged to the master branch.
In case you would like to continue you should start from the current code in the master branch.

nvdl:
Understood correctly. :wink: 
I will pull the changes then.

PR #292, opened 2015-06-29
Fixed library paths for user library components

ra3xdh:
This fixes the following bug. If there are user libraries in `$HOME/.qucs/user_lib` you cannot insert components from them using left-side LibraryTree dock. Components in user libraries appeared as squares. You can insert them into schematic only using Library Tool `Tools->Component Library`. 

Now this bug is fixed. User-defined components can be inserted in schematic from left-side LibraryTree dock.

### DIFF
--- a/qucs/qucs-lib/qucslib_common.h
+++ b/qucs/qucs-lib/qucslib_common.h
@@ -19,6 +19,7 @@
 #define _QUCSLIB_COMMON_H_
 
 #include <QFile>
+#include <QFileInfo>
 #include <QTextStream>
 #include <QDebug>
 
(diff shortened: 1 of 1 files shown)


in3otd:
I'm not familiar with user libraries, but shouldn't the path from `QucsSettings` be used for them, as done in https://github.com/Qucs/qucs/blob/master/qucs/qucs/qucs.cpp#L467 ?

ra3xdh:
Replaced with user library path from `QucsSettings` (last commit updtated).  I thought it is unable to use `QucsSettings` in different source trees.

guitorri:
Looks good to me.
Is the library usage documented somewhere?

guitorri:
Rebased and merged locally.

ra3xdh:
@guitorri , I cannot find anywhere in documentation the description of the process of user libraries development.

Issue #285, opened 2015-06-25
data source independend marker code


felix-salfelder:
the marker code (particularly the move operation) is higly entangled with the double arrays contained in the class holding the data corresponding to a symbol in a .dat file (`DatFile` or something).
this makes it impossible to encapsulate the dat parser into its own class.

ideally, the dat file contents should just be labelled forward iterable waves (or containers with waves), one idea is to copy over the relevant data to the graph instance and do the moving there. why I think this is optimal: because there are only a few pixels on the screen, and it also makes sense to *select* a subset of the data for plotting purposes. this attempt could solve both these issues (provided someone implements a selector).

*maybe* there's a suitable Qt class that implements line segments with user data. it might make sense to use a vector of those replacing Graph::ScrPoints. well, *maybe*.

nvdl:
If we have to sort this out, it should be from the origin.

As a first step, we need a ".dat" to HDF5 converter.

Later, for compatibility with other simulators, we will need a "x-simulator to HDF5" converter.

I personally believe that text based simulation output is good for debugging purposes but not for efficient storage.
It also affects the simulation speed with small timestamps unless one is using SSD.

Later, we will see how to configure "qucsator" to write out HDF5 directly.

I am voting for HDF5 but other possibilities can also be looked into.

felix-salfelder:
> If we have to sort this out, it should be from the origin.

at the origin, there is an *interface*, not a *file format*. file formats lock you into storing data in files. an interface to a data source (be it a dat file or a hdf file) can be used more flexibly.

> Later, we will see how to configure "qucsator" to write out HDF5 directly.

hdf5 possibly is the right output format for qucsator. lets assume it really is. now other simulators will still produce other files... what if hdf6 comes out? will you rewrite qucs another time?

nvdl:
_at the origin, there is an interface, not a file format. file formats lock you into storing data in files. an interface to a data source (be it a dat file or a hdf file) can be used more flexibly._

Sorry, I cannot understand it properly.

__what if hdf6 comes out?__

People are still using HDF4. If HDF6 comes out, we may not need to new features or the upgrade should be minimal.

I am thinking of taking the whole file format thing out from "qucs-gui".

What if we use shared memory between "qucs-gui" and "qucsator" to share simulation data; no file conversion and disk I/O overhead. May be an issue for large simulation data and less memory.
For large simulations, we can emulate memory region while still reading from a memory image file on disk?

[...]

PR #293, opened 2015-06-29
marker cleanup continued

felix-salfelder:
these overdue changes have appeared while working on issue #285. nothing really changes yet...

### DIFF
--- a/qucs/qucs/diagrams/curvediagram.cpp
+++ b/qucs/qucs/diagrams/curvediagram.cpp
@@ -52,7 +52,7 @@ CurveDiagram::~CurveDiagram()
 
 // ------------------------------------------------------------
 void CurveDiagram::calcCoordinate(const double*, const double* yD, const double*,
-				  float *px, float *py, Axis *pa) const
+				  float *px, float *py, Axis const *pa) const
 {
   double yr = yD[0];
   double yi = yD[1];
--- a/qucs/qucs/diagrams/curvediagram.h
+++ b/qucs/qucs/diagrams/curvediagram.h
@@ -31,7 +31,7 @@ public:
   static Element* info(QString&, char* &, bool getNewOne=false);
   int  calcDiagram();
   void calcLimits();
-  void calcCoordinate(const double*, const double*, const double*, float*, float*, Axis*) const;
+  void calcCoordinate(const double*, const double*, const double*, float*, float*, Axis const*) const;
   void finishMarkerCoordinates(float&, float&) const;
   bool insideDiagram(float, float) const;
 
--- a/qucs/qucs/diagrams/diagram.cpp
+++ b/qucs/qucs/diagrams/diagram.cpp
@@ -868,12 +868,18 @@ void Diagram::updateGraphData()
 
   // Setting markers must be done last, because in 3D diagram "Mem"
   // is released in "createAxisLabels()".
-  foreach(Graph *pg, Graphs)
-    foreach(Marker *pm, pg->Markers)
-      pm->createText();
+  foreach(Graph *pg, Graphs){
+    pg->createMarkerText();
+  }
 }
 
 // --------------------------------------------------------------------------
+/*!
+ * does not (yet) load a dat file. only part of it.
+ * this way, it would belong to graph.cpp. but it's too obsolete, lets see..
+ *
+ * FIXME: must invalidate markers.
+ */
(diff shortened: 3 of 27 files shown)


in3otd:
I did some quick testing and LGTM. (but sometimes wondered what the `// BUG` comments were about exactly, it was not always obvious to me...)

felix-salfelder:
thank you.

most `// BUG` comments indicate misplaced code, code that does not do the right thing, or code that shouldn't be there at all, scheduled for cleanup. i started to add them to lines that i had to modify, but couldn't do things right yet...

i will try to be more verbose in future comits.

felix-salfelder:
four more commits, nothing changes. hopefully nothing breaks either...

guitorri:
It is crashing when I try to open this schematic (.dat file added for convenience):
https://gist.github.com/guitorri/667304e29ea41eb3391c

This is from the crash report:
```
0   libsystem_kernel.dylib        	0x00007fff99317866 __pthread_kill + 10
1   libsystem_pthread.dylib       	0x00007fff8faf535c pthread_kill + 92
2   libsystem_c.dylib             	0x00007fff8cf3db1a abort + 125
3   libsystem_c.dylib             	0x00007fff8cf079bf __assert_rtn + 321
4   org.qucs                      	0x000000010f81b01c Graph::ScrPt::setScrY(float) + 92 (graph.cpp:325)
5   org.qucs                      	0x000000010f81b076 Graph::ScrPt::setScr(float, float) + 54 (graph.cpp:332)
6   org.qucs                      	0x000000010f8246f2 Diagram::calcCoordinateP(double const*, double const*, double const*, std::__1::__wrap_iter<Graph::ScrPt*>&, Axis const*) const + 146 (diagram.cpp:2047)
7   org.qucs                      	0x000000010f822531 Diagram::calcData(Graph*) + 1761 (diagram.cpp:522)
8   org.qucs                      	0x000000010f826f7b Diagram::updateGraphData() + 299 (diagram.cpp:848)
9   org.qucs                      	0x000000010f8255a4 Diagram::loadGraphData(QString const&) + 964 (diagram.cpp:805)
10  org.qucs                      	0x000000010f45e822 Schematic::reloadGraphs() + 210 (schematic.cpp:1261)
11  org.qucs                      	0x000000010f45d92f Schematic::becomeCurrent(bool) + 1695 (schematic.cpp:224)
12  org.qucs                      	0x000000010f45e90c non-virtual thunk to Schematic::becomeCurrent(bool) + 44 (schematic.cpp:226)
13  org.qucs                      	0x000000010f480ba8 QucsApp::slotChangeView(QWidget*) + 456 (qucs.cpp:1745)
14  org.qucs                      	0x000000010f474a1e QucsApp::gotoPage(QString const&) + 1902 (qucs.cpp:1385)
15  org.qucs                      	0x000000010f47a6ca QucsApp::slotOpenContent(QModelIndex const&) + 970 (qucs.cpp:2204)
```

felix-salfelder:
@guitorri just loading and plotting from your dat does not seem to crash. am I supposed to open this in connection with a schematic?

(i suspect this has something to do with S parameter stuff...)

thank you

felix-salfelder:
@guitorri interesting corner case. i suspect that negative y are intended here (i did not introduce them, afaics). with dbd3376 all plots from `sparam.sch` look like master to me.

maybe collecting test plots is not too bad a thing to do... `:)`

in3otd:
@guitorri do you think the `-print` option of `qucs` could be expanded to print also the Data Display files? This could maybe allow to have an equivalent `qucs-test` for the diagrams...

guitorri:
That was the idea of having `-print` from the command line. :smile:  To automate generation of reports, documentation and testing... Indeed, my intention is to eventually hook it up into `qucs-test`. I don't remember why I did not try to print the .dpl files. I recall moving some diagrams into the schematic because of this limitation. If you come up with something let me know.

guitorri:
This looks good to me. Let me check #297, then I can merge only that PR right?

in3otd:
it seems that, in general, `qucs` can already print also the `.dpl` files. I didn't try that before...
Following your suggestion, I expanded a bit `qucs-test` to allow printing of `.sch` only, `.dpl` only or both and to have a (sort of) comparison between two different `qucs` version, similar to what can already be done for `qucsator`. See https://github.com/Qucs/qucs-test/pull/7 .

PR #294, opened 2015-06-30
trying to simplify .va compilation rules and fix #164

### LINKED ISSUE #164: Autotools qucs-core parallel build race condition, error.
Some times the `qucs-core` build will error out during a `make -j 8` while running ADMS.
Something on the dependencies is wrong. The compiler wants the verilob-a auto-generated [.h, .cpp] files that are not yet available.

The flex, bison parser generation also fails at times.

felix-salfelder:
- add -o flag to .xml, so the output is less random
- use in build system(s)


felix-salfelder:
ok, this is ready, can someone have a look please?

qucs-core/src/components/verilog/analogfunction.xml
```diff
@@ -1072,8 +1090,8 @@ double $(module)_d_$(function) (<admst:join select="variable[input='yes']" separ
 
 <admst:variable name="globalanalogfunction"/>
 
-<admst:open file="$(module).analogfunction.h">/*
- * $(module).analogfunction.h - analog function header
+<admst:open file="$(filename).$(module).analogfunction.h">/*
```
guitorri: Why did you choose for this two level naming?

qucs-core/src/components/verilog/analogfunction.xml
```diff
@@ -1090,10 +1108,10 @@ double $(module)_d_$(function) (<admst:join select="variable[input='yes']" separ
 #endif /* __$(module)_ANALOGFUNCTION_H__ */
 <admst:text format="\n"/>
 </admst:open>
-<admst:message format="$(module).analogfunction.h created\n"/>
+<admst:message format="$(filename).$(module).analogfunction.h created\n"/>
 
-<admst:open file="$(module).analogfunction.cpp">/*
- * $(module).analogfunction.cpp - analog function implementations
+<admst:open file="$(filename).$(module).analogfunction.cpp">/*
```
guitorri: The message after the file is created needs to be updated:
https://github.com/Qucs/qucs/pull/294/files#diff-da08accacf866067d2f72d9c2d6c60ffL1118

qucs-core/src/components/verilog/qucsMODULEdefs.xml
```diff
@@ -167,9 +167,46 @@
...
+  <admst:if test="[$(filename_state)='1']">
+    <admst:variable name="filename" select="$arg"/>
+    <admst:variable name="filename_state" select="2"/>
+  </admst:if>
+  <admst:if test="[$(filename_state)='0' and $(arg)='-o']">
+    <admst:variable name="filename_state" select="1"/>
+  </admst:if>
+</admst:for-each>
+
+<admst:open file="$(filename).cpp">/*
```
guitorri: The idea here is just to accommodate the namespace or to pull all the model code inside a single file?

guitorri:
It works fine. See some comments above :) 

Changing the names of the files need also to be updated in the documentation: 
http://qucs.sourceforge.net/docs/report/verilog.pdf

For the next bit of documentation one we don't have the sources, MikeBrinson might have it: http://qucs.sourceforge.net/docs/report/verilog.pdf

If we indeed change the file names it might be necessary to update the va2cpp.makefile used for the dynamic loader. I believe the rest of the internals (qucsator, qucsconv) rely on the module name. Things should be fine there.

It is a bit confusing to pass a '-o name' to ADMS and the the actual file output is further mangled by the module name.

qucs-core/src/components/verilog/analogfunction.xml
felix-salfelder: this is just to avoid name clashes/file collisions. sure -- it is a crude workaround. a perfect implementation would not need either a header or even multiple headers for individual modules. we are not there yet.

felix-salfelder:
> If we indeed change the file names it might be necessary to update the va2cpp.makefile used for the dynamic loader. I believe the rest of the internals (qucsator, qucsconv) rely on the module name. Things should be fine there.

good point. will have a look (i'm curious how this works..).

> It is a bit confusing to pass a '-o name' to ADMS

this is really consistent with what other compilers do. think of `gcc foo.cc -o foo`. yet, gcc uses `a.out` as the default, and ADMS uses some random names picked from the input file. it would be nice to not require the `-o` flag and just use the input file name, to create the output file name. but this input name is not accessible from the templates afaics.

> and the the actual file output is further mangled by the module name.

there are additional "helper" files which might be of no use (i am not completely sure). the goal of this chaged is make them irrelevant for the compilation rules. in another iteration one might want to put everything in a single file.

guitorri:
Oops. I missed that you updated the dynamic loader makefiles as well. I will test it on my side.
The parallel build is definitely welcome... we can sort out the generated files (quantity and names) later on.

felix-salfelder:
@guitorri I have hacked some adaptions. i wanted to test, then i got distracted. i do not have a testbench...

this is really hard to fix properly, I only intend to adapt to the first patch. please complain if i broke something. ':)'

guitorri:
This test project should still work:
https://www.dropbox.com/s/oydemg300czruqp/va_loader_test_prj.zip

felix-salfelder:
thank you.

the gui part doesn't work for me. it tries `make -f /home/felix/extern/qucs/qucs/include/qucs-core/va2cpp.makefile ADMSXML=/home/felix/extern/qucs/qucs/admsXml PREFIX=/home/felix/extern/qucs/qucs MODEL=mypotentiometer`. bot, the path to the makefile and the path to admsXml is wrong. need to investigate....

this `make` call seems overengineered to me. it should be something close to `(/path/to/)qucsadms.mk -C $BUILDDIR VPATH=$SRDCIR mypotentiometer.so`.

qucs-core/src/components/verilog/cpp2lib.makefile
```diff
@@ -62,20 +62,19 @@ endif
...
 .PHONY: cpp2lib
 
 # Main target
-# clean before building
-cpp2lib: clean $(MODEL)$(DLLEXT)
+cpp2lib: $(MODEL)$(DLLEXT)
 
 # Build library
-$(MODEL)$(DLLEXT): $(SRC)
+%.$(DLLEXT): %.cpp
```
guitorri: MSYS2 make also dislikes this pattern.

qucs-core/src/components/verilog/va2cpp.makefile
```diff
@@ -48,23 +47,22 @@ ADMS_GEN=.adms.implicit.xml .interface.xml .$(MODEL).va.adms
...
-va2cpp: clean $(GEN_SRC)
+va2cpp: $(MODEL).cpp
 
 # Run admsXml, create C++ out of Verilog-A
 # Last argument '-A dyload' enable code generation for the dynamic loader
 # The '-A' is not really defined. admsXml will take almost any `-[CHAR] command`
 # Include the search path to use constants.vams disciplines.vams distributed by qucs
 # wich are known to work.
-$(GEN_SRC): $(MODEL)$(VA)
+%.cpp: %.va:
```
guitorri: ... and this.

guitorri:
Besides the make pattern rule, it seem to work (with some tweaking).
Yes, the `make` call is mangling the prefix with the bin path. I am looking at it.
The makefiles and the the whole dynamic loader concept are a bit fragile. It was a bit early to release, but back then we didn't have enough people looking at the code... Mike and I decided to put it out as 'alpha'. It seemed useful at that moment. Now I see that I left some of debris all over the place... I will try clean it up.

Can you elaborate further on your suggestion to reduce complexity?

felix-salfelder:
> MSYS2 make

ok, as i'm now familiar with nongnu pattern rules ':D', i'd like to give `.va.cpp` a try.
does your nongnu make support the stem operator `$*`? example:

.SUFFIXES: .from .to
    .from.to:
    \techo $*

`touch file.from; make file.to` should echo "file".

felix-salfelder:
> Can you elaborate further on your suggestion to reduce complexity?

- one makefile should provide functionality.
- this makefile should be executable so $PATH takes care of the whereabouts
- variables should have proper defaults. this is what configure is for.
- should use make semantics, not extra variables (such as `$MODEL`).

here's an example (gnu style, but works) http://git.savannah.gnu.org/cgit/gnucap/gnucap-adms.git/tree/gnucap-adms.mk.in?h=unstable

(i suggest to postpone this...)

guitorri:
`%.cpp: %.va:`  rule with `mingw32-make` yields the following:

```
C:/Users/brondani/local/qucs/parallel/include/qucs-core/va2cpp.makefile:57: ***
mixed implicit and static pattern rules.  Stop.`
```

I guess this is what we need:
`.SUFFIXES: .va .cpp .h` 

To your example:
```
# file.from
.SUFFIXES: .from .to
.from.to:
	echo $*
```
```
> mingw32-make.exe -f file.from file.to
echo file
file
```

I will try to make it work with the current makefiles. Later we see on how to improve. It should be easier to build models from the command line (or from qucsator, perhaps from a .hdl netlist entry) instead of resorting to the GUI... Some of these issues are already noted on the [wiki](https://github.com/Qucs/qucs/wiki/QEP%3A-Verilog-AMS-interface).

felix-salfelder:
> I will try to make it work with the current makefiles

i've already got the patches ready, just wanted to reassure that mingthing supports stems...

shall i still push?

guitorri:
Please push. If not to this PR, perhaps to some other branch I can see :)

felix-salfelder:
it's there. hth

felix-salfelder:
@guitorri retrying to run your inverter example (in the light of #374). somehow, I managed to compile so files from va. the enclosed `README.txt` says "7) Project > Load Verilog-A Module". however, this requires `_symbol.json` files (`QucsApp::slotLoadModule()`). where are these supposed to come from?

(sorry if this is stupid. yet i don't have a single `.json` file in my qucs tree -- i suspect this is related to qucs refusing to open `test.sch`)

ra3xdh:
@felix-salfelder  `.json` files are auto-generated from component symbol. This presentation http://www.mos-ak.org/berkeley_2014/presentations/04_Wladek_Grabinski_MOS-AK_Berkeley_2014.pdf can help you to understand this process. It may be unclear at the beginning. Did you really never use these functions before?

felix-salfelder:
> .json files are auto-generated from component symbol

@ra3xdh thanks. okay, there are ".sym" files in the archive i am talking about. but the `README.txt` file does not mention anything about `.json` files. are they supposed to be generated automatically? by whom?

thanks for the pdf link, i will have a look.

(sorry for being stupid, i will think of a more straightforward way to implement this, but i first need to understand what's there.)

felix-salfelder:
@ra3xdh found it. it seems, admsXml is used to create the json file. (this is not entirely clear from the slides, but they certainly helped.)

ra3xdh:
It's need to document Verilog-A modules dynamic linking process somewhere. Tutorial `Help->Technical reports->verilog.pdf` describes only static linking process.

felix-salfelder:
honestly, i'd prefer do do it right, and *then* document what's left.

currently, i have no luck opening test.sch (from the aforementioned archive), however it becomes clear, that the process is far too complex. lets discuss this later, on a corresponding issue or PR. this one seems ready to merge.

guitorri:
I will test this PR in a minute on Windows...

@felix-salfelder you are right. The readme does not mention that you have to open the .va file and create/edit the symbol and **save** to produce the missing .json files. See below for more details.

I apologize for the lack of documentation on the dynamic loader. 
The feature was rolled out as a bit as a proof of concept. Too bad I never had the time to come back and clean up the mess I left behind. Clearly there are many ways it could be improved... 

Some of the design draft and identified issues are noted down in the [wiki](https://github.com/Qucs/qucs/wiki/QEP%3A-Verilog-AMS-interface)

We better open another issue, but the basic idea was the following:

 - Create symbol: With a .va open the user can _File->Edit Text Symbol_. When the symbol is saved admsXml is called to fetch/format the model parameters. Parameters and the symbol (drawing) are merged into a .json. The .json files are later loaded with a `vacomponent.cpp` to be registered and inserted into the schematic. (Why .json you may ask. Because I was not keen on using XML for simple POD :smile:. It turns out it was a bad decision, because Qt4 has no builtin serializer or writer for JSON... )

- Build module: With a .va open the user does _Project->Build Verilog-a module_. It calls admsXml to produce code and the C++ compiler to build a loadable component.

- Load desired modules within a project(qucs): _Project->Load Verilog-a module_. Allow for user to select loadable modules to be added to the component dock. Add/change icons.

- Running of the compiled modules (qucsator): Path and names of loaded components (compiled model), are passed as comment/annotation in the end of the netlist (I did not want to mess with the netlist entries at that point). Qucsator then loads and register the models before processing the netlist for simulation.

I will be looking forward to hear what would be the _right_ way of doing it... :smirk:

felix-salfelder:
@guitorri thanks for the verbose explanation. i almost figured it out. on my end, something about finding a corresponding .png is broken (need to have another look).

> I will be looking forward to hear what would be the right way of doing it... 

as a tentative outlook, i consider doing this more like the gEDA guys. a symbol is just a symbol, and it references a component type by means of a string identifier (`COMPONENT`, in gEDA).
with this it will be possible to just reference the device implementation, and let the simulator glue figure out the rest. e.g. "if there is a symbol with a (direct/indirect/whatever) reference to foo.va, then make sure that foo.so is ready before runing qucsator (and tell qucsator about it)". there is no need to tell the user what is happening behind the scenes.

there are some details to be figured out, but i haven't found any obstacles yet.

guitorri:
@felix-salfelder I just sent you a PR with a small final fix. With that I think this PR is ready to merge. 

Yes, you have to assign a PNG icon on the _Load Verilog-a model_ dialog. Otherwise it will complain each time it has to use a dummy icon on the component palette. I often save/export the symbol as PNG and use that. I had requests to automate the process and to the save-as-PNG by default.

Yes, the symbol could simply point to the model implementation and everything else is managed by the tool. However If there is an issue you need to able to hack the model and perhaps even the XML files. If the model builds nicely the automatic part you mention is easy to do. My first intention was to provide a barebone IDE that allow the user to sift thy admsXml error messages (errors in the admsXml dock should higlight the corresponding error line in .va file). User can either fix foo.va to match what is currently implemented or to hack the XML files further.

felix-salfelder:
> Yes, you have to assign a PNG icon on the Load Verilog-a model dialog. Otherwise it will complain each time it has to use a dummy icon on the component palette. I often save/export the symbol as PNG and use that. I had requests to automate the process and to the save-as-PNG by default.

iirc, the problem is, when i *try* to assign a png (using the select png button), it says: "cannot find .png". (yes `$emptystring.png`!).

i have no idea yet what is happening. if this is a bug, i'll open an issue. thanks for your help, but lets no longer abuse this thread...

PR #297, opened 2015-07-02
use qt dashed lines implementation

felix-salfelder:
this is actually only one commit on top of the other PR. maybe some motivation :)

(edit: rebased, new commit in said PR)


guitorri:
Looks good to me.

PR #302, opened 2015-07-06
fix a constness issue in Marker

felix-salfelder:
this PR fixes one particular weird case, the `const` Marker::pGraph. some additional code sanitation is required.


guitorri:
Looks good to me. Merged

PR #303, opened 2015-07-08
Searchboxes

in3otd:
The refactor done some time ago on the Project View brought also a nice (instant) search box, at the bottom of the Components tab. But when doing a search, the string shown in the combo box at the top of the Components tab, normally used to select the components family shown, did not change so it was something confusing, since the components icons shown were actually the results of the search.
So I changed the code to show "Search results" in the combo box when searching, to made things more clear.

Then I recalled that also the Library Tool has a search function, for the library components of course, and thought it will be nicer if it had the same kind of instant search function. The current search is not so user friendly, as one needs to click a first button to open the Search Dialog, then press search after entering the string in the Search Dialog, then click a second button to perform the search. And repeat all this again if you would like to do another search. So I changed also the Library tool to have the same kind of search, with a similar appearance to the Component Search in Qucs.

Please check if everything still works fine on your side also.

qucs_lib

(Will be nice to add the same kind of search to the Libraries Tab in the main Qucs GUI, one day, maybe...)


guitorri:
The "Search results" is nice. However, your patch also changed the previous behavior. If you did a search and then selected one component, it would set the family related to that component when the search is cleared. The idea was to take the user to the 'box of components' where he/she is picking the item from. I think it helps people to learn where to find stuff. Shall we keep the previous behavior?

The search on Qucs-lib is welcome. Can we extend it to search also in the `user_lib` which can be either in the `QucsHomeDir` and/or inside a project directory. 

We also need a search function for the 'Edit Component Properties", for the list of parameters. I have a half-working version on a experimental branch for bsim6 (which has some 700 parameters). I will dig it out. 

Other than that, it is working great on my side.

in3otd:
ah, so the previous behavior was intended :confused: ... I purposely changed it to return to the last used component family, since I found it distracting. But I understand it can be helpful for new users to learn where the components are. Should not be too difficult to restore the previous behavior, let me take a look. This also reminds us (me) we should try to document the intended GUI behavior somewhere.

I didn't notice that the `user_lib`is not searched, will fix that.

For BSIM6, you can really have an 'Edit Component Properties' dialog with so many parameters listed?? I understand you need a search there...

guitorri:
Yes, UI documentation is lacking... this should guide the development, not the other way around.

in3otd:
I have restored the previous search behavior and included the User Libraries in the search.

Re-reading your comment "*`user_lib` which can be either in the `QucsHomeDir` and/or inside a project directory*" I'm not sure I understood/agree: AFAIU the User Libraries can not be in a project directory, right??

...just saw there is more work to do: the `user_lib` location is not defined correctly in `qucs-lib/main.cpp` as it uses `QDir::homePath()` instead of `QucsSettings.QucsHomeDir` as parent of `/user_lib/`, will correct that later...

guitorri:
You are right, `user_lib` should be inside `QucsHomeDir`.

in3otd:
corrected the `user_lib/` location definition in `qucs-lib/main.cpp`
Might be nicer squashing some commits, let me know.

Not strictly related to this PR: looking around the code I have somewhat the feeling that the various paths handling is not always done properly: e.g. do the definitions [here] (https://github.com/Qucs/qucs/blob/master/qucs/qucs-lib/main.cpp#L99-L100) work correctly also for user installing in a non-default directory from precompiled packages? I saw that in the Qucs main GUI these are [defined differently] (https://github.com/Qucs/qucs/blob/master/qucs/qucs/main.cpp#L668-L669) in a way which should work in all cases, but most of the tools use the former definition.

guitorri:
I am confused. On Windows the compiler is complaining about 
```
C:\projects\qucs\qucs\qucs-lib\main.cpp:93:36: error: 'homeDirPath' is not a member of 'QDir'
   QucsSettings.QucsHomeDir.setPath(QDir::homeDirPath() + "/.qucs");
                                    ^
``` 
I guess it should be [`QDir::homePath()`](http://doc.qt.io/qt-4.8/qdir.html#homePath), but why is this building on Linux/OSX ?!

guitorri:
Forgot the AppVeyor log reference:
https://ci.appveyor.com/project/qucs/qucs/build/1.0.229#L845

in3otd:
uhm, just noticed that `cmake` fails also on Linux... I (almost) always use the autotools and didn't notice that compiling with `cmake` failed.
Apparently with `cmake` the macro `QT3_SUPPORT` is not defined when compiling `main.cpp`, so the deprecated function is not found. How comes that it was possible to compile until now, given that we still have lot of Qt3 deprecated functions, I have no idea.
Could you/someone please take a look at this, I'm not familiar with the `cmake` stuff, thanks!

in3otd:
BTW, of course the right fix is to use `QDir::homePath()` (will do) but I was wondering why `cmake` is not defining `QT3_SUPPORT`

in3otd:
I have changed `QDir::homeDirPath()` to `QDir::homePath()`, now here it compiles also using `cmake`. Still, I would like to understand why `QT3_SUPPORT` is not defined when using `cmake`...

guitorri:
About 50% of the time, qucslib crashes on launch on OS X. 
Not sure if this is new... I don't use `qucslib` that much.
```
[in3otd-searchboxes]~/git/qucs/qucs $ ./qucs-lib/qucslib 
QWidget::setMinimumSize: (/SymbolWidget) The largest allowed size is (16777215,16777215)
QWidget::setMinimumSize: (/SymbolWidget) The largest allowed size is (16777215,16777215)
Assertion failed: (winRgnBounds.size.height <= 1000000), function CGSNewWindowWithOpaqueShape, file Services/Windows/CGSWindow.c, line 594.
Abort trap: 6
```
And the crash report:
```
Exception Type:  EXC_CRASH (SIGABRT)
Exception Codes: 0x0000000000000000, 0x0000000000000000

Application Specific Information:
Assertion failed: (winRgnBounds.size.height <= 1000000), function CGSNewWindowWithOpaqueShape, file Services/Windows/CGSWindow.c, line 594.
 

0   libsystem_kernel.dylib        	0x00007fff954b2866 __pthread_kill + 10
1   libsystem_pthread.dylib       	0x00007fff8bc9035c pthread_kill + 92
2   libsystem_c.dylib             	0x00007fff890d8b1a abort + 125
3   libsystem_c.dylib             	0x00007fff890a29bf __assert_rtn + 321
4   com.apple.CoreGraphics        	0x00007fff90cbe000 CGSNewWindowWithOpaqueShape + 1229
5   com.apple.AppKit              	0x00007fff8fa2ab36 _NSCreateWindowWithOpaqueShape2 + 540
6   com.apple.AppKit              	0x00007fff8fa29a21 -[NSWindow _commonAwake] + 3720
7   com.apple.AppKit              	0x00007fff8fa286f7 -[NSWindow _makeKeyRegardlessOfVisibility] + 85
8   com.apple.AppKit              	0x00007fff8fa28670 -[NSWindow makeKeyAndOrderFront:] + 29
9   QtGui                         	0x000000010ffd70f3 QWidgetPrivate::show_sys() + 1139
10  QtGui                         	0x0000000110099738 QWidgetPrivate::show_helper() + 408
11  QtGui                         	0x0000000110099a6f QWidget::setVisible(bool) + 511
12  qucslib                       	0x000000010ff5eb7b QWidget::show() + 27 (qwidget.h:497)
13  qucslib                       	0x000000010ff5e43a main + 1770 (main.cpp:125)
14  libdyld.dylib                 	0x00007fff88a1b5fd start + 1
```
Do we have to set the size? :confused: 
I will poke it a bit tomorrow. 
When it works... it works as advertised! :smile:

in3otd:
? I guess the crash comes from (this line) [https://github.com/in3otd/qucs/blob/master/qucs/qucs-lib/symbolwidget.cpp#L452] but I have no idea why. I have never seen `qucslib` crashing here. Do you have any user library installed? Can you place components from the Library Tool to the schematic?

guitorri:
You are right. The `y1`, `y2`, ..., variables are not initialized and for some reason these are picking up garbage. Sometimes very large numbers are picked up, hence failing on the mentioned assert...
I will send you a PR.

guitorri:
I will merge this and send another PR with the fix for the mentioned crash on OSX.
Thank you!.

PR #304, opened 2015-07-09
Modif eu symbol

smile5:
Hello

Back, with more clean git repo, squash, rebase...

Always propose symbol for US user and European user for gnd, Voltage an current sources.

Bye


guitorri:
Why do we need a new component group for 'EU symbol'?

PR #306, opened 2015-07-12
Correct CPW formulae in QucsTranscalc and qucsator

in3otd:
This is a fix for [bug 183 on SF] (https://sourceforge.net/p/qucs/bugs/183/).
The coplanar waveguide (CPW) characteristics formulae were incorrect, mainly due to an incorrect usage of the algorithm used to compute the complete elliptic integral of the first kind `K()`; the argument passed to the function was the elliptic modulus `k`, while the function was expecting the parameter `m=k^2`. Interestingly enough, the CPW calculations mainly need the ratio `K(k)/K(sqrt(1-k^2))`, with `k` in the range from 0 to 1 and the incorrectly used ratio `K(m)/K(1-m)` differs from that by 7% at most over the entire range and is just a few percent off for typical `k` values, so the results were not completely off.
The function computing `K(k)` was rewritten, in a slightly simpler form, to accept the elliptic modulus `k` as input parameter and another calculation for the thickness dependency of the CPW with ground configuration was corrected, in both QucsTranscalc and `qucsator`. (a long-standing issue is the code duplication between QucsTranscalc and the transmission line code in `qucsator`, we should have a common library for the transmission line functions)
The Technical Papers document was updated with a short description of the algorithm used to compute K(k) (and a couple of typos there was fixed).
Out of curiosity, a similar issue was in the documentation of GNU Octave, http://savannah.gnu.org/bugs/?45522 , and this confused me a bit at the beginning...

### DIFF
--- a/qucs-core/src/components/microstrip/cpwline.cpp
+++ b/qucs-core/src/components/microstrip/cpwline.cpp
@@ -47,61 +47,48 @@ cpwline::cpwline () : circuit (2) {
    algorithm (AGM) by Abramowitz and Stegun.
 \todo move to common math
 */
-void cpwline::ellipke (nr_double_t arg, nr_double_t &k, nr_double_t &e) {
-  int iMax = 16;
-  if (arg == 1.0) {
-    k = std::numeric_limits<nr_double_t>::infinity();
-    e = 0;
-  }
-  else if (std::isinf (arg) && arg < 0) {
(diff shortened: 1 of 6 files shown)


guitorri:
I will trust you on this. :wink:

in3otd:
...I just trusted the formulae found in the books :grin: .

PR #309, opened 2015-07-13
Quarter wave bandpass filter

andresmmera:
The quarter wavelength bandpass and notch topologies were added to the qucs-filter tool. Both the transmission line and the microstrip implementation are available.


guitorri:
new PR in #310

PR #313, opened 2015-07-16
remove another two too optimistic assertions.

felix-salfelder:
negative screen coordinates are used to control the painter. just not
always. eject a warning for now.


guitorri:
Thanks!

PR #314, opened 2015-07-16
rework Graph and Marker

felix-salfelder:
- be explicit about Graphs vs GraphDeques
- rearrange data used in diagrams
- store marker position relative to data set
- simplify marker move. now O(numAxes), not O(numSamples)

### DIFF
--- a/qucs/qucs/Makefile.am
+++ b/qucs/qucs/Makefile.am
@@ -93,6 +93,7 @@ EXTRA_DIST = \
 	$(hicolor64_DATA) $(hicolor128_DATA) $(hicolor256_DATA) $(hicolor512_DATA) \
 	$(hicolorsc_DATA) $(desktop_DATA)
 
+SUFFIXES = .moc.cpp .h
 
 .h.moc.cpp:
 	$(MOC) -o $@ $<
--- a/qucs/qucs/diagrams/diagram.cpp
+++ b/qucs/qucs/diagrams/diagram.cpp
@@ -521,6 +521,7 @@ void Diagram::calcData(GraphDeque *g)
 
   g->resizeScrPoints(Size);
   auto p = g->_begin();
+  auto graphbegin = p;
   auto p_end = g->_begin();
   p_end += Size - 9;   // limit of buffer
   p->setStrokeEnd();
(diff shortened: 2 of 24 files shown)


guitorri:
Clang++ (on Linux) is complaining about a copy assign operator.
https://travis-ci.org/Qucs/qucs/jobs/71301491#L4188

felix-salfelder:
weird. using clang++-3.4

    Debian clang version 3.4.2-14 (tags/RELEASE_34/dot2-final) (based on LLVM 3.4.2)
    Target: x86_64-pc-linux-gnu
    Thread model: posix

this does not happen. also, i cant seem to find evidence for

    ./graph.h:133:24: note: copy assignment operator of 'Graph' is implicitly
    deleted because field 'b' has no copy assignment operator
    const const_iterator b;

reported few lines underneath. sure, `b` has a copy assignment operator (?!)

any idea what do try next?

guitorri:
I don't know. This could well be a bug on the used compiler available with Travis.

```
$ clang --version
clang version 3.4 (tags/RELEASE_34/final)
Target: x86_64-unknown-linux-gnu`
```

We can try to setup a newer version of Clang++ on Travis.

felix-salfelder:
i suspect this has to do with a broken stl rather than with clang. the switch to c++11 certainly requires changes to stl that might not be fully available with the currently installed version....

(but i havent had time to check any of this)

felix-salfelder:
ok, seems to work now.

in3otd:
...may I suggest to define a specific macro to enable printing the debug messages related to the Graphs stuff, as I easily get several tens of kB of debug messages on the console when I open a `.dpl` window and this makes difficult to see/scroll back to the other debug messages... thanks!

felix-salfelder:
cannot reopen (?!) see #370

in3otd:
? GitHub thinks you deleted your `graph_rework` branch, maybe this is why it does not allow to reopen the PR

felix-salfelder:
github is right, i deleted the branch. then (later) i recreated the branch... reopen seemed still not possible.

maybe next time i should not delete, but rather push a dummy instead.

whatever, sorry for the noise `:|`

in3otd:
not sure I understood what you wanted to do, couldn't you use `push --force` to the remote `graph_rework` branch ? Anyway, we will continue on 370...

felix-salfelder:
> `push --force`

this is exactly what i did, just before i noticed that i had rebased to a wrong base (stupid me)....

hope this clarifies what happened.

PR #315, opened 2015-07-17
support for multiple simulators

felix-salfelder:
as mentioned in #275, we need a basis for supporting different/several simulators and netlist languages. these commits introduce some basic infrastructure and a first stub (for qucsator) taking care of some netlisting.

the simulators modules are meant to be dynamically loaded using dlopen. this is not complete and might cause some linking or even run time headache. the cmake build is incomplete, please help...

EDIT: now also works with cmake (preliminary solution -- same as automake)

### DIFF
--- a/qucs/autogen.sh
+++ b/qucs/autogen.sh
@@ -31,6 +31,8 @@ echo "done."
 echo -n "Creating config.h.in... "
 autoheader
 echo "done."
+echo "Running libtoolize"
+libtoolize || exit 1
 echo -n "Creating Makefile.in(s)... "
 ${AUTOMAKE:-automake} -a -f -c
 echo "done."
--- a/qucs/autogen.sh
+++ b/qucs/autogen.sh
@@ -25,6 +25,11 @@
 here=`pwd`
 cd `dirname $0`
 
+case `uname` in
+  *Darwin*) LIBTOOLIZE=glibtoolize ;;
+  *) LIBTOOLIZE=libtoolize ;;
+esac
+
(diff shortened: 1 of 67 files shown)


qucs/autogen.sh
```diff
@@ -31,6 +31,8 @@ echo "done."
 echo -n "Creating config.h.in... "
 autoheader
 echo "done."
+echo "Running libtoolize"
+libtoolize || exit 1
```
guitorri: I did not read any of the code, but this should fix the OS X  build:
https://github.com/Qucs/qucs/blob/master/qucs-core/bootstrap.sh#L53-L58

felix-salfelder:
now it calls glibtoolize on `*Darwin*`. also, i accidentally rebased to master and force-pushed.

felix-salfelder:
i managed to retrigger travis. something went wrong during the first run.

i need these commits for the data sources (such as .dat parser rework and alternatives). could someone please merge (this does not change any functionality yet)?

in3otd:
I think the Travis issues are related to Sourceforge being still (partly) down. If needed, guitorri can manually restart Travis.

If you plan to continue working on this branch, would it make more sense to wait before merging it? You could still (re)base your other branches on this one, if you need the code here.

felix-salfelder:
some other changes (and not only mine) require the Simulator class introduced here.

there's nothing else that should really go into this branch, but i can certainly do some more polish.
give me a few hours...

felix-salfelder:
part of refactor branch

Issue #320, opened 2015-07-20
Export as image does not include the schematic Frame.


guitorri:
In `File > Document Settings` one can add a `Frame` border to the schematic.
Using the `File > Export as Image` the `Frame` is not present in the exported image.

ra3xdh:
See #324 for fix

Issue #319, opened 2015-07-20
Component dialog move up/down fail to move row.


guitorri:
The move up/down on a Equation component only move items on the first column. It should move the complete row.

This messes up the left-hand-side and right-hand-side of the equations.

![snap 2015-07-20 at 11 22 11](https://cloud.githubusercontent.com/assets/1388949/8773337/c17173fc-2ed3-11e5-8d7c-28f97a0cf3fc.png)
![snap 2015-07-20 at 11 22 25](https://cloud.githubusercontent.com/assets/1388949/8773338/c173b7fc-2ed3-11e5-9992-3cda70b4ef9d.png)

felix-salfelder:
is there any indication that this is related to the marker move cleanup? or am i just being paranoid?

guitorri:
:smile: Don't worry... this is related to `componentdialog`, the `slotButtUp` and `slotButtDown` to be specific.
Several dialogs had to be tweaked/ported to fit with Qt4, bugs were introduced in the process I think I was the last to mess with this dialog... I opened the ticked so we don't forget about it.

in3otd:
yep, as guitorri said, I've just checked, it was broken even before the marker move cleanup. (I did not have the time to bisect to find out the bad commit). So you can relax :smile:

guitorri:
I will take the blame on this again... 84f51e42

ra3xdh:
See #324 for fix

PR #324, opened 2015-07-21
Fix for #319 and #320

ra3xdh:
This PR is fix for both #319 and #320 .  Now Up/Down buttons work properly and Frame can be exported to image with Schematic. Please test on your side. Also dropped unused method `Schematic::getSchWidthAndHeight` .

### DIFF
--- a/qucs/qucs/components/componentdialog.cpp
+++ b/qucs/qucs/components/componentdialog.cpp
@@ -1218,6 +1218,11 @@ void ComponentDialog::slotButtUp()
   QTableWidgetItem *target = prop->takeItem(curRow-1,0);
   prop->setItem(curRow-1, 0, source);
   prop->setItem(curRow, 0, target);
+  source = prop->takeItem(curRow  ,1);
+  target = prop->takeItem(curRow-1,1);
+  prop->setItem(curRow-1, 1, source);
+  prop->setItem(curRow, 1, target);
+
 
   // select moved row
   prop->selectRow(curRow-1);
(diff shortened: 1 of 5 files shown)


felix-salfelder:
could you do this the other way around? please fix the "unused method" (which actually was not unused) instead of repeating the code twice?

ra3xdh:
@felix-salfelder , Done as you wish. Duplicate code moved to method.

ra3xdh:
Something strange happened today with Travis. It cannot even start MacOS build. It seems Sourceforge still be down.

guitorri:
I am trying to fix Travis. I is timing out during the Qt install on the OS X...

felix-salfelder:
are you sure that `Schematic::getSchWidthAndHeight` was of absolutely no use? it did a lot more than what you added now. looks like bounding box computation...

apart from that: looks good. please squash the commits before merge, and write a commit message. consider to include an answer to my question please.

ra3xdh:
Squashed commits in one. Indeed `getSchWidthAndHeight()` method was used only for bounding box calculation in `ImageWriter`. You can proof it using "Search of usage" feature of IDEs.

guitorri:
Previously it would take the bounding box of whatever was in the schematic (comps, wires...) now it only takes the size of the frame. Is that right? What if the frame is not set?

ra3xdh:
Yes, all is right. It was tested for schematics with and without frame. All could be exported normally. Here is an example without frame:
export
And with frame:
export-fr

guitorri:
Thanks!

in3otd:
...too fast @guitorri, IMHO.
As expected, now the exported image is limited to the frame boundaries, if a frame is present. But if there are components outside the frame, those will not be printed.
Sure, it's normally a user error to have components placed outside a frame, but I do not see why we should silently ignore those components. When we do a "View all" in the schematic, all the components are shown, inside or outside a frame and this is of course the Right behavior. Same for netlisting, we do not care if a frame is there, we netlist all the components, as the schematic is composed of all the components, inside or outside the frame, which is just a graphical decoration.

So I will strongly suggest to have all components printed as shown on the screen, inside or outside a frame. If the user has mistakenly placed components outside a frame, this mistake should be shown in the printed output.
If we would like to be even more flexible, add a "limit print to frame boundaries" option, but this should not be the default behavior.

felix-salfelder:
@in3otd thank you! i couln't find the place where the bounding box computation is actually required...

>  I will strongly suggest to have all components printed

seconded

ra3xdh:
See #331 for fix

PR #331, opened 2015-07-23
Continuing fixing #320

ra3xdh:
Enabled printing of components that fall out of the frame. 
export

### DIFF
--- a/qucs/qucs/imagewriter.cpp
+++ b/qucs/qucs/imagewriter.cpp
@@ -286,9 +286,11 @@ void ImageWriter::getSchWidthAndHeight(Schematic *sch, int &w, int &h, int &xmin
 
     int f_w, f_h;
     if (sch->sizeOfFrame(f_w,f_h)) {
-        xmin = 0;
-        ymin = 0;
-        w = f_w;
-        h = f_h;
+        xmin = std::min(0,xmin); // For components
+        ymin = std::min(0,ymin); // that fall out of frame
+        w = abs(std::max(f_w,sch->UsedX2) - xmin);
+        h = abs(std::max(f_h,sch->UsedY2) - ymin);
+        //w = std::max(w,f_w);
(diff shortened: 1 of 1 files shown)


in3otd:
uhm, this does not print the entire frame in case the circuit goes out of the frame on the left side... the circuit will be printed in its entirety, but the frame will be cut on the right side.

What about
``` C++
void ImageWriter::getSchWidthAndHeight(Schematic *sch, int &w, int &h, int &xmin, int &ymin)
{
    xmin = sch->UsedX1;
    ymin = sch->UsedY1;
    int xmax = sch->UsedX2;
    int ymax = sch->UsedY2;

    int f_w, f_h;
    if (sch->sizeOfFrame(f_w,f_h)) {
        xmin = std::min(0,xmin); // For components
        ymin = std::min(0,ymin); // that fall out of frame
        xmax = std::max(xmax,f_w);
        ymax = std::max(ymax,f_h);
    }

    w = abs(xmax - xmin);
    h = abs(ymax - ymin);
}
```
so the overall width and height are computed from the min and max extent of the schematic and the frame?

ra3xdh:
@in3otd Are you sure that you didn't use outdated version? I have just testes my code. Schematic is not cut in all directions. There are no lost components. Here is an example. 
export

in3otd:
I pulled the code for testing yesterday evening and it agreed with the patch content shown above f7f89fc76bb99c26e2742cedb473a6d4785b63ba .
Can you please test with a **small** schematic which goes out of the frame on the left side. Just a single wire going over the frame border on the left side will do. Looking at the code, for this case a width equal to the frame width will be used for printing, but the origin will start from the leftmost point of the schematic, so the right part of the frame will be cut.

ra3xdh:
Now it should work. @in3otd , please check on your side. Here is an example:
export

in3otd:
works fine now, thanks.

PR #332, opened 2015-07-25
Gui improvements

nvdl:
Work in progress but can be reviewed. I picked up the cherries.


felix-salfelder:
haven't looked at the code, but may i remind you of the 50/72 rule?

nvdl:
The are many violations as it was the start. It has been lying dormant for a while.
So yes, I keep the rule in mind now (except for the blank after the header).
Eventually, it will be rebased with a final message.

nvdl:
@in3otd Did you find time to give it a look?
I have been occupied with other things. Looks like, I have to split it into many PRs and issue them one by one.

in3otd:
> @in3otd Did you find time to give it a look?
> I have been occupied with other things. 

well, no, Same lack of time here, until recently.

> Looks like, I have to split it into many PRs and issue them one by one.

yes, very likely some splitting is needed. IIRC you had some real bug fixing (components "ghosting" while moving, etc.) and some "personal preferences" (new icons, variables names change, text cAsE changes, etc.) mixed in. I'll suggest to start with the former...

nvdl:
Yes, I will keep my personal stuff out of the invoked PRs or at least it will be discussed if needed by others.

PR #339, opened 2015-08-04
Matching tool

andresmmera:
Hello all, 

I have added some methods to the matching tool. Should I 'squash' the old commits?


in3otd:
Thanks for the matching tool additions. Somehow you have included commits from  others which were already merged in the master branch. To clean this up, I suggest the following steps:

- be sure you are on your local master branch:   
`git checkout master`   
- be sure you have your local master branch in sync with the remote one (I assume you called the remote Qucs repository `upstream`, following the convention in the [Guidelines] (https://github.com/Qucs/qucs/wiki/Contribution#get-a-copy-of-the-source-code), otherwise use the appropriate remote name)  
`git pull upstream master`
- rename your `MatchingTool` branch:  
`git branch -m MatchingTool MatchingTool_old`
- make a new MatchingTool branch and check it out:  
`git checkout -b MatchingTool`
- copy only your commits from the old branch to the new one:  
`git cherry-pick 45247ee`  
`git cherry-pick f76acee`  
`git cherry-pick d2d92ff`  

now you should have a `MatchingTool` branch which starts from the current master and includes only your commits.
Later you could squash your commits, but it's not so urgent.

Other random stuff I saw:
- the constant 3e8 should likely be 299792458.0 (no reason to approximate). Then this and `Z_FIELD` should go in a common header, one day, since these are used in several places in the Qucs tools. Can remain like this for the moment
- there are some strange characters at the very beginning of `matchdialog.cpp` (but the compiler does not complain ?)
- there are a lot of warnings generated by the compiler on the files you modified, please get rid of the ones related to the code you added and, if you feel confident, correct also the ones related to the deprecated Qt4 functions you copied from elsewhere.
- there is some code duplication, in particular for the functions related to 2-port matching. You can reuse `calcBiMatch()` to compute the input and output reflection coefficients your network should present and move the call to `calcMatching()` for the LC case outside of it
- also e.g. the stub calculations are often duplicated, just a couple of functions like `compute_stub(Zl, Z0, ..., t)` to be used everywhere should be enough ?
- the fact that when one port matching is selected, the schematic includes the S parameter simulation block is a different behavior than for LC sections, and there is no reason why it should be different. Either include a simulation block always, or never or allow the user to choose (I prefer one of the last two)
- small stuff, but maybe you could define a function using an [alternative method] (https://en.wikipedia.org/wiki/Binomial_coefficient#Computing_the_value_of_binomial_coefficients) for computing the binomial coefficients 
- it will be better to avoid using tabs for indenting, since almost all the code is formatted with spaces: I guess you have set very small tabs, but here I have 8-spaces tabs and following the body of an 'if' indented by 32 spaces is not always easy...
- more comments will follow in the (near) future :grin:

andresmmera:
Thanks for helping me with Github, Claudio. It is still a little bit messy
for me, but I hope to improve my Github skills soon.

I completely agree with your suggestions so I'll try to modify the code in
the next few days :)

I don't understand what you meant when you say that there are strange
characters at the beginning of "matchdialog.cpp". I've just checked that
file and there are nothing but comments (marked as /* */) and then the
#include<> lines.

Thank you again for your help.

2015-08-04 23:00 GMT+02:00 in3otd <notifications@github.com>:

> Thanks for the matching tool additions. Somehow you have included commits
> from others which were already merged in the master branch. To clean this
> up, I suggest the following steps:
>
>    - be sure you are on your local master branch:
>    git checkout master
>    - be sure you have your local master branch in sync with the remote
>    one (I assume you called the remote Qucs repository upstream,
>    following the convention in the Guidelines
>    <https://github.com/Qucs/qucs/wiki/Contribution#get-a-copy-of-the-source-code>,
>    otherwise use the appropriate remote name)
>    git pull upstream master
>    - rename your MatchingTool branch:
>    git branch -m MatchingTool MatchingTool_old
>    - make a new MatchingTool branch and check it out:
>    git checkout -b MatchingTool
>    - copy only your commits from the old branch to the new one:
>    git cherry-pick 45247ee
>    git cherry-pick f76acee
>    git cherry-pick d2d92ff
>
> now you should have a MatchingTool branch which starts from the current
> master and includes only your commits.
> Later you could squash your commits, but it's not so urgent.
>
> Other random stuff I saw:
>
>    - the constant 3e8 should likely be 299792458.0 (no reason to
>    approximate). Then this and Z_FIELD should go in a common header, one
>    day, since these are used in several places in the Qucs tools. Can remain
>    like this for the moment
>    - there are some strange characters at the very beginning of
>    matchdialog.cpp (but the compiler does not complain ?)
>    - there are a lot of warnings generated by the compiler on the files
>    you modified, please get rid of the ones related to the code you added and,
>    if you feel confident, correct also the ones related to the deprecated Qt4
>    functions you copied from elsewhere.
>    - there is some code duplication, in particular for the functions
>    related to 2-port matching. You can reuse calcBiMatch() to compute the
>    input and output reflection coefficients your network should present and
>    move the call to calcMatching() for the LC case outside of it
>    - also e.g. the stub calculations are often duplicated, just a couple
>    of functions like compute_stub(Zl, Z0, ..., t) to be used everywhere
>    should be enough ?
>    - the fact that when one port matching is selected, the schematic
>    includes the S parameter simulation block is a different behavior than for
>    LC sections, and there is no reason why it should be different. Either
>    include a simulation block always, or never or allow the user to choose (I
>    prefer one of the last two)
>    - small stuff, but maybe you could define a function using an alternative
>    method
>    <https://en.wikipedia.org/wiki/Binomial_coefficient#Computing_the_value_of_binomial_coefficients>
>    for computing the binomial coefficients
>    - it will be better to avoid using tabs for indenting, since almost
>    all the code is formatted with spaces: I guess you have set very small
>    tabs, but here I have 8-spaces tabs and following the body of an 'if'
>    indented by 32 spaces is not always easy...
>    - more comments will follow in the (near) future [image: :grin:]
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Qucs/qucs/pull/339#issuecomment-127760274>.
>



-- 
Andrés Martínez Mera

in3otd:
Googling around I saw that the characters at the beginning of `matchdialog.cpp` are a [byte order mark] (https://en.wikipedia.org/wiki/Byte_order_mark), telling that your file is using `UTF-8`. Never heard of that before. If I look at the file in GitHub using the browser the characters are not visible, also in 'raw' mode. Your editor might hide/not display them. In linux, I see them with `less`, `more`, `vi`, `xemacs` but not `emacs`, with the default configs I have. `gitk` also shows them. They were introduce with [this commit] (https://github.com/andresmmera/qucs/commit/f76acee6d3fdf5eed495e970570b65d5d72c07c1#diff-9459c038344855d1be7af728739c4239L1). Maybe your editor is introducing them automatically, not sure we need it.
It seems to be the only file using that in `qucs`, searching using `grep -r --include \*.h --include \*.cpp -P "\xEF\xBB\xBF" .`

andresmmera:
I usually employ the QtCreator to code and 'matchdialog.cpp' was not an
exception. I've opened this file with more, vi, atom and gedit but I
couldn't see nothing, However, I can see the extra characters if I use a
binary editor (bless). Screenshot:
https://dl.dropboxusercontent.com/u/21294824/Screenshot%20from%202015-08-05%2021%3A35%3A32.png

Could it be possible that the QtCreator adds these bytes automatically?
I've opened other cpp files from other projects of mine and these
characters do not appear. So, I wonder if this could have happened as a
result of a wrong git command...

Should I remove the MatchingTool branch and start from scratch?

2015-08-05 21:23 GMT+02:00 in3otd <notifications@github.com>:

> Googling around I saw that the characters at the beginning of
> matchdialog.cpp are a byte order mark
> <https://en.wikipedia.org/wiki/Byte_order_mark>, telling that your file
> is using UTF-8. Never heard of that before. If I look at the file in
> GitHub using the browser the characters are not visible, also in 'raw'
> mode. Your editor might hide/not display them. In linux, I see them with
> less, more, vi, xemacs but not emacs, with the default configs I have.
> gitk also shows them. They were introduce with this commit
> <https://github.com/andresmmera/qucs/commit/f76acee6d3fdf5eed495e970570b65d5d72c07c1#diff-9459c038344855d1be7af728739c4239L1>.
> Maybe your editor is introducing them automatically, not sure we need it.
> It seems to be the only file using that in qucs, searching using grep -r
> --include \*.h --include \*.cpp -P "\xEF\xBB\xBF" .
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Qucs/qucs/pull/339#issuecomment-128117802>.
>



-- 
Andrés Martínez Mera

andresmmera:
By the way, I've also find that some extra lines were added to the file
imagewriter.cpp. I can assure you that I've never opened that file, so I
don't really know how this could happen. After that, I removed the extra
lines because the compiler returned an error.

2015-08-05 22:03 GMT+02:00 Andrés Martínez Mera <
andresmartinezmera@gmail.com>:

> I usually employ the QtCreator to code and 'matchdialog.cpp' was not an
> exception. I've opened this file with more, vi, atom and gedit but I
> couldn't see nothing, However, I can see the extra characters if I use a
> binary editor (bless). Screenshot:
> https://dl.dropboxusercontent.com/u/21294824/Screenshot%20from%202015-08-05%2021%3A35%3A32.png
>
> Could it be possible that the QtCreator adds these bytes automatically?
> I've opened other cpp files from other projects of mine and these
> characters do not appear. So, I wonder if this could have happened as a
> result of a wrong git command...
>
> Should I remove the MatchingTool branch and start from scratch?
>
> 2015-08-05 21:23 GMT+02:00 in3otd <notifications@github.com>:
>
>> Googling around I saw that the characters at the beginning of
>> matchdialog.cpp are a byte order mark
>> <https://en.wikipedia.org/wiki/Byte_order_mark>, telling that your file
>> is using UTF-8. Never heard of that before. If I look at the file in
>> GitHub using the browser the characters are not visible, also in 'raw'
>> mode. Your editor might hide/not display them. In linux, I see them with
>> less, more, vi, xemacs but not emacs, with the default configs I have.
>> gitk also shows them. They were introduce with this commit
>> <https://github.com/andresmmera/qucs/commit/f76acee6d3fdf5eed495e970570b65d5d72c07c1#diff-9459c038344855d1be7af728739c4239L1>.
>> Maybe your editor is introducing them automatically, not sure we need it.
>> It seems to be the only file using that in qucs, searching using grep -r
>> --include \*.h --include \*.cpp -P "\xEF\xBB\xBF" .
>>
>> —
>> Reply to this email directly or view it on GitHub
>> <https://github.com/Qucs/qucs/pull/339#issuecomment-128117802>.
>>
>
>
>
> --
> Andrés Martínez Mera
>



-- 
Andrés Martínez Mera

in3otd:
I do not know if QtCreator adds these 'byte order marks', I do not use it, but I think others do use it and, as said, I did not find other files with that byte sequence at the beginning.

`imagewriter.cpp` was modified in master by a recent commit, but you should have no compilation errors due to these modifications.

I think there is no real need to delete your MatchingTool branch and start from scratch, if you were able to isolate your commits as described above you can continue from there.

andresmmera:
@in3otd I've corrected the code according to your suggestions and I've just pushed it to my fork. 

1) The duplication was minimized. Specific functions were written to calculate the individual matching sections and they are call from one-port matching and two-ports matching functions. In this sense, I've tried to reuse the existing functions for one-port and two-port matching.
2) All warnings were wiped out, even those related to the Qt libraries (at least my compiler said it was clean) 
3) The binomial coefficients were calculated according to the multiplicative formula.
4) I've included a checkbox to let the user to include a S-param block or not.
5) The auto-indent tool of the QtCreator was used to change the old indentation. Thus, no more tab indents. (Yes, I'm prone to using tab indents... :) )
6) I've checked matchdialog.cpp with the binary text editor and now it weren't added strange bytes at the beginning.

I hope the code quality have been improved. If you see something else or if you have any other idea, please let me know

in3otd:
ok, now the branch contains your commits only, good.

The Appveyor build (Windows) fails because MinGW does not define `M_PI` (there was a discussion at #171 and related PR #190), you should use `pi`, which is defined in `qucs/main.h`.

I still have some warnings, like
```
matchdialog.cpp:413:47: warning: unused parameter 'on' [-Wunused-parameter]
void MatchDialog::slotSetMicrostripCheck(bool on)
                                              ^
matchdialog.cpp:472:48: warning: unused parameter 'Index' [-Wunused-parameter]
void MatchDialog::slotChangeMode_TopoCombo(int Index)
                                               ^
matchdialog.cpp:1393:17: warning: equality comparison with extraneous parentheses [-Wparentheses-equality]
        if ((XL == 0))Schematic +=   QString("<Line 180 -50 0 290 #000000 0 1>\n" "<Text 200 220 12 #000000 0 \"Load: %1  %2 @ %3 GHz\">\n" "</Painti...
             ~~~^~~~
matchdialog.cpp:1393:17: note: remove extraneous parentheses around the comparison to silence this warning
        if ((XL == 0))Schematic +=   QString("<Line 180 -50 0 290 #000000 0 1>\n" "<Text 200 220 12 #000000 0 \"Load: %1  %2 @ %3 GHz\">\n" "</Painti...
            ~   ^   ~
matchdialog.cpp:1393:17: note: use '=' to turn this equality comparison into an assignment
        if ((XL == 0))Schematic +=   QString("<Line 180 -50 0 290 #000000 0 1>\n" "<Text 200 220 12 #000000 0 \"Load: %1  %2 @ %3 GHz\">\n" "</Painti...
                ^~
                =
matchdialog.cpp:1482:1: warning: control may reach end of non-void function [-Wreturn-type]
}
```


Also, when I open the matching tool, select "Multistage binomial lambda/4" and press "Create", Qucs crashes with `Illegal instruction (core dumped)`, did not have the time to investigate further.

andresmmera:
@in3otd Something odd happened,... It seems that the code was not on the latest version (fortunately, I have another copy on local, just in case). Maybe, this could be the reason why it crashed.

Hum,... I cannot see these warnings, but I've tried to correct them anyway. To compile, I do 'sudo make install' but it doesn't show any warning. You can check the 'make' output here: https://dl.dropboxusercontent.com/u/21294824/log  Should I use other command to compile?

P.S.: I've added Chebyshev weighting to the lambda/4 cascaded sections. Please be aware that so far it only works for N<=3.

in3otd:
looking at the compile output, it seems you did not use the `--enable-maintainer-mode` flag with `configure`, so you did not have all the warnings enabled. Try to rerun `configure` with that option, recompile and check if there are any warnings left

andresmmera:
@in3otd Ok, according to your instructions, the compile output is: https://dl.dropboxusercontent.com/u/21294824/warnings There are a huge number of warnings, but none of them coming from 'matchdialog.cpp' nor 'matchdialog.h'

in3otd:
yep, most of the warnings are about deprecated Qt3 functions, which are being (should be) slowly cleaned up.

You did not compute the input frequency correctly in the functions
```
Create2Port_SingleStub_matching_Schematic()
Create2Port_DoubleStub_matching_Schematic()
Create2Port_BinomialQW_matching_Schematic()
```
as you need to take into account the selected unit, as done [here] (https://github.com/Qucs/qucs/blob/master/qucs/qucs/dialogs/matchdialog.cpp#L396-L397).

Besides, I have the feeling that the related code should be organized differently:
`slotButtCreate()` reads `Z1`, `Z2`, `Freq` and the other input parameters and pass them to the functions which actually compute the matching circuits. Why do the above called functions re-read these parameters from the dialog? They should receive them as input parameters.
I'll try to see if I can find a different way of organizing the calculations, the 1- and 2-port cases are handled differently, the first draws the circuit directly, the second saves the circuits in strings and calls the above function to actually draw them, so there is surely some duplication around.

Please try to write a comment before every function to explain what it does, even if it is sometimes obvious today, it won't be re-reading the code in a few months or years. Ideally the function description should be picked up by Doxygen, so it will end up in the [auto-generated documentation] (http://qucs.github.io/qucs-doxygen/qucs/index.html).
Also, some functions parameters names are a bit "obfuscated", like in `calcSingleStub(double RL, double XL)` (and similar), where `RL` and `XL` that should be passed are **not** the real and imaginary part of an impedance, as their names seem to imply,  but the real and imaginary part of a reflection coefficient and later reused for the impedance. Do as [here] (https://github.com/Qucs/qucs/blob/master/qucs/qucs/dialogs/matchdialog.cpp#L472-L476), it is much clearer and there is no need to spare two local variables, that the compiler will likely optimize out anyway.

(besides, you left some tabs in the recent commits :grin:)

andresmmera:
@in3otd  

I've corrected these issues but I have some comments:
i) When I wrote the new functions, I thought that the code would be clearer if the number of variables passed to the functions were minimized. This is why the functions were re-reading the input values. Anyway, you're right, now it's much better :)

ii) Since the beginning, I was aware that using (RL, XL) instead of (r_real, r_imag) could be confusing. However, (RL, XL) become the real and the imaginary part of the load impedance after calling 'r2z(RL, XL, Z0)'. I've changed the code according to your suggestions but I'm still keeping (RL, XL) instead of (Zreal, Zimag) to do the rest of the calculations. Should I change them?

qucs/qucs/dialogs/matchdialog.h
```diff
@@ -20,16 +20,51 @@
 
 #include <QDialog>
 #include <QVBoxLayout>
+#include <QGroupBox>
 #include <QLabel>
+#include <math.h>
```
guitorri: We are using <cmath> throughout.

in3otd:
I'll take a look at this; there are some conflicts to merge in `develop` due to the previously-merged 
#482.
If that's ok, I'll the force push the rebased and fixed version to this branch.

ra3xdh:
I have checked this and noticed no problems too. @in3otd , You may clone an original branch, rebase it and resolve conflicts and open another PR from youself. Such PRs even may be self-merged, because they contains only commits of other people. I think it is lesser  evil than force pushing to the original branch in author's repository.

in3otd:
ok, will open another PR with the rebased PR. I had a hard time resolving conflicts also because at first I tried to keep most of the original commits, without squashing too much. But there were too many differences, due also to whitespace changes, so in the end I squashed and then rebased the resulting single commit.

guitorri:
This PR is now #666.

PR #340, opened 2015-08-14
A "What's new" window (WIP)

in3otd:
here is a customary "What's new" window that might be useful to inform the (new) users about the most important changes.
As usual, the window is automatically shown at startup (if the current Qucs version is newer than the last version used) and the user can choose not to see the "What's new" window again. It remains available in the 'Help' menu anyway.
I label this as a Work In Progress since
- the actual text to show still needs to be written (not an easy task...)
- I think that the window appearance and design could be improved

qucs_whatsnew

and, as usual, while writing this I (had to) change other stuff; I added a `SemVer` class to handle the comparison between version strings, since there was nothing already that could tell "is version 0.0.18 earlier than 0.0.19 or not?".
The existing `misc::checkVersion()` handled only one kind of comparison ('greather than') and also did not handle major versions (important if we will ever release `Qucs 1.0.0`, one day :grin:)

### DIFF
--- a/qucs/qucs-help/docs/en/Makefile.am
+++ b/qucs/qucs-help/docs/en/Makefile.am
@@ -24,7 +24,7 @@
 
 HTML = index.html short.html start.html mathfunc.html subcircuit.html       \
   programs.html internal.html characters.html matching.html start_digi.html \
-  start_opt.html octave.html
+  start_opt.html octave.html whatsnew.html
 PIC  = qucsmain.png paste.png wire.png select.png contab.png subcircuit.png \
   qucsdigi.png optimization1.png optimization2.png optimization3.png        \
   optimization4.png optimization5.png optimization6.png optimization7.png   \
--- a/qucs/qucs-help/docs/en/whatsnew.html
+++ b/qucs/qucs-help/docs/en/whatsnew.html
@@ -1,0 +1,107 @@
+<h1>Qucs 0.0.19 Release Notes</h1>
+
+<p>This release brings improvements on many parts of Qucs.
+The port from Qt3 to Qt4 is still ongoing and many bugs and usability issues have been fixed in the process.
+A few new features were added and others are in development.
+See below a short summary of the enhancements.</p>
+
+<h2>Qucs</h2>
+
+<ul>
+<li>Fix diagram color picker button (<a href="https://github.com/Qucs/qucs/issues/30">#30</a>)</li>
+<li>Fix zoom in a box (<a href="https://github.com/Qucs/qucs/issues/31">#31</a>)</li>
+<li>[NEW] Command line to save schematic as image (png, svg, pdf) (<a href="https://github.com/Qucs/qucs/issues/32">#32</a>)</li>
+<li>[NEW] Add component MutualX (<a href="https://github.com/Qucs/qucs/issues/54">#54</a>)</li>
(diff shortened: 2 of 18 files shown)


felix-salfelder:
what does `Sem` in `SemVer` mean? something like `VersionTriplet` sounds more straightforward (ymmv).

in3otd:
...it was a reference to [Semantic Versioning] (http://semver.org/); even if not supporting all the possibilities of the "official" semantic versioning, it seemed more or less appropriate. But I'm of course open to more descriptive names :grin:

(BTW, this reminds me I think I saw other places in the code where Qucs versions are/should be checked, I'll try to take a look at that...later)

felix-salfelder:
okay, i didn't know about "semantic versioning". from the article

    Software using Semantic Versioning MUST declare a public API. This API could be declared in the code itself or exist strictly in documentation. However it is done, it should be precise and comprehensive.

does QUCS declare a (nonempty) public API? imo it's too early to address versioning issues. we should add your version triplet comparison implementation regardless of the semantics. it makes perfect sense for the "do not show again" checkbox.

there is a `NEWS` file in the QUCS top source directory. it seems to coincide with what the new dialog should display. we should install `NEWS` instead of the independent `whatsnew.html` (or at least generate the latter from `NEWS`).

guitorri:
Originally each sub-directory (`qucs`, `qucs-core`, `qucs-doc`) had a NEWS file. I created a another `qucs/NEWS.md` file. The release script flattens the directories to create the tarball (`qucs` becomes the top dir). I though it made sense to have a single NEWS reflecting all the changes to the project.

I was working on the new NEWS. I am short of time right now... but the latest version is in here:
https://gist.github.com/guitorri/5ce296f510980ec64a45
See the _sample one, which is the result of the cross-referencing script.
Using the Milestones and Tags help a lot on sorting out the contributions.

I am not sure if we should continue with the top NEWS... but I agree that it would make it out look nice on the 'Whats new' dialog.

in3otd:
yep, not sure the "must declare a public API" makes sense for (every) application. But `VersionTriplet` sounds  good enough and does not suggest compliance with Semantic Versioning...

Generating the "What's new" from a `NEWS` file makes sense of course, I thought to use directly that file at the beginning, but then wanted to be able to use some nice formatting with html and did not want to link an external library just to parse one markdown file. Can maybe just translate the markdown to html when compiling/preparing for a release.  
Note that I put the `whatsnew.html` in `qucs/qucs-help/docs/en/` thinking that it could be translated in other languages. Not sure how often that will happen (but looking at the more or less regular contributors we should be able to have at least 4 other languages, besides English), can likely be elsewhere if we do not plan to have it translated.

in3otd:
...changed the name of the version string stuff class as suggested and updated the "What's new" message with the text from guitorri, converted to html with one of the many online converters.

guitorri:
This will be affected by the removal of qucs-help.

in3otd:
> This will be affected by the removal of qucs-help

due to the conflicts you mean? I quickly went thru the code (did not remember what I wrote) and did not see a direct connection with `qucs-help`. I can rebase and clean up for merging, if you think that makes sense.
But as I wrote above, it would be better to have the NEWS file markdown translated automatically to `html` when compiling/preparing for a release, so the *What's new* window text is always up to date.

guitorri:
You have the `whatsnew.html` inside `qucs-help/`.
I thing we should merge #652 before you rebase.

PR #343, opened 2015-08-24
Extended system OpAmps library.

ra3xdh:
I would like to present a work of my apprentice Timofey Moiseev. We prepared a massive extension of system Qucs OpAmps library. Added the following popular ICs models:
- LM358
- LM324
- LM311
- LF411
- NE5532
- NE5534
- AD822 family

Models are converted from datasheet SPICE models provided by electronic components manufacturers. I think this library extension will be useful for upcoming Qucs release. Here is an example test circuit: 
test

### DIFF
--- a/qucs/qucs-lib/library/OpAmps.lib
+++ b/qucs/qucs-lib/library/OpAmps.lib
@@ -1124,3 +1124,1057 @@ Sub:X1 _net0 _net1 _net4 _net3 _net2 _net5 gnd Type="lm3886_mod"
   <Line 13 3 0 -6 #000000 1 1>
   </Symbol>
 </Component>
+
+<Component LM311>
+  <Description>
+Authors: Vadim Kuznetsov <ra3xdh@gmail.com> and Timofey Moiseev <voig1396@gmail.com>
+  </Description>
+  <Model>
+.Def:OpAmps_LM311 _net0 _net1 _net2 _net3 _net5 _net4
+Sub:X1 _net3 _net0 _net1 _net2 _net4 _net5 gnd Type="LM311_cir"
+.Def:End
+.Def:LM311_cir _net4 _net1 _net2 _net3 _net6 _net5 _ref
+  .Def:LM311 _ref _net1 _net2 _net3 _net4 _net5 _net6
+  CCCS:F1 _net10 _net9 _net3 _cnet0 G="1"
+  Idc:IEE _net7 _net3 I="0.0001"
+  Vdc:VI1 _net21 _net1 U="0.45"
+  Vdc:VI2 _net22 _net2 U="0.45"
+  BJT:Q1 _net21 _net9 _net7 _ref Type="pnp" Is="8e-16" Bf="500" Nf="1" Nr="1" Ikf="0" Ikr="0" Vaf="0" Var="0" Ise="0" Ne="1.5" Isc="0" Nc="2" Br="1" Rbm="0" Irb="0" Cje="0" Vje="0.75" Mje="0.33" Cjc="0" Vjc="0.75" Mjc="0.33" Xcjc="1" Cjs="0" Vjs="0.75" Mjs="0" Fc="0.5" Vtf="0" Tf="0" Xtf="0" Itf="0" Tr="0"
+  BJT:Q2 _net22 _net8 _net7 _ref Type="pnp" Is="8e-16" Bf="500" Nf="1" Nr="1" Ikf="0" Ikr="0" Vaf="0" Var="0" Ise="0" Ne="1.5" Isc="0" Nc="2" Br="1" Rbm="0" Irb="0" Cje="0" Vje="0.75" Mje="0.33" Cjc="0" Vjc="0.75" Mjc="0.33" Xcjc="1" Cjs="0" Vjs="0.75" Mjs="0" Fc="0.5" Vtf="0" Tf="0" Xtf="0" Itf="0" Tr="0"
+  BJT:Q3 _net8 _net9 _net4 _ref Type="npn" Is="8e-16" Bf="1000" Cjc="1e-15" Tr="1.025e-07" Nf="1" Nr="1" Ikf="0" Ikr="0" Vaf="0" Var="0" Ise="0" Ne="1.5" Isc="0" Nc="2" Br="1" Rbm="0" Irb="0" Cje="0" Vje="0.75" Mje="0.33" Vjc="0.75" Mjc="0.33" Xcjc="1" Cjs="0" Vjs="0.75" Mjs="0" Fc="0.5" Vtf="0" Tf="0" Xtf="0" Itf="0"
+  BJT:Q4 _net8 _net8 _net4 _ref Type="npn" Is="8e-16" Bf="1002" Nf="1" Nr="1" Ikf="0" Ikr="0" Vaf="0" Var="0" Ise="0" Ne="1.5" Isc="0" Nc="2" Br="1" Rbm="0" Irb="0" Cje="0" Vje="0.75" Mje="0.33" Cjc="0" Vjc="0.75" Mjc="0.33" Xcjc="1" Cjs="0" Vjs="0.75" Mjs="0" Fc="0.5" Vtf="0" Tf="0" Xtf="0" Itf="0" Tr="0"
+  VCVS:E1 _net9 _net10 _net6 _net4 G="1"
+  Vdc:V1 _cnet0 _net11 U="0"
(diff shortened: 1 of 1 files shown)


felix-salfelder:
these OPAMP models appear to be netlist representations. what is the licensing and copyright situation? in any case it would be advantageous to remove dubious (non-free) cases from the main package, rather than adding more.

(otherwise, this looks very useful)

ra3xdh:
@felix-salfelder , I have no information on licensing of ICs SPICE-models provided by manufacturers. Maybe these models were developed when open-source software was not widely used and licensing was not taken into account.

ra3xdh:
Only AnalogDevices provides license agreement for SPICE models. See http://www.analog.com/Analog_Root/static/techSupport/designTools/spiceModels/license/spice_general.html for more details. I think other models terms of use are similar. From these license agreement follows that model can be freely distributed with any software.

felix-salfelder:
> @felix-salfelder , I have no information on licensing of ICs SPICE-models provided by manufacturers.

maybe, we should cross fingers here (IANAL)? in any case you should add a fair reference to the source. `Author: your_names` is certainly misleading. how about

    Obtained from: where you got this from
    Imported by: your_names

> Maybe these models were developed when open-source software was not widely used and licensing was not taken into account.

this is not about open source. c.f. the wiki/Contributing, it says

"IMPORTANT: By submitting your contribution, you agree to license your code under the GPLv2+ license used by Qucs"

note the GPL part. the question is whether you are permitted to license the netlists accordingly. this is no problem for textbook examples, but maybe for nontrivial commercial stuff.

felix-salfelder:
> From these license agreement follows that model can be freely distributed with any software.

this is really surprising, after reading the license agreement you found. it says

"You may not sell, loan, rent, lease, or license the SPICE Model, in whole, in part, or in modified form, to anyone outside Your company. You may modify this SPICE Model to suit Your specific applications, and You may make copies of this SPICE Model for use within Your company."

in particular, it says "You may *not* license the Model, in modified form". what am I missing?

FWIW: this reminds me of "werner2101/spicelib". this project tries (tried?) to work around the pita imposed by the licensing situation of spice models.

ra3xdh:
@felix-salfelder , I'm sorry. I have read AD's license agreement superficially. In such case we have to remove at least AD's model, because its distribution is forbidden. It's need to introduce a some way to include such models. I think it could be done similar as for patented codecs (MP3, etc.) in Linux distributions. It's need to provide compilation time-flag or to introduce model-dowloading client in Qucs.   

Also I will modify `Author` line as you suggested.

felix-salfelder:
> In such case we have to remove at least AD's models, because its distribution is forbidden.

i suspect, the same holds for the other models.

> It's need to introduce a some way to include such models.

as it is illegal to share these models, we need a program to
* download tarballs
* extract models
* filter out/replace SPICE peculiarities
* figure out symbols and names

this is essentially, what werner2101/spicelib does (or did). imo, this makes a good start.

somehow, we need to make qucsator read in spice models, or add a qucsator backend to spicelib. other simulators might have no problems to read in the models directly.

> Also I will modify `Author` line as you suggested.

i'm afraid this might not help a lot.

timofonic:
Hello.

Sorry to interrupt the experts and "hijack" this discussion in some way.
I'm just a newbie electronics student and mere long time Linux user. I'm
very interested in QUCS and KiCad as both learning and design tools, so I
like to follow their development very closely.

This came into my mind and needed to write it, to calm my restless
curiosity:

- What's the perverse reason of these so restrictive licenses for SPICE
models? I don't get it!

- Wouldn't be better for long term goal to recreate these models and avoid
to use these restrictive licenses? Is it so difficult to redo them to have
same or better accuracy but not breaking licenses?

-If it's so difficult:

* Would it be possible to do this effort collaboratively between different
simulator projects? And what about some way to stimulate this effort to be
made, motivating to contribute to the development of a truly Open Library
for circuit simulation?

* Would it be considered as a long term goal? Wouldn't needing to download
tarballs and such make he situation even worse than MP3/H264 and certain
Microsoft Windows DLLs to and to run certain applications under Wine?

Kind regards.

felix-salfelder:
>  What's the perverse reason of these so restrictive licenses for SPICE models? I don't get it!

this is one perspective. why do people buy or use stuff that is patented/copyrighted/licensed and legally irreproducible? (what are the alternatives?)

> Wouldn't be better for long term goal to recreate these models and avoid to use these restrictive licenses? Is it so difficult to redo them to have same or better accuracy but not breaking licenses?

what does "recreate" mean? whichever way you create a netlist of a LMxyz opamp, the netlist will be copyrighted by its (original) authors. i am not sure if you are allowed to distribute a behavioural model of an opamp parameterized to fit some LMxyz. but even if, such a model will likely fail to replace the netlist model in practice.

guitorri:
This library would be very useful indeed.

@ra3xdh you said the models where created from the datasheet? In that case I believe it should be OK. Validation of the model is the next roadblock. We have to make sure the user can trust the models.

If the models are derived from non-free SPICE/PSPICE netlists it might indeed be a problem. I haven't read the license very carefully.

I wonder how the existing models in qucs-lib made into the codebase. There is little info about the source, or how the models where derived or implemented. 

At some point we should check and purge non-free code from the project repository...

In many cases qucsconv does a fair job in importing SPICE to qucslib format. Maybe going in a similar way to werner2101/spicelib might be an option in case we are not free do distribute and modify such files.

timofonic:
Hello.

Maybe my opinion is totally flawed and lack fundamentals. And maybe it
would be a very time consuming task.

Why just purge them and lose the functionality of having those components?
What about doing something similar to the audit and refactoring method like
the one used in ReactOS[1]?

Aren't really some legal way to make copyleft models of these components?
Could clean-room reverse engineering be applied to component simulation
models? If yes, why not use that?

* [1] https://en.m.wikipedia.org/wiki/ReactOS#Internal_audit

Kind regards.

felix-salfelder:
> Could clean-room reverse engineering be applied to component simulation models?

this is again off topic, however appears to be a legitimate question. here's my attempt on an explanation, YMMV.

in my understanding, "clean-room" involves two disjoint sets of people. one set works out a specification of a system and the other set reimplements the specification, not knowing about the (protected/copyrighted) existing implementation(s). now, this will not work for analogue circuit components, because a specification of an analog component is generally too *weak* to permit a faithful reimplementation.

but lets try. say, you (as part of the first set) figure out that LMxyz is an opamp that dissipates P, has an output range of R and a gain of A. your engineers (the other set, ignorant of LMxyz) think of a shiny, free opamp design matching your specification. they pass you a netlist. now using that netlist, you find, that your circuit behaves strangely different when using the free opamp. you easily figure out why, the free one is much faster! so you add a slew rate obtained from the LMxyz to the specification and kindly ask your engineers to design another opamp. the new opamp netlist is much better, but you will still find discrepancies. you restart with an additional temperature coefficient...

usually you cannot expect that this will terminate (e.g. your engineers find the original netlist), because there are too many non-equivalent opamp netlists close to your specification (note that transistor device parameters are real valued). eventually you may be satisfied with one that is just *close enough*. but there's a catch: you simply cannot tell (what does "close enough" even mean, to begin with?). and then, even if it's accurate for you, it might be not precise enough for someone else, while convincing the others remains your problem.

in3otd:
FYI, there was a brief discussions about including Spice models from Analog Devices with someone from the company a [couple of years ago] (https://sourceforge.net/p/qucs/mailman/qucs-devel/thread/51B0F7AE.8070704%40gmail.com/#msg31008486) . He even said "to have permission from [the ADI Spice modeling group] to add them to open source simulators - like Qucs" but it seems he didn't actually add them. I see that at the time also @crobarcro and @fransschreuder were involved, don't know if they have more info on this. We could always try to contact him again.

Regarding the actual distribution of their model, the way I see it is that we would be allowed to distribute their models with Qucs ("You may include copies of Analog Devices’ SPICE models with any software you sell or distribute. However, you may not make changes to the redistributed copies of Analog Devices SPICE models other than to", etc.) *if we did not translate them*. We are not licensing the models, they are not covered by the GPL, as they are not part of the program code: Qucs can be compiled and run without them, they are just input to the program. Sure we should try to make this clear for the users, maybe show a message during install or the first program run?

guitorri:
Their license does not allow modification. I guess this is a similar issue with the [Accelera Verilog-A headers in ADMS](https://github.com/Qucs/ADMS/wiki/Header-license-issue) (I will turn this in a issue so we can discuss it further). 

We clearly **don't want to modify** reference files, but if we ship those files in a GPLed package the end user should be **allowed to modify** anything thy wish to... but I am not a lawyer...

We should ask for advice on debian-legal@lists.debian.org. 
Perhaps having a non-free package just for the models might be a way out. 

Something along the lines of https://github.com/werner2101/spicelib might be a solution. Sort of a package manager for models fetched on demand by the user.

felix-salfelder:
> Perhaps having a non-free package just for the models might be a way out. 

exactly. non-free models should be packaged separately. note that they need to be distrubutable still.

> Something along the lines of https://github.com/werner2101/spicelib might be a solution.

this is only required for models that are not legally distributable.

felix-salfelder:
now, there's a repo for this, qucs-nonfree

ra3xdh:
Yes, I will retarget this PR to `qucs-nonfree`. I am closing this.

Issue #347, opened 2015-09-03
Wrong result in impedance calculation for rectangular Waveguide with dielectric medium


StefanD986:
When calculating the impedance of a rectangular waveguide filled with a dielectric (Er != 1) the result is wrong. It looks like a factor of sqrt(Er) was forgotten in the equation: 

Example: given a rect. waveguide a=100mm, b=44.4mm filled with an Er=1.69 the wave impedance for the TE10 mode should be:
Z_F = (Z_0/sqrt(Er)) / sqrt(1 - lambda0 / lambdaCE)

with 
Z_0 = free space impedance (120Pi Ohm)
Er = 1.69
lambda0 = c_0/f
lambdaCE = lambdaC * sqrt(Er)
lambdaC = 2/sqrt( (m/a)^2 + (n/a)^2 ) = 2*a for the TE10 mode.

The expected result should be Z_F = 355 Ohm
Qucs / Qucstrans gives a value of Z_F = 461 Ohm for this example. I played around with the equation and it looks like that in the qucs source code Z_0 is not multiplied with sqrt(Er). 

Sources for the Z_F equation I used is "Taschenbuch der Hochfrequenztechnik, Band 2 - Meinke / Gundlach", Chapter K.4.3.
Side note to that: When the relative permeability is also taken into account I *assume* that the equation should be (but this case is not considered in my source):
Z_F = (Z_0*sqrt(Mur / Er)) / sqrt(1 - lambda0 / lambdaCE)    with lambdaCE = lambdaC * sqrt(Er / Mur)

For quick comparison this might be helpful: http://www.amanogawa.com/archive/RectWaveGuide/RectWaveGuide-2.html

I have tested this in qucs 0.0.18 qucstrans on OSX 10.10.5 (via homebrew). 
Example circuit:

```
<Qucs Schematic 0.0.18>
<Components>
<Pac P1 1 230 250 -74 -26 1 1 "1" 1 "50 Ohm" 1 "0 dBm" 0 "1 GHz" 0 "26.85" 0>
<Pac P2 1 410 250 18 -26 0 1 "2" 1 "50 Ohm" 1 "0 dBm" 0 "1 GHz" 0 "26.85" 0>
<GND * 1 230 280 0 0 0 0>
<GND * 1 410 280 0 0 0 0>
[...]

PR #349, opened 2015-09-07
Several fixes concerning rectangular waveguides

andresmmera:
As suggested [here] (https://github.com/Qucs/qucs/issues/347), it seems that the equation for calculating the wave impedance was wrong. According to Pozar's book (Page113, Equation 3.86):
Z_{TE} = k * eta / beta
where k is the wave number, beta is the propagation constant and eta is
the intrinsic impedance of the waveguide. By definition, eta=sqrt(mu/e) (where 'mu' is the permeability and 'e' the permittivity). However, transcalc calculated eta as eta = sqrt(mu_0/ e_0), so it worked only for e_r = 1 and mu_r = 1

I think that the same issue occurs at qucs-core/src/components/rectline.cpp

### DIFF
--- a/qucs-core/src/components/rectline.cpp
+++ b/qucs-core/src/components/rectline.cpp
@@ -183,7 +183,7 @@ void rectline::calcPropagation (nr_double_t frequency) {
     // resistive
     rs = std::sqrt (pi * frequency * mur * MU0 * rho);
     ac = rs * (2.0 * b * sqr (pi) + cubic (a) * sqr (k0)) /
-      (cubic (a) * b * beta * k0 *  Z0);
+      (cubic (a) * b * beta * k0 *  Z0  * std::sqrt(mur/er));
     alpha = (ad + ac);
 
     // wave impedance
--- a/qucs-core/src/components/rectline.cpp
+++ b/qucs-core/src/components/rectline.cpp
@@ -187,14 +187,14 @@ void rectline::calcPropagation (nr_double_t frequency) {
     alpha = (ad + ac);
 
     // wave impedance
-    zl = (k0 * Z0) / beta;
+    zl = (k0 * Z0  * std::sqrt(mur/er)) / beta;
 
   } else {
     /* according to [2] eq 3.207 */
     beta = 0;
     alpha = -std::sqrt (- (sqr (k0) - sqr (kc)));
     // wave impedance
-    zl = (k0 * Z0) / nr_complex_t (0, -alpha) ;
+    zl = (k0 * Z0 * std::sqrt(mur/er)) / nr_complex_t (0, -alpha) ;
   }
 }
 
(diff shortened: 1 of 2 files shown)


andresmmera:
I am checking the S-parameters of the simulation attached [here](https://github.com/Qucs/qucs/issues/347) with the ADS. Surprisingly, (or not) I found that Qucs and the ADS give slightly different values.

Frequency: 2 GHz
                                                                  
[ADS](https://dl.dropboxusercontent.com/u/21294824/Sparameters-WR.csv): S11  = 0.949 / 3.385 ,  S12 = 0.315 / 93.385 ,  S21= 0.315 / 93.385 ,  S22 = 0.949 / 3.385 
Qucs (this branch): S11 = 0.96/3.02, S12 =  0.281/93,  S21=0.281/93, S22 = 0.96/3.02
Qucs (master branch): S11 = 0.976/2.34, S12 =  0.218/92.3,  S21= 0.218/92.3,  S22 =  0.976/2.34     

So I'm wondering if there are more errors at 'rectline.cpp'... Maybe the ADS is wrong (but I think it is less probable)

in3otd:
The corrections seems fine to me.
The discrepancy w.r.t. that other simulator is because it uses a different definition of *characteristic impedance*, namely the so called power-voltage definition. Details on the different typical definitions of characteristic impedance for a rectangular waveguide are, e.g. [here] (https://books.google.com/books?id=iEvgmwH1esgC&lpg=SA3-PA26&pg=PA3-PA26).
The strange thing is that there is an article which contains a comparison of the Qucs waveguide model with the ADS model (see [here] (http://arxiv.org/pdf/1011.6363v2.pdf), pag. 6) and the results agreed , at the time the simulation was done. Looking at the change history, the Qucs model should not have changed since then and I'll guess also the ADS model did not change, so it's rather strange the results agreed. @bastien-roucaries, do you know more about that simulation? AFAIK, you initially wrote the RWG model in qucsator, did you do also the comparison in the article?

A nice thing will be to have the possibility to choose which impedance to compute, both for qucstrans and qucsator.

andresmmera:
@in3otd Be aware that I reported a small difference in natural units whereas the plot you've cited is in dB.
I'll try to reproduce that graph.

andresmmera:
Yep, the graph you've cited is correct. 
WR-22, 800mm long S21 [dB] from 30GHz to 50 GHz, 
selection_018

in3otd:
ok, seen like this the results seem almost identical, except maybe for the two lowest peaks. So likely the difference has always been there, but nobody really noticed...

in3otd:
Merged into release-0.0.19 (c3f7352a3c02de940910382b565cad3514177478)

guitorri:
Please tag it. I am using the tags to sort out the changelog.

PR #352, opened 2015-09-14
[travis] Enable schematic to netlist testing

guitorri:


guitorri:
I will merge this. We can continue discussing the `make dist` and testing elsewhere.

PR #355, opened 2015-09-14
Fix Qucsconv handling of hicumL2v2p31

guitorri:
This is related to converting SPICE model cards and instances into qucsator and qucslib formats.

Can be tested with something like:

```
echo ".MODEL  HICUM  NPN (LEVEL = 2  VERSION = 2.31)" >> model.sp
echo "Q1 1 2 3 HICUM" >> model.sp
qucsconv -i model.sp -if spice -of qucs
qucsconv -i model.sp -if spice -of qucslib
```


guitorri:
Rebased to resolve conflicts.

PR #346, opened 2015-08-31
Power combining

andresmmera:
This is a tool for synthesizing power combining networks. It has some basic structures (Wilkinson, branchline, Gysel, travelling wave,... ), but I hope to extend it in the future.

PR #356, opened 2015-09-15
Circular waveguide component

andresmmera:
This is a component for modelling the TE11 mode of a circular waveguide. The code is very similar to the rectangular waveguide model.

In order to check it works, [here](https://dl.dropboxusercontent.com/u/21294824/CircularWG.sch) is attached a schematic with a S-parameter simulation. The results seem to be coherent with [this Octave script](https://dl.dropboxusercontent.com/u/21294824/CWG.m), which implements the theoretical equations given by the theory.

Qucs schematic:
screenshot from 2015-09-15 18 18 46

Qucs results:
screenshot from 2015-09-15 18 19 42

S11 (Script)
s11

S21 (Script)
s21


qucs-core/build-aux/ar-lib
```diff
@@ -0,0 +1,270 @@
+#! /bin/sh
```
in3otd: I think this file is generated by the autotools and so it should not be committed.
We should likely add it to `.gitignore`

qucs-core/src/components/circline.cpp
```diff
@@ -0,0 +1,273 @@
+#if HAVE_CONFIG_H
```
in3otd: The usual GPL blurb at the beginning of the file is missing here

qucs-core/src/components/circline.h
```diff
@@ -0,0 +1,38 @@
+#ifndef __CIRCLINE_H__
```
in3otd: The usual GPL blurb at the beginning of the file is missing here

qucs/qucs/components/circline.cpp
```diff
@@ -0,0 +1,79 @@
+/***************************************************************************
+                              circline.cpp
+                             -----------------
+    copyright            : (C) 2008 by ROUCARIES Bastien
```
in3otd: ...did really Bastien write this 7 years ago ? :grin:

andresmmera:
@in3otd Thank you :-)
 I have just added the missing GPL headers. By the way, I didn't include this line:
 Copyright (C) [Year] [Name] [email address]
Should I do it?

in3otd:
uhm, AFAIU you are not required to add a copyright line, git keeps track of who did what, so one can find who is/are the copyright holder(s), if needed. Yet, we "usually" add these information in the header and I think it's a good thing to be able to know when the code was written (and by who).

andresmmera:
I've noticed that AppVeyor crashes because it takes [more than 40 min to build...](https://ci.appveyor.com/project/qucs/qucs/build/1.0.218/messages)
Is it possible to run the test again without pushing more commits?

The same issue also happens at https://github.com/Qucs/qucs/pull/349

guitorri:
Indeed, right now we have to download and install Qt on AppVeyor each time. Depending on download speed time add up and causing the build to time out. I made a request to install Qt4.8.7 on their side. Don't know if they will do. I will re-run AppVeyor to check if we get lucky on a fast download...

in3otd:
BTW @andresmmera , you know you can convert `.mat` files into Qucs datasets with `qucsconv` (or 'Project->Import/Export Data' in the GUI), so you can plot both sets of data on the same graph? Or you can have a GNU Octave script automatically launched at the end of the simulation ('File->Document Settings', which can read the Qucs dataset and plot/compare the results :grin:

guitorri:
Guys, if you can put aside a few nice looking schematics containing transmission line components I really appreciate. I am trying to increase the coverage, so far we have very few examples involving these components. This is an outdated [report](https://github.com/Qucs/qucs-test/blob/master/report_coverage_Qucsator_0.0.18.txt) from qucs-test.
See, qucs-test (one day) might serve the double purpose of examples and test-suite. Please let me know if you have something to share.

andresmmera:
@in3otd Well, I didn't know about that feature, so I have been playing a little bit with it :-) . I've followed your instructions and I found that the Qucs simulation does not match (exactly) the theoretical formulae.

Magnitude S11:
mags11

Angle S11:
angles11

Magnitude S21:
mags21

Angle S21:
angles21

Here are the [Qucs schematic](https://dl.dropboxusercontent.com/u/21294824/CircularWaveguide.sch) and the [octave script](https://dl.dropboxusercontent.com/u/21294824/CWG.m) to reproduce the plots above. I reckon that I'm missing something at the octave script. Perhaps, the debris are caused by the thermal model, not included in the script. I need to do some research about this issue... 

BTW, octave+Qucs are not working as duly on my side. It is not possible to resize nor close the window. Besides this, It cannot export the current plot and sometimes it gets stuck...

in3otd:
uhm, from a quick glance at the plot I'll say that the Qucs losses are higher at low frequency than the Octave ones. I didn't look at the Qucs and Octave sources, are they supposed to implement exactly the same model??

andresmmera:
@in3otd Ok, I've found the bug :-)
I missed the ./ operator when calculating the wave impedance at the Octave code. Besides this, the specific resistivity was different... Now Octave perfectly matches the Qucs results. It was expected because they implement exactly the same equations. I didn't upload the screenshots because it is not possible to appreciate any difference between the Qucs and the Octave traces.

In case you want to check the results, these are the sources: [Octave script](https://gist.github.com/andresmmera/0dd57ad5c0ec1dfbf43e) [Qucs schematic](https://gist.github.com/andresmmera/e9302a2610f6524d29b4)

@guitorri I have made a [tool] (https://github.com/Qucs/qucs/pull/346) for synthesizing power combiners with ideal transmission lines and microstrip lines.  I think it could be useful to build some examples.

andresmmera:
Ops! I closed (and reopened) this PR accidentally...

guitorri:
@andresmmera I did not have time to test your synthesizing tool. I am trying to cut a release. 
@in3otd can you help us here to see if this code and the synthesizer tool should be added now or wait a future release? 
I really hope we can reduce the time between releases... :smile:

ra3xdh:
@guitorri  I think we should not merge *any* new feature right before release. Even if it seems to be tested, there may be undiscovered bugs that could be infiltrated in packages. I suggest to delay merging until 0.0.19 will be released.

 Maybe it has sense to open a special issue where we can put final countdown before relase (what issues and PRs should be megred)?

in3otd:
@andresmmera good! You see, the integration of Qucs and Octave can be quite useful :grin:

@guitorri I understood from the burst of activity there was a release about tho happen :grin:
I know we are all short on time, but will be good to try to define a timeline. Until when we/you will merge new features? This weekend? The next one? We should then leave one or two weeks were we are supposed just to test the existing code before release without adding anything new.

ra3xdh:
@in3otd , Yes, deadline for new features is very good idea.

guitorri:
@andresmmera sorry to hijack the thread. 
I wonder if we shouldn't be making more use of the qucs-devel.

@ra3xdh, I also think a bit more testing is needed, that is why I passed the ball to you guys. :smile: 
We can use milestones to track the interned additions to a given release. See https://github.com/Qucs/qucs/milestones/0.0.19 
https://github.com/Qucs/qucs/milestones/0.0.20
So far I have been arbitrarily scheduling issues I think are doable and good to go and setting the deadline... which is already past due :)

@in3otd I was thinking about a snapshot. Hopefully this weekend (if the baby continues to sleep well). 

What a about a snapshot as soon as possible and a release by next weekend? Then we send a heads-up on qucs-devel to tell people the project is still alive... 
Please help tag the milestone on issues and PRs, so we can discuss/prioritize what gets where an when...

andresmmera:
@guitorri Maybe you misinterpreted my comment:  I reckon that https://github.com/Qucs/qucs/pull/346 is useful for generating examples, but I don't think it should be included in the current release. In fact, I couldn't agree more with @in3otd and @ra3xdh, it could be dangerous or harmful to merge new features without prior testing.

@in3otd Yep, Qucs+Octave is cool, but Octave gets stuck when managing the plots... and it is really frustrating... :D

andresmmera:
Well, I'm realizing that my last comment came late... :-)

ra3xdh:
@andresmmera I have tested your code. Your circular waveguide model seems to be correct. I think we can consider merge after release.

ra3xdh:
Closing, because rebased and merged in #654.

PR #358, opened 2015-09-16
make distcheck work in qucs-core.

felix-salfelder:
... almost work. still need
DISTCHECK_CONFIGURE_FLAGS=--enable-maintainer-mode\ --with-mkadms=admsXml

this fixes
- dont try to distribute nonexisting depcomp
- ADMS_SUBDIR (now taken care of by AC_SUBST)
- line continuation in *_HEADERS block
- some CPPFLAGS (add builddir where necessary)
- proper paths in test command
- complete CLEANFILES

### DIFF
--- a/qucs-core/Makefile.am
+++ b/qucs-core/Makefile.am
@@ -28,21 +28,9 @@ CODE_COVERAGE_LCOV_OPTIONS ?= --no-external --rc lcov_branch_coverage=1 $(CODE_C
 CODE_COVERAGE_GENHTML_OPTIONS ?= --rc lcov_branch_coverage=1
 @CODE_COVERAGE_RULES@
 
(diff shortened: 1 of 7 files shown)


guitorri:
Looks good to me! Merging now.

felix-salfelder:
ok. working on the same for `/qucs`.

there is no top level makefile/configure. Do you intend to distribute 2 tarballs? otherwise (or: in any case) a toplevel build-all would be handy... i think the build-bot should do a top-level `make distcheck`.

guitorri:
I guess it makes sense to have a top level configure. Doing so we don't have to move `qucs-core` and `qucs-doc` inside `~/git/qucs`. The current release script does that, and I find it a bit of a hack...
The tarball structure should look the same as the repo...

PR #361, opened 2015-09-17
make distcheck work in qucs, also fix more distcheck in qucs-core

felix-salfelder:
- fix broken .h->.moc.cpp rule
- add various _HEADERS
- distribute qucs.qrc
- add several qrc files
- add more filter headers
- sort out qm rule
- add translation files to dist


qucs/translations/Makefile.am
```diff
@@ -75,7 +77,7 @@ qucs_.ts: $(qucs_SOURCES)
 	  -locations relative -ts $(srcdir)/$$file; \
 	done
 
-.ts.qm:
+%.qm: %.ts
```
guitorri: Is this portable?
http://stackoverflow.com/questions/767922/portable-equivalent-of-gnu-make-style-pattern-rules
I recall someone using FreeBSD comented about it... perhaps not a big issue, just need to use gmake.

felix-salfelder:
the broken `a.b.` rule is unexpected (does it work on your side?). i do not really want to touch it. (yes, i fancy portable rules... need to investigate.)

right now, i consider distcheck much more important and useful.

felix-salfelder:
ok. looks better now.

how about qucs-doc? doesn't seem to build at all here. all i get is errors such as

    ! LaTeX Error: File `mszl' not found.

ever tried?

guitorri:
So, building qucs-doc is a bit convoluted... it depends on some Octave module (epsTk?) to create figures. You have to install epsTk and add it to the Octave path... Not so sure, let me see if I can find something in my notes. 

Perhaps we can add the final figures instead of building from scratch each time.

felix-salfelder:
> it depends on some Octave module (e it depends on some Octave module

good to know. this should better be checked by configure.

> Perhaps we can add the final figures instead of building from scratch each time.

this will add extra complexity. i prefer to provide a binary package for those, who do not want to build.

guitorri:
I mean, is it really necessary to run octave each time to generate the figure? Add the figure and be done with it. 

I am not sure what do you mean by 'binary package'.
Note that from a release tarball, the documentation is already build. A `./configure $$ make install` I only installs the PDFs. If the user messes with bootstrap/autogen then the whole latex, octave... needs to be there, pretty much like building from Git.

felix-salfelder:
> I mean, is it really necessary to run octave each time to generate the figure? Add the figure and be done with it. 

the figure is only generated if it's not there. this only happens once, not each time. it is (should be) present in the dist tarball (did I break this?), so no generation necessary there. this particularly makes sense, if octave is not available.

as a rule of thumb, generated files should *never* be commited to git. this will just silently break the path from source to binary some day.

a binary package is a tarball, which does not contain sources. something you get by

    make install DESTDIR=/tmp/mydist
    cd /tmp/mydist
    tar -cjxf dist.bz2 *

> If the user messes with bootstrap/autogen then the whole latex, octave... needs to be there, pretty much like building from Git.

no, (s)he will only need octave if (s)he changes the sources. now, if octave is not meant to be used, (s)he will get a warning about outdated files, not wrong images silently.

felix-salfelder:
rebased ..

felix-salfelder:
cannot find the place where appveyor fails. what am i supposed to search for?

(sorry, but the logfile is >500lines long)

in3otd:
...you have to go to the "Messages" tab in Appveyor to discover that "*Build execution time has reached the maximum allowed time for your plan (40 minutes).*"... So it "simply" timed out...

guitorri:
Note that Appveyor is building with CMake only (for now). It's sole purpose is to catch compilation errors with MinGW, which is used for building the Windows installer.

Appveryor may timeout because or Qt4. Qt4 is not installed on their machines and the binaries are downloaded each time... if the connection is slow... it can easily time out. I know, I know...

AppVeyour recently instaled MSYS2. With that it is easier to run Autotools on Windows. It also has all the packages we need (based on pacman)... I have it working locally. Need to find some time to update the .appveyor file. 

It is the same story with OSX on Travis. The Qt4 package supplied by Homebrew does not have Qt3Support in it. The official binary is pulled down each time. It is less often but it may also timeout. Another big waste of bandwidth. I could create myself a Homebrew package with Qt3Support, but I have to look into the details of which OSX they have on their virtual machines to see I can create compatible binary with my machine.

guitorri:
Travis running Clang (both Linux and OSX) is stalling at: 
`./qucsator -l > qucsdefs.h`

I probably has to to with this (which I don't see a good reason to keep it guarded as such): 
https://github.com/Qucs/qucs/blob/master/qucs-core/src/ucs.cpp#L106-L108

I guess the maintainer mode was defining the `DEBUG`. But I cannot see it been set for GCC, why it is not failing as well?! :confused:

felix-salfelder:
what is `qucsdefs.h` to begin with? is this for debugging or actually in use?
appears to be just another wet spot, https://en.wikipedia.org/wiki/Don't_Repeat_Yourself 

\me shrugs

guitorri:
`qucsdefs.h` is a dump of the registered components in `qucsator`. Devices attributes (name, number of pins,...) and properties (name, default value, range) are dumped in structs. The file was generated in maintainer mode to be used with `qucsconv`. The data in it is used to map across spice, qucslib and the qucsnetlist (you know what I mean). 
So, `qucsconv` depends on `qucsator` for the list of components descriptions. I remember this dependency from the time when I wrote the CMake files...

felix-salfelder:
> So, qucsconv depends on qucsator for the list of components descriptionshe list of components descriptions

which does not make a lot of sense. qucsconv and qucsator are built together, their sources could include the same component database. but i do not complain, i don't need qucsator, and i don't opt in to fix this `:)`.

do you agree with aa0bd71? there's hardly any extra code involved,

guitorri:
Sure, why not expose `-l` by default. Let us focus into the purpose of this PR... :wink: 

I am doing:
```
./bootstrap
./configure --disable-doc --with-admsxml=`which admsxml`
make -j8 
``` 
Sometimes I get this error in`qucs-core/src/converter`: https://gist.github.com/guitorri/b6d2309ec9256dde2463

So far it always completes if I run `make -j8` once again after the error.

Should `tokens_spice.h` be also a target? Or can we mark it as advanced somehow?

felix-salfelder:
> `--with-admsxml=`which admsxml`

this is redundant. if admsXml is there, it will be used. dead simple. but i will have to test the other cases, maybe add a version check etc.

> Should tokens_spice.h be also a target?

thanks. did not notice -- will have a look.

guitorri:
Any idea about this:

```
[felix-salfelder-distcheck]~/git/qucs $ make dist
/Applications/Xcode.app/Contents/Developer/usr/bin/make  dist-gzip am__post_remove_distdir='@:'
if test -d "qucs-0.0.19"; then find "qucs-0.0.19" -type d ! -perm -200 -exec chmod u+w {} ';' && rm -rf "qucs-0.0.19" || { sleep 5 && rm -rf "qucs-0.0.19"; }; else :; fi
test -d "qucs-0.0.19" || mkdir "qucs-0.0.19"
 (cd qucs && /Applications/Xcode.app/Contents/Developer/usr/bin/make  top_distdir=../qucs-0.0.19 distdir=../qucs-0.0.19/qucs \
     am__remove_distdir=: am__skip_length_check=: am__skip_mode_fix=: distdir)
make[2]: *** No rule to make target `depcomp', needed by `distdir'.  Stop.
make[1]: *** [distdir] Error 1
make: *** [dist] Error 2
```

The `make dist` works in `qucs-core`. The following doesn't:
```
[felix-salfelder-distcheck]~/git/qucs/qucs-core $ make distcheck
...snip...
./src/components/verilog/.adms.implicit.xml
./src/components/verilog/.interface.xml
./src/components/verilog/constants.vams
./src/components/verilog/discipline.vams
./src/components/verilog/disciplines.vams
make: *** [distcheck] Error 1
```

felix-salfelder:
> ERROR: files left in build directory after distclean:

this is already fixed in the toplevel-distcheck branch.

felix-salfelder:
> No rule to make target `depcomp', needed by `distdir'.  Stop.

on my end, `./bootstrap` creates `qucs/depcomp`. on your box it does not?

felix-salfelder:
@guitorri i have removed the manual lex/yacc rules. however i had to rename two headers.
if this is working, we should adapt cmake.

felix-salfelder:
> All checks have passed

wohoo!

PR #369, opened 2015-09-18
Re-add .moc.cpp to SUFFIXES to correct dependency

in3otd:
While moving back and forth between branches and recompiling, I got some strange errors on `qucs.moc.cpp`. Digging a little bit, I saw that `make` did not recognize that `qucs.h` changed and so did not regenerate the corresponding `qucs.moc.cpp`.
Apparently this is due to the fact that the line
```
SUFFIXES = .moc.cpp
```
was erroneously removed from `qucs/Makefile.am` in commit cb288ebc8f7c35586ca9ed72685e61fe711d6af5 so that the dependency
on the `.h` file was not honored.

Not sure why the rule worked when `qucs.moc.cpp` was not present but not when `qucs.h` was modified...

### DIFF
--- a/qucs/qucs/Makefile.am
+++ b/qucs/qucs/Makefile.am
@@ -94,6 +94,7 @@ EXTRA_DIST = \
 	$(hicolorsc_DATA) $(desktop_DATA)
 
 
+SUFFIXES = .moc.cpp
 .h.moc.cpp:
 	$(MOC) -o $@ $<
 


felix-salfelder:
d'uh. guess what d7598a10e4ad0c is about. it took me half a day to figure out what's actually wrong.
apparently, removing the suffixes causes even more trouble.

':D'

in3otd:
oops, I guess I should try to take more time to check the open PRs after being away for some time...

Sorry for the noise :grin:

felix-salfelder:
nevermind, the topic was entirely different. anyway i am glad the 'distcheck' work reveals actual bugs ':)'.

PR #371, opened 2015-09-19
Correct the Ideal OpAmp model

### LINKED ISSUE #140: OpAmp from the Ideal library generate netlist with syntax error
There is a bug on the OpAmp from the Ideal library. See https://sourceforge.net/p/qucs/discussion/311050/thread/cccc70a5/

In the netlist line 5 it complains about the following:
```
R:R1 _net1 _net3 R="1E3/sqrt(2*pi*fg)" Temp="26.85" Tc1="0.0" Tc2="0.0" Tnom="26.85"
```

```
line 5: syntax error, unrecognized character: `/'
line 5: syntax error, unexpected InvalidCharacter, expecting '"'
```

It declares a resistor which value is defined with an expression. This does not seem to be allowed by the parser. Other devices on the library use a Equ to precompute the values of components.

in3otd:
This fixes #140 . The OpAmp model in the Ideal library had several errors in its definition (strange but true...)

A Qucs Package with some test circuits can be found [here] (https://gist.github.com/in3otd/ee3b324524a1be277292/raw/986f5a91b86b25437c7e3e5fe79f4c254f944b28/IdealOpAmpTest.qucs)

Some sample simulations results:
idealopamp_ac

idealopamp_dc

### DIFF
--- a/qucs/qucs-lib/library/Ideal.lib
+++ b/qucs/qucs-lib/library/Ideal.lib
@@ -168,16 +168,16 @@ Simple operational amplifier model
   </Description>
   <Model>
 .Def:Ideal_OpAmp _net0 _net2 _net6 GBP="1E6" AOLDC="106" RO="75" VLIMP="14" VLIMN="-14"
-VCVS:SRC1 _net0 _net1 gnd _net2 G="G" T="0"
-R:R1 _net1 _net3 R="1E3/sqrt(2*pi*fg)" Temp="26.85" Tc1="0.0" Tc2="0.0" Tnom="26.85"
-C:C1 gnd _net3 C="1E-3/sqrt(2*pi*fg)" V=""
+VCVS:SRC1 _net0 _net1 gnd _net2 G="OLG" T="0"
+R:R1 _net1 _net3 R="R1" Temp="26.85" Tc1="0.0" Tc2="0.0" Tnom="26.85"
+C:C1 gnd _net3 C="C1" V=""
 CCVS:SRC2 _net4 _net5 gnd gnd G="1 Ohm" T="0"
-Eqn:Eqn1 OLG="10^(AOLDC/20)" fg="GBP/OLG" Export="yes"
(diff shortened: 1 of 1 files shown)


in3otd:
...night brings counsel... The equations in `qucsator` **are** evaluated only at the beginning and at the end of a simulations, not *during* a simulation. 

This is my understanding of what the code is doing:
- in the main function [`runAnalysys()`] (https://github.com/Qucs/qucs/blob/master/qucs-core/src/ucs.cpp#L251) is called
- which in turn, calls [`runSolver()`] (https://github.com/Qucs/qucs/blob/master/qucs-core/src/net.cpp#L263) which is the solver for the equations only, located [here] (https://github.com/Qucs/qucs/blob/master/qucs-core/src/environment.cpp#L214). Note that `runSolver()` uses `equationSolver()` (see below)
- then, after `runSolver()` the actual [analysis runs] (https://github.com/Qucs/qucs/blob/master/qucs-core/src/net.cpp#L264)  (`solve()` is the main method of every circuit analysis)
- when the analysis is finished the equations are [recalculated] (https://github.com/Qucs/qucs/blob/master/qucs-core/src/ucs.cpp#L255)  taking into account also the output dataset values using [`equationSolver()`] (https://github.com/Qucs/qucs/blob/master/qucs-core/src/environment.cpp#L192) again

Your example equation above is not evaluated during the simulation. Simply it's evaluated at the beginning (and there `time` is undefined so any equation using it could not be used to assign a value to a component parameter) and at the end (and there `time` is a vector and the equation system nicely handles scalar/vector/matrices computations and simply returns the vector result).

in3otd:
Travis on Mac timed out... can someone restart, please?

guitorri:
Restarted. Have you tried restarting? Owners of the repo should have access to Travis to restart the builds... unless I did something wrong on the Travis setup...

in3otd:
I didn't know I should be allowed to restart Travis so I did not even try... sorry.

ra3xdh:
Ideal opamp model now seems to be fixed. Thank you!

PR #373, opened 2015-09-20
Fixed missing wire in Transcalc

andresmmera:
It was missing a wire when generating a microstrip line in Transcalc.
This happened because the line 1249 at 'qucstranscalc.cpp' was
commented.

Steps to reproduce:
1) Open Transcalc (Ctrl+4)
2) Select 'Microstrip line' at the 'Transmission line type' combobox
3) Execute->Copy to clipboard
4) Paste in the Qucs schematic.
5) The vertical wire on the right is missing

I simply uncommented the aforementioned line...

Screenshots:
screenshot from 2015-09-20 08 54 45
screenshot from 2015-09-20 08 54 59


felix-salfelder:
interesting. `ctrl+4` just complains "cannot start line calculation program!". (sure, qucstrans is installed).

let me suggest 2 changes (sorry for the hijack):
- remove exclamation mark (from all such messages). they are annoying.
- run `qucscalc` not `/some/funny/path/qucscalc`. if `$QUCSCALC` is set, execute `$QUCSCALC` instead.i.e. copy&paste the qucsator invocation code.

in3otd:
thanks. BTW I think it's always good to check when a bug like this was introduced. I mean, there might have been a reason for that, which maybe is no longer valid. Using `git blame` (from GitHub, actually) one quickly finds that the  line was commented out [here] (https://github.com/Qucs/qucs/commit/af41d3a853c60cab275b52ab20011a9fb0c765fe#diff-3b346c1f2f367c98dac05bbed4dc08f0R1245) . In this case I assume it was jut a mistake/leftover while changing a lot of stuff during the Qt4 port...

@felix-salfelder currently we do not support a `$QUCSCALC` variable (not sure we should); even if the actual code is different, we launch `qucsator` and `qucstrans` (or any other Qucs tool) in the same way, using the absolute path `QucsSettings.BinDir` . Which is the actual invocation for `qucstrans` used by your Qucs install (should be shown in the error message dialog)? Any idea why it's not found where it should?

An absolute path is used so that you can have multiple versions of Qucs installed and every Qucs GUI will launch the corresponding tools (and simulator); if we use a relative path always the programs in the user `$PATH` will be launched (was so in the past, IIRC).

felix-salfelder:
@in3otd there is something deeply wrong. see #374

ra3xdh:
@andresmmera , I think this PR could be merged. 

@felix-salfelder ,
> interesting. ctrl+4 just complains "cannot start line calculation program!". (sure, qucstrans is installed)

`$QUCSDIR` setting helps in such cases. You can use Qucs installation in some test directory, for example: `/tmp/path_to_qucs/` and set `$QUCSDIR` before Qucs execution: 
~~~
QUCSDIR=/tmp/path_to_qucs/ /tmp/path_to_qucs/bin/qucs
~~~

> $QUCSCALC 

I am strongly against of such environment variables multiplication. We should not have specific environment variable for every tool.

guitorri:
@felix-salfelder please open another issue related to 'remove exclamation mark' if you like. Certainly there are many other warnings/messages that could be more helpful, telling at least where it was looking for the executable for instance.

I have no idea how that commented out line sneaked in... I will merge this PR.

Let us continue in #374 with the discussion on run-path and external tools.

felix-salfelder:
@guitorri good idea! sorry for being off topic, it's now #378.

PR #375, opened 2015-09-20
Correct pin assignment in LM317 regulator

### LINKED ISSUE #367: LM317 models not working as expected
Vin sweep
Vout should be 2.5V, and it should settle above around Vin=5V

Parameter sweep on R2, with constant Vin also is not at all as expected.

in3otd:
Fixes #367. The pins OUT and ADJ were swapped in the LM317 model. Note that the model has/causes some convergence issues. 
The two LM317 model in the library are very similar but not identical (a few BJT have different area and/or saturation current), so some results, like the exact output voltage or the maximum current are different.

Simulations results below, with the two LM317 models (and with the LM337 just to show it works well, with a negative voltage at input)

test317_sch

test317_dpl


ra3xdh:
Merged. Thanks @in3otd !

PR #376, opened 2015-09-20
builddir run time environment

### LINKED ISSUE #374: wrong run time paths
when running qucs out of `$(builddir), it cannot execute the other tools, this is because of some QucsBinDir hack, that does not seem to make sense in all cases.

looking from further up it seems that the run time path philosophy currently is
- qucs needs to be installed to work properly
- external tools (such as admsXml, octave) don't need to be installed, but some environment variables point to them (their build location maybe?)

instead i propose to pursue the following:
- qucs does not have to be installed in order to work properly.
- external (3rd party) tools must be installed, so qucs works right with default settings (other settings may still be possible, yet unneeded).

a package is "installed" when it can be invoked (and works) from the usual (unmodified) user environment). this typically is the case after "make install" has been run, or a package manager has installed a package (apt-get, yum and the like).

felix-salfelder:
fixes #374


in3otd:
we currently have 7 external Qucs tools (soon 8, maybe 9, I think), will we really need to define and document and test 7 separate environment variables? Plus one variable for the simulator path, one for the Library files, one for the Language files, one for the Help files, one for the Example schematic files. Nothing is at the same place in the builddir w.r.t. the install dir.
All of the external Qucs tools need to know where the Language files are, some need also to know where the Help or Library files are.
Won't it be simpler to define a command line option, e.g. `--no-install-mode` for the Qucs GUI and its Tools to tell them to use the build tree relative paths instead of the install tree relative paths. Isn't it better to get rid of all the Qucs tools environment variables ? (ok, `$QUCSATOR` is a special case, since it seems there will be an alternative/compatible simulator)

felix-salfelder:
qucs (the wrapper) should only define `PATH` in the end. nothing qucs specific to document or test. PATH is implemented somewhere deep down in `execvp` (or something like this), which we can trust here, for practical reasons.

> Won't it be simpler to define a command line option, e.g. --no-install-mode for the Qucs GUI and its Tools to tell them to use the build tree relative paths instead of the install tree relative paths.

no, because that makes the impression that we implement a special case, which it is not.

> Isn't it better to get rid of all the Qucs tools environment variables 

i'm all in. remove QUCSDIR (which breaks PATH) and then don't need the others.

felix-salfelder:
@in3otd: this is now very close to your suggestion

- does not introduce more environment variables
- praparing to get rid of QUCSDIR
- still does not affect installations
- does not conflict with documentation, just implements expected behaviour in a "corner case".

felix-salfelder:
i guess this ready for develop. i can do a complete rebase if you prefer.

PR #377, opened 2015-09-20
Fix for #372, hardcoded .qucs paths in Qucs Help and Qucs Transcalc

### LINKED ISSUE #372: Hardcoded .qucs paths
The "old style" resource files and path are still used by Qucs Help [here](https://github.com/Qucs/qucs/blob/master/qucs/qucs-help/main.cpp#L51), so that it always saves an `helprc` file in a `.qucs` directory, even if the QucsHome has been defined to be elsewhere, and reads back the settings from the same place. As a side effect, Qucs Help does not use the language selected in the main Qucs GUI.

Similarly Qucs Transcalc still tries to load a startup file from a fixed location [here](https://github.com/Qucs/qucs/blob/master/qucs/qucs-transcalc/main.cpp#L135) and sets the initial working directory to a [fixed path](https://github.com/Qucs/qucs/blob/master/qucs/qucs-transcalc/main.cpp#L116) independently of QucsHome.

in3otd:
This fixes #372. 
Now Qucs Help uses the standard QucsSetting to store/retrieve its settings and the main GUI settings (language and font).
Qucs Transcalc now stores the last used GUI input values in QucsHome instead of `~/.qucs`

### DIFF
--- a/qucs/qucs-help/main.cpp
+++ b/qucs/qucs-help/main.cpp
@@ -31,6 +31,7 @@
 #include <QDir>
 #include <QFont>
 #include <QTextCodec>
+#include <QSettings>
 #include <QDebug>
 
 #include "qucshelp.h"
@@ -43,77 +44,38 @@ QDir QucsHelpDir; // directory to find helps files
 tQucsSettings QucsSettings; // application settings
 
 // #########################################################################
-// Loads the settings file and stores the settings.
+// Loads the application settings and stores them into QucsSettings.
+// TODO: should not return a bool. it has no meaning
+// FIXME: error handling is broken, due to Qt and/or -fno-exceptions
+//        (errors loading settings are not checked/reported)
 bool loadSettings()
 {
-  bool result = true;
-
-  QFile file(QDir::homePath()+QDir::convertSeparators ("/.qucs/helprc"));
-  if(!file.open(QIODevice::ReadOnly))
-    result = false; // settings file doesn't exist
-  else {
-    QTextStream stream(&file);
-    QString Line, Setting;
-    while(!stream.atEnd()) {
-      Line = stream.readLine();
-      Setting = Line.section('=',0,0);
-      Line = Line.section('=',1,1);
-      if(Setting == "HelpWindow") {
-	QucsSettings.x  = Line.section(",",0,0).toInt();
-	QucsSettings.y  = Line.section(",",1,1).toInt();
-	QucsSettings.dx = Line.section(",",2,2).toInt();
-	QucsSettings.dy = Line.section(",",3,3).toInt();
-	break; }
-    }
-    file.close();
-  }
-
-  file.setFileName(QDir::homePath()+QDir::convertSeparators ("/.qucs/qucsrc"));
-  if(!file.open(QIODevice::ReadOnly))
-    result = true; // qucs settings not necessary
-  else {
-    QTextStream stream(&file);
-    QString Line, Setting;
-    while(!stream.atEnd()) {
-      Line = stream.readLine();
-      Setting = Line.section('=',0,0);
-      Line = Line.section('=',1,1).trimmed();
-      if(Setting == "Font")
-	QucsSettings.font.fromString(Line);
-      else if(Setting == "Language")
-	QucsSettings.Language = Line;
-    }
-    file.close();
-  }
-  return result;
+  QSettings settings("qucs","qucs");
+  settings.beginGroup("QucsHelp");
+  if(settings.contains("x"))QucsSettings.x=settings.value("x").toInt();
+  if(settings.contains("y"))QucsSettings.y=settings.value("y").toInt();
+  if(settings.contains("dx"))QucsSettings.dx=settings.value("dx").toInt();
+  if(settings.contains("dy"))QucsSettings.dy=settings.value("dy").toInt();
+  settings.endGroup();
+  if(settings.contains("font"))QucsSettings.font.fromString(settings.value("font").toString());
+  if(settings.contains("Language"))QucsSettings.Language=settings.value("Language").toString();
+  return true;
 }
 
 // #########################################################################
-// Saves the settings in the settings file.
+// Saves the application settings.
+// TODO: should not return a bool. it has no meaning
+// FIXME: error handling is broken, due to Qt and/or -fno-exceptions
+//        (errors saving settings are not checked/reported)
 bool saveApplSettings(QucsHelp *qucs)
 {
-  if(qucs->x() == QucsSettings.x)
-    if(qucs->y() == QucsSettings.y)
-      if(qucs->width() == QucsSettings.dx)
-	if(qucs->height() == QucsSettings.dy)
-	  return true;   // nothing has changed
-
-
-  QFile file(QDir::homePath()+QDir::convertSeparators ("/.qucs/helprc"));
-  if(!file.open(QIODevice::WriteOnly)) {
-    QMessageBox::warning(0, QObject::tr("Warning"),
-			QObject::tr("Cannot save settings !"));
-    return false;
-  }
-
-  QString Line;
-  QTextStream stream(&file);
-
-  stream << "Settings file, Qucs Help System " PACKAGE_VERSION "\n"
-	 << "HelpWindow=" << qucs->x() << ',' << qucs->y() << ','
-	 << qucs->width() << ',' << qucs->height() << '\n';
-
-  file.close();
+  QSettings settings ("qucs","qucs");
+  settings.beginGroup("QucsHelp");
+  settings.setValue("x", qucs->x());
+  settings.setValue("y", qucs->y());
+  settings.setValue("dx", qucs->width());
+  settings.setValue("dy", qucs->height());
+  settings.endGroup();
   return true;
 }
 
--- a/qucs/qucs-transcalc/main.cpp
+++ b/qucs/qucs-transcalc/main.cpp
@@ -58,6 +58,11 @@ bool loadSettings()
     settings.endGroup();
     if(settings.contains("font"))QucsSettings.font.fromString(settings.value("font").toString());
     if(settings.contains("Language"))QucsSettings.Language=settings.value("Language").toString();
+    if(settings.contains("QucsHomeDir"))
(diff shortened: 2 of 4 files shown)


in3otd:
rebased

Should actually go to `develop` or even `release-0.0.19` nowadays, but the PR was originally opened to `master`...

qucs/qucs-help/main.cpp
```diff
@@ -46,74 +47,29 @@ tQucsSettings QucsSettings; // application settings
...
-
-  file.close();
+  QSettings settings ("qucs","qucs");
+  settings.beginGroup("QucsHelp");
+  settings.setValue("x", qucs->x());
+  settings.setValue("y", qucs->y());
+  settings.setValue("dx", qucs->width());
+  settings.setValue("dy", qucs->height());
+  settings.endGroup();
   return true;
```
felix-salfelder: here, `true` means "nothing has changed". in your implementation, it seems to have another meaning... (which?!)

qucs/qucs-help/main.cpp
```diff
@@ -46,74 +47,29 @@ tQucsSettings QucsSettings; // application settings
...
+  if(settings.contains("font"))QucsSettings.font.fromString(settings.value("font").toString());
+  if(settings.contains("Language"))QucsSettings.Language=settings.value("Language").toString();
+  return true;
 }
 
 // #########################################################################
 // Saves the settings in the settings file.
 bool saveApplSettings(QucsHelp *qucs)
 {
-  if(qucs->x() == QucsSettings.x)
```
felix-salfelder: without this check, the behaviour is changing. what do you do to compensate for this? (cannot find anything within your commit).

qucs/qucs-help/main.cpp
in3otd: looking also at the similar `saveSettings()` functions around I understood that the function returned true when the settings were successfully saved or if the already saved settings are the same as the current application settings, which from the user's point of view is the same, I mean, if the function returned true, the user will have found the same settings when the application was restarted.
This was maybe useful in the past, when the settings were "manually" saved to a local file and not using `QSettings`. In any case I have the impression that the return value of the various `saveSettings()` function is actually never checked.

qucs/qucs-help/main.cpp
in3otd: yes, it's slightly changed, as now the settings are saved even if the currently saved ones are identical. But the end result should be practically the same, or?

felix-salfelder: > as now the settings are saved even

this seems to be a regression to me. writing out a file unnecessarily is bad practice. this should not be changed, particularly because it was already there.

what's even worse: does `QSettings settings ("qucs","qucs");` *parse* the configuration?

> But the end result should be practically the same

i cant see, how the settings are written out *at all* (is this hidden in the destructor of qucssettings?!). maybe there should be a global handle on settings, with a "flush()" member, instead of random file i/o.

    auto& settings=global_qucssetitings;
    settings.set("x", 17);
    settings.flush(); // write out lazily, and error handling here.

makes sense, no? i don't think that touching the code without actually imeplementing it in a sane way is worth the trouble (review/testing). it won't converge.

in3otd: all the actual setting reading/writing is done by the Qt `QSettings` class. AFAIU the idea of `QSettings` is really that the programmer does not need to care where the settings go (in Linux they usually go to a hidden file in the user's home, in Windows they go to the registry, in OS X, I do not know). `QSettings` may cache the settings, may avoid unnecessary writing to file or memory by comparing the new values with the old ones, may delay the actual writing until the destructor is called or the entire application is closed, I do not know and I do not see why I need to know.
Why are you implying that the changes here are not sane? As said, before all was done "manually", settings explicitly written to a file by the application and `false` was returned if the settings could not be written. Then the return value was anyway ignored. Seems as sane as it is now.

felix-salfelder:
> Why are you implying that the changes here are not sane?

because
- every time settings are written, a file must be parsed (a regression).
- the approach pushes data into a temporary object, which should not be temporary at all.
- something implicit/magic appears to happen (i am guessing) when deleting that object.

i claim that
- the current implementation is already correct, so why change it (i mean, the behaviour, not the elegance)?
- previously there was a global object that took care. this is certainly more correct.
- flushing somehing to a file (or to whatever else) should be more explicit. the same way it should be made more obvious that a file is being *parsed* for no apparent reason.

more general, i would say that a PR should *either* change the behavior of the code, *or* cleanup/straighten/deduplicate the implementation. here a cleanup (adding some elegance) is entangled with a behavioural change (to the worse). needless to say that this is hard to review -- i'm merely guessing, not understanding the code, and the parts that i seem to understand are obviously not improvements.

> As said, before all was done "manually"

i fully agree with you that the current code is ugly. if it needs a fix, then *only* fix it. for example, the meaning of the return value is unclear. if it needs a rewrite, go for a *complete* rewrite, reflecting what the author had in mind (which is not all that bad). the parts you delete will just be gone indefinitely.

hth

in3otd:
I think you were expecting too much from this PR: as said, its main purpose it to fix #372.
To do that it aligns the method used to read/store the application settings to the one already used by the other applications.

Reading the settings file takes about 150 &mu;s on my machine; I think we can afford to do it exactly *twice* in an application - I guess parsing them takes much less. Besides, on other OSs the settings may not even be in a file and, as you noticed, we do not know much about the internals of `QSettings`, if/when/how it read/parses stuff. I maintain that we do not need to know, that's the reason of using a more abstract class instead of doing raw file I/O. Optimization for accessing the settings values should go in `QSettings` not in the application using it.

BTW, thinking about it, [`QSettings`] (http://doc.qt.io/qt-4.8/qsettings.html) should support even having multiple processes accessing the same settings, so it should correctly manage settings even if two instances of the same application are running, a thing that the old method did not handle correctly (but as said this was not the motivation for this PR).

felix-salfelder:
> I think you were expecting too much from this PR

i explained the difficulties i am having with the patches. i do not have a problem with somebody else merging this, ignoring my objections. i am already aware that my objections may be formal, pedant or nonsense to others. but i *did* read the code and i do not agree.

> Reading the settings file takes about 150 μs on my machine [..] 

to me, it does not matter how long it takes. it's a structural regression. to be more precise: it *looks* like one to me. (global config object gone and stuff, see above)

> we do not know [..] if/when/how it read/parses stuff.

it doesn't work like that. i will not merge unless/before i know.

try this: pick a random PR, at best one that passes CI, possibly one you don't really need. read it. then either hit "merge" or explain why you wont. work out alternatives that you would be willing to merge, without writing another unreviewed patchset yourself.

in3otd:
> i am already aware that my objections may be formal, pedant or nonsense to others. but i did read the code and i do not agree.

yep, I have some difficulties in following your remarks sometimes, but thanks for reading the code

> it's a structural regression. to be more precise: it looks like one to me. (global config object gone and stuff, see above)

I would not have deleted the "settings not changed" part, had I done only minor modifications to the code, I agree it made sense (but see the **Note** below) even if it had no practical impact for the user, but since I had rewritten that part to use `QSettings` I am not convinced it still makes sense

> > we do not know [..] if/when/how it read/parses stuff.

> it doesn't work like that. i will not merge unless/before i know.

but we do not control what `QSettings` does and its docs does not guarantee any particular behavior

> try this: pick a random PR, at best one that passes CI, possibly one you don't really need. read it. then either hit "merge" or explain why you wont. work out alternatives that you would be willing to merge, without writing another unreviewed patchset yourself.

I understand the problem here; either merge a suboptimal (but working) code, wait for someone else to help out if the OP does not change the PR or add your own unreviewed fixes. I have done all three but not happy with any of these "solutions".

**Note**: as said before, I've realized that the "[check for settings not changed] (https://github.com/Qucs/qucs/blob/master/qucs/qucs-help/main.cpp#L95-L99)" is wrong. Try this:
- launch Qucs
- press F1 to launch the Help
- do not move/resize this Help window (so its settings do not change)
- (move the focus on Qucs again and) press F1 again to open another help window
- move around/resize this window, then close it
- close also the 1st Help window
- press F1 to reopen the Help: it will not have the position/size of the last Help window closed, but of the previous one...

(yes, pressing F1 multiple times should probably just maximize a previously open Help window but one should still be able to explicitly launch/use a different one. Will be another PR, one day)

felix-salfelder:
ok. i have browsed the QT manual and code a bit, and it seems that your usage of `QSettings` is proper or at least as intended. as it seems, Qsettings in fact does hide a sort of cache behind the scenes, or maybe should. well it possibly *could* and even *might* do the right thing in some implicit way... (whatever, sorry for the delay).

i'd like to add a note about the nonsensical return code, such as

    // TODO: should not return a bool. it has no meaning
    // FIXME: error handling is broken, due to QT and/or -fno-exceptions

then merge into the 0.0.19 release. do you agree?

in3otd:
> i'd like to add a note about the nonsensical return code, such as
> `// TODO: should not return a bool. it has no meaning`

ok, this I understand

> `// FIXME: error handling is broken, due to QT and/or -fno-exceptions`

here I do not get what you mean...

felix-salfelder:
`-fno-exceptions` disables exceptions. exceptions are the standard way to communicate trouble, such as "cannot open `qucs.rc`, file not found", "cannot write to `qucs.rc`. permission denied" or "cannot create `qucs.rc`, no space left on device". or whatever else, the extra layer might encounter when talking to the config db.

now this sort of exception is exactly what `void saveApplSettings(QucsHelp *qucs)` should throw. but it's disabled right now (and there's a stale unused bool which makes me shudder)...

clearly, error handling should be available (and implemented!) throughout the project, not just here. i think adding a comment is better than ignoring the issue.

in3otd:
updated with comments and force-pushed

felix-salfelder:
merged into `develop-0.0.19`. github did not notice ... probably.

can you confirm and close this PR?

in3otd:
yes, merged into `release-0.0.19` (e5a28edea5960816416667b22ddcf0805d11ad45), works as expected (but I think I saw an unrelated issue while testing...), thanks.

Seems GitHub does not close automatically if not merged to the original target branch. There may be a way to have it closed automagically with some keywords in the merge commit comment but I don't think it's worth it.

PR #378, opened 2015-09-21
remove exclamation marks from error messages!

felix-salfelder:
scanning through https://en.wikipedia.org/wiki/Exclamation_mark, i cannot find a compelling reason why to remove exclamation marks. but there's some evidence. make up your own opinion...

this commit was a quick `sed -i`, but it looks correct at first glance. note that it removes the punctuation independent of the language.

### DIFF
--- a/qucs/translations/qucs_ar.ts
+++ b/qucs/translations/qucs_ar.ts
@@ -5768,7 +5768,7 @@ Wrong line start!</source>
     <message>
         <location line="+20"/>
         <source>Format Error:
-Unknown component!
+Unknown component
 %1
 
 Do you make use of loadable components?</source>
@@ -12795,7 +12795,7 @@ Wrong &apos;component&apos; line format!</source>
     </message>
     <message>
         <location line="+13"/>
-        <source>ERROR: Cannot create user library subdirectory !
+        <source>ERROR: Cannot create user library subdirectory
 </source>
         <translation> خطأ : لا يمكن إنشاء دليل فرعي لمستخدم المكتبة !</translation>
     </message>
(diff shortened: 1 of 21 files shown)


guitorri:
I am not really bothered by the exclamation marks... It does seems that people don't like to be yelled  at by a computer, [from this](http://programmers.stackexchange.com/questions/13556/exclamation-points-in-error-messages).

You have to update the source code first. The translatable strings originate in the `tr()` macro.
The `source` strings are populated into the `.ts` by `lupdate` (see `qucs/qucs/translations`). 
When source strings change, the translations are marked as modified or fuzzy (I forgot the right term).
Typically manual intervention is necessary to clear/validate the translation. I believe translations are not shown if marked as fuzzy.

So, you might need to change the sources, run lupdate and fix the translations accordingly. 
Something like dd369787bea8975fcba3b21782bb3e6abc66ff38 and d1d9f797d041f53c30d5f2352f572cd0704b0307. 

Then we have to synchronize with Transifex to avoid overwrites and losing translations. :smile:

in3otd:
yep, there is also https://techbase.kde.org/Projects/Usability/HIG/Exclamation_points . But can anyone please explain me the rationale behind the sentence there "*if the game could not be saved to disk, the error message would not take an exclamation mark.*" ??

Anyway, I'm not for or against exclamation marks in particular, but I'm for having (more or less) uniform error messages. If currently all the error messages have exclamation marks leave them as they are, if there are error messages with and without them, choose one format and apply that to all messages.

Not sure if we should remove the exclamation marks from the translated sentences, Wikipedia says the exclamation mark usage "varies slightly between languages", maybe other languages use them more (or less?) often.

felix-salfelder:
> So, you might need to change the sources, run lupdate and fix the translations accordingly.
Something like dd36978 and d1d9f79.

oops. it seems i edited autogenerated files. but at least I did not add them `:D`

(i will have another look. maybe, some day. as you may have guessed, this is not a priority case!!!1)

guitorri:
They are not autogenerated. The `.ts` files work as a translation database. The binary version of them, the `.qm` files, are autogenerated. It is about the same with .po files managed with GNU gettext. 
When translatable strings are changed in the code, they need to be updated in the database (as source first and then the respective translations).
Ideally there should be a period of string-freeze. No more changes in `tr()` fields are allowed. To give time to translators to update the latest strings.

felix-salfelder:
ok, sounds like a can of worms i'd better open after the .19 release

sorry for the noise

guitorri:
Changes must be made on the `.cpp` source strings and then propagated to `.ts` files.
Look for the `tr()` functions.
Closing this one.

PR #381, opened 2015-09-25
Update Perl modules installation instructions

resetnow:
For Debian/Ubuntu, a preferred way to install these Perl modules is to use ones from the package repository. (in my case CPAN installed modules are not detected during configure on Ubuntu 15.04).

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,11 +69,10 @@ For ADMS
     sudo apt-get install zlib1g zlib1g-dev
     sudo apt-get install libgd2-xpm-dev
 
-Install Perl modules for ADMS. Configure cpan, no waiting for user input.
+Install Perl modules for ADMS:
 
-    (echo y;echo o conf prerequisites_policy follow;echo o conf commit)| sudo cpan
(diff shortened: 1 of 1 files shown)


resetnow:
1. I didn't change neither add this line
2. As far as I understand, those are runtime dependencies
3. There was `sudo cpan` before, which is questionable because it is possible to make cpan work without `sudo`. `apt-get`, however, needs `sudo` most of the time. Also, there is a `sudo apt-get` a couple of lines above, so I have added it here too for consistency:  https://github.com/resetnow/qucs/commit/7ff2a95fad02ff8e00622c198e651b07867aa847#diff-04c6e90faac2675aa89e2176d2eec7d8L70

felix-salfelder:
your PR made me have a closer look at README.md, hence the remarks. as i wrote, the PR is consistent with what we have. i will address the other issues in another PR. merging now.

slightly off topic: your assertion about apt-get and sudo is false. when in doubt, consider the manual.

PR #387, opened 2015-09-29
Toplevel distcheck

felix-salfelder:
make distcheck work at toplevel. depends on #361.

also, include the travis batch file into the check target

- extend the scope of travis checks to "distcheck"
- simplify travis rules (do a distcheck, nothing else)
- improve reproducibility (make (dist)check at home)
- include tests into dist tarball (where they belong)

Issue #205, opened 2015-02-21
Reminder about qucs-test, platform dependent issues and numerical differences.


guitorri:
This issue is a reminder about the numerical issues mentioned below. 

This issue should cleared/understood before the next release.

Tests failing the numerical check are being skipped (on [Linux](https://github.com/Qucs/qucs-test/blob/master/skip.txt) and [OSX](https://github.com/Qucs/qucs-test/blob/master/skip_OSX.txt)) to allow Travis CI to finish with success.

Running the [qucs-test](https://github.com/Qucs/qucs-test) on different platforms may lead to numerical differences (with respect to results generated in qucs 0.0.17, linux 64).

See these commits with figures for Darwin, Windows and Linux (see branch name):
https://github.com/Qucs/qucs-test/commit/aee5f5a5809a6abaecc07a7fa1725b25e768ce32
https://github.com/Qucs/qucs-test/commit/e86c82ec4da1215593aa40d4f2a155d99577558e
https://github.com/Qucs/qucs-test/commit/8941f2b8e93b78bdb21d2529a21133594f89277a

The above tests were executed with rtol=0.1 and atol=1e-5.
The numpy.allclose returns true if the absolute error of two arrays is lower-equal than the sum of two tolerances. Like: abs(a-b) <= (atol + rtol*b)

Looking at the pictures more carefully the results are not that far apart.
Indeed fast switching and accumulated error seem to be related to these issues.

On TR_Puls3b_prj the tswitch seems to be toggled one step before/after of what used to be.
On Windows (32bit !!) some circuits have a phase reversal at start and the accumulated error is much more pronounced.

yodalee:
@guitorri maybe you can write this into wiki?

guitorri:
Write what in the wiki?? 
Platform dependency is an issue. So far we could not trace the root of the problem. We need more examples and cross-check with other simulators to see if we can pinpoint the source of the differences.

yodalee:
I think I misunderstand the issue, sorry about that.

guitorri:
Reminder to check what happened in #623

guitorri:
Some reference files still show differences after #206.

PR #389, opened 2015-09-30
integrate qucs tests into qucs package

felix-salfelder:
currently some tests are not included with qucs. including these tests will increase test coverage.

with this, `make check` will run all tests we have, not just few of them.


felix-salfelder:
almost...

    Warning: QucsConv not found:  "qucsconv" 
    Error: Could not prepare the netlist...
    Critical: "ERROR: Cannot load library component "D1"." 

can you merge #376? i'd like to not mix it up.

felix-salfelder:
@guitorri some tests don't pass on osx (I stumbled upon the extra skiplist). this is unexpected. can you point me to the reason for this?

felix-salfelder:
> All checks have passed

hey, with this, we could finally produce release candidate tarballs...

@guitorri i have tentatively removed the coverage thing. may I suggest to compute coverage only for the `develop` branch? this will cut the check run time for PRs. and also introduce a place where we could add more checking/snapshotting/profiling of any duration we might want to have some day.

guitorri:
This is looking good. :+1: 

With respect o the skipped tests. See #205 for some more details. There are platform dependency issues in the transient code or components. 

About coverage. We could have branches specialized for each task. Coverage, static analysis, packaging, doxygen, nightly builds... sure. The .travis.yml grew as I managed to hook more stuff in it. Perhaps it is time to clean it up.

I agree, perhaps is not necessary to check coverage on each PR.

The wget patch 8765e8d  is really about Googletest integration (I still don't get what you mean by software distribution...). The reason why I put it into the configure (as an option) was to facilitate the download and compilation. It is supposed to be compiled along with the same flags as the rest of the code. If enable it runs with `make check`, adding more unit tests. I know GoogleTest is currently underutilized, but it might come in handy when crafting unit tests to cover corner cases that are hard to reach. Alternatively we could the code it into the repository, but for some reason I forgot I did not include at that time.

I will try to find some time tomorrow to kick the tires of this PR.

felix-salfelder:
> about coverage. We could have branches specialized for each task. Coverage, static analysis, packaging, doxygen, nightly builds... sure. The .travis.yml grew as I managed to hook more stuff in it. Perhaps it is time to clean it up.

i will/might start/experiment with a develop branch. i am not yet fully familiar with the travis thing. lets add/reassign tasks to other branches as it seems appropriate.

> The wget patch 8765e8d is really about Googletest integration (I still don't get what you mean by software distribution...).

in my understanding. a *distribution* is something that puts together and delivers *packages*. qucs is one such package. imo it should not try to process other packages.

NB: i am aware of the ADMS "sub"package, i consider this a leftover from times without an ADMS release tarball. or should we also ship a kernel a shell and a c++-compiler...? (imo: no. others are truly better at this).

> facilitate the download and compilation.

wget in configure is clearly beyond the scope of the configure program. configure is originally meant to create makefiles (+ related things) --  and this is what the user expects it to do (least-surprise-stuff...). note that the wget command leaks user data, and can fail if there's a firewall. let me suggest to turn the wget block into something like the following.

    if $something
      AC_MSG_ERROR([not found. consider running get-$packagename.sh to fetch from $somewhere])
    fi

to reduce the surprise factor. i don't insist on removing them `:)`.

felix-salfelder:
this seems to be working as intended now..

guitorri:
Looking good! Is there anything else you want to do here? 

Yesterday I was about to merge this. 
I got some issue with `make distcheck` (dist and check worked), related to `qucs-core\tests\basic`. I will try to reproduce.

felix-salfelder:
> Looking good! Is there anything else you want to do here?

no. that's it. note that this is actually 3 PRs. when in doubt, merge the others first.

> I got some issue with make distcheck (dist and check worked), related to qucs-core\tests\basic. I will try to reproduce.

thanks. yes -- lets fix issues first, let me know the details.

yesterday, after I merged master, something strange happened (see bd822ea for the workaround). if you have an idea, lets properly fix this as well.

guitorri:
I saw that this includes #361 and #387. I am OK with merging this PR. I also want to see if GitHub will close the intermediate PRs.

The distcheck issue I mentioned did not surfaced again. 

I have two small patches to send later. One includes to distdir some pdf figures needed by qucs-doc. If I try to extract and build the tarball made run `make dist` in another location latex complains about missing figures.

The `qucs-test/.git/` directory is also being copied into distdir. Yes, we need to improve this later.

I will merge this and we work from there. 

Thank you so much!

PR #392, opened 2015-10-06
structured simulator output (WIP)

felix-salfelder:
this is meant to implement the data structures for generic data sources. these are required to read data from different files (hdf5?) or different simulators.

still waiting for a Simulator base class (?).

independently, the current implementation is using boost for iterator type erasure...


felix-salfelder:
not going to happen in 0.0.20 branch

Issue #386, opened 2015-09-29
3D cartesian plot is broken


ra3xdh:
I have just found another critical bug in Qucs-GUI. 3D-cartesian plot is broken in `master`. Open two schamtics with 3D plots in 0.0.18 and 0.0.19 and compare results to reproduce this bug. 3D plot in 0.0.19 looks unreadable.  Here is example screenshot. Left is Qucs-0.0.18, right is Qucs-0.0.19. 
![3d-diagr](https://cloud.githubusercontent.com/assets/4920080/10156918/583bfc2a-668e-11e5-8177-bed3cb3234e4.png)

felix-salfelder:
it seems, the new code does not print as many cross connections.

the missing blue sample dots in the left picture are more interesting. is this related to the z buffer? what happens if you slightly turn the plot to the left (= move the camera to the right), will they reappear?

please post the data file and the schematic somewhere, i will have a look.

ra3xdh:
Here is test schematic https://gist.github.com/ra3xdh/eb92c81d7d72d401668b/raw/12be21db82ff579e91a9a7e4aa63370a25df2ba3/diode_dblswp_qucs.sch  3D cartesian plot is embedded in schematic. 

> what happens if you slightly turn the plot to the left (= move the camera to the right), will they reappear?

Yes, missing points could appear when plot is rotated at specific angle around Z-axis.

felix-salfelder:
thanks

whether the cross connections are painted or not depends on the angle. must be a bug in the zbuffer code. also, i get a lot of "dangerous: returning negative screen coordinate" messages, which may or may not be related...

felix-salfelder:
apparently, hideLines is broken for 3dplots. i am working on this in #370. previously the z buffer has been monkey-patched into the 2d ploting routines. these have been replaced by the QtPainter.

i think i will implement 3dgraphs with their own drawings...

guitorri:
If at a certain angle the user still can get a useful image, then I will count it as know issue/regression. 
I will move this to the next release, I hope it is ok if you guys.

felix-salfelder:
i have failed to make the old code work again. wasted a lot of time on this `:(`

using qt for the 2d plots is working pretty well. i think we should head for something like qwtplot to do the 3d plots. unfortunately qwtplot does not support multiple layers (yet)...

ldpgh:
Simple check with x=5 SweepPts  :  y=5 SweepPts
[...]

PR #393, opened 2015-10-08
3dplot (WIP)

felix-salfelder:
starting to work on a fix for #386.


felix-salfelder:
the basepr has changed. needs more time.

PR #395, opened 2015-10-12
Wattmeter and internal resistor

Pedro-project:
In this code it was added an wattmeter, which is capable of measuring: tension, intensity, real power, reactive power, apparent power and power factor.
It was also addad an internal resistor to Vdc, Vac, Idc, Iac, Wprobe (wattmeter i created) and amperimeter.

All works fine with me and i would like to know your opinion. I am new on GitHub so if anything is wrong please let me know.

If possible, i would like a quick review of it, because your approval will would be a good help for my graduation presentation.

Thank you.


felix-salfelder:
something went wrong in 510ef3637. i suggest to `git rebase -ip master`, and edit out that commit.

PR #370, opened 2015-09-19
Graph rework

felix-salfelder:
continuing #314.... rebased

@in3otd i did not manage to disable qDebug per-file. i cleaned up some unnecessary output, does this improve the situation on your end?

PR #407, opened 2015-12-31
Enhanced markers

andresmmera:
Hello all, 

This PR modifies the existing code of the Marker class so as to allow the marker objects to work as delta markers. Additionally, it was modified the MarkerDialog class to let the user to decide de line width and the color of the markers.

It is not a big contribution, but I hope it helps ;)

Normal mode
selection_005
Delta mode
selection_006
MarkerDialog window
selection_007

P.S.: Originally, my idea was to include the markers in the dataset to perform math operations in the display window, like the ADS does,... but, at some point, I've realized that Qucs cannot handle equations after the simulation :(

### DIFF
--- a/qucs/qucs/diagrams/marker.cpp
+++ b/qucs/qucs/diagrams/marker.cpp
@@ -4,6 +4,10 @@
     begin                : Sat Apr 10 2004
     copyright            : (C) 2003 by Michael Margraf
     email                : michael.margraf@alumni.tu-berlin.de
+****************************************************************************
+    Update               : Added delta markers feature. March 2017
+    Author		 : Andres Martinez-Mera
+    email                : andresmartinezmera@gmail.com
  ***************************************************************************/
 
 /***************************************************************************
@@ -29,7 +33,7 @@
 #include <QString>
 #include <QPainter>
 #include <QDebug>
-
+#include <QFile>
 #include <limits.h>
 #include <cmath>
 #include <stdlib.h>
@@ -46,7 +50,7 @@ static double default_Z0=50;
  * marker position is the sampling point closest to the click.
  */
 
-Marker::Marker(Graph *pg_, int branchNo, int cx_, int cy_) :
+Marker::Marker(Graph *pg_, int branchNo, int cx_, int cy_, QString markerID) :
   Element(),
   pGraph(pg_),
   Precision(3),
@@ -55,11 +59,20 @@ Marker::Marker(Graph *pg_, int branchNo, int cx_, int cy_) :
 {
   Type = isMarker;
   isSelected = transparent = false;
-
   cx =  cx_;
   cy = -cy_;
   fCX = float(cx);
   fCY = float(cy);
+  
+  MarkerID=markerID;
+  MarkerColor = Qt::black;
+
+  ReferenceMarkerData = {0,0,0,0,0,0};
+  MarkerLineWidth = 1;
+  MarkerMode=0;// MarkerMode = 0 => Conventional marker
+               // MarkerMode = 1 => Delta marker
+  ReferenceMarkerID = "#NO_REF#";//No reference
+
   if(!pGraph){
     makeInvalid();
   }else{
@@ -70,11 +83,63 @@ Marker::Marker(Graph *pg_, int branchNo, int cx_, int cy_) :
 
   x1 =  cx + 60;
   y1 = -cy - 60;
-
 }
 
 Marker::~Marker()
 {
+}
+
+QMap<QString, std::vector<double>> Marker::getMarkersMap()
+{
+  return ActiveMarkers;
+}
+
+QString Marker::getID(void)
+{
+return MarkerID;
+}
+
+void Marker::setID(QString ID)
+{
+MarkerID = ID;
+createText();
+}
+
+
+
+int Marker::getLineWidth()
+{
+  return MarkerLineWidth;
+}
+
+void Marker::setLineWidth(int width)
+{
+  MarkerLineWidth = width;
+}
+
+void Marker::setReferenceMarkerID(QString ref)
+{
+   ReferenceMarkerID=ref;
+}
+
+
+// This function returns the marker values. Its main use is to provide data to the Diagram object so
+// as to build a map with the data of all the active markers.
+std::vector<double> Marker::getData()
+{
+  std::vector<double> data(6);
+  data[0] = VarDep[0];//Real part
+  data[1] = VarDep[1];//Imaginary part
+  data[2] = VarPos[0];//Independent variable
+  data[3] = fCX;//Marker cordinates
+  data[4] = fCY;
+  data[5] = MarkerMode;//This allows to save the marker mode in a map structure. This map is used by markerdialog class to separate those conventional from delta markers
+  return data;
+}
+
+void Marker::setMarkersMap(QMap<QString, std::vector<double>> markersMap)
+{
+   ActiveMarkers = markersMap;
 }
 
 // ---------------------------------------------------------------------
@@ -211,7 +276,8 @@ void Marker::createText()
     makeInvalid();
     return;
   }
-
+  // independent variables
+  MarkerMode ? Text = MarkerID + QString("\nDelta mode, Ref.: ") + ReferenceMarkerID + "\n" :  Text = MarkerID + "\n";
   unsigned nVarPos = VarPos.size();
 
   if(nVarPos > pGraph->numAxes()){
@@ -224,9 +290,13 @@ void Marker::createText()
       VarPos[nVarPos++] = 0.; // pad
     }
   }
-
-  // independent variables
-  Text = "";
+  std::vector<double> VarRef(nVarPos+2);//nVarPos independent variables + 1 complex dependent variable 
+  
+  if (MarkerMode == 1)//Delta mode activated
+  {  
+    //Get reference marker data
+    ReferenceMarkerData = ActiveMarkers[ReferenceMarkerID];
+  }
   double *pp;
   nVarPos = pGraph->numAxes();
   DataX const *pD;
@@ -257,19 +327,58 @@ void Marker::createText()
 
   // now actually create text.
   for(unsigned ii=0; (pD=pGraph->axis(ii)); ++ii) {
-    if(ii==0 && diag()->Name=="Waveac")
+    if (MarkerMode == 0)//Default marker
+    { 
+     if(ii==0 && diag()->Name=="Waveac")
       Text += "Time: " + unit(VarPos[ii]) + "\n";
-    else
+     else
       Text += pD->Var + ": " + QString::number(VarPos[ii],'g',Precision) + "\n";
+    }
+    else//Delta marker
+    {
+     Text += QChar(0x0394) + pD->Var + ": " + QString::number(VarPos[ii]-ReferenceMarkerData[2],'g',Precision) + "\n";
+    }
+  }
+  
+  if (MarkerMode == 1)//Delta marker
+  {
+     Text += QChar(0x0394) + pGraph->Var + ": ";
+  }
+  else//Default marker
+  {
+     Text += pGraph->Var + ": ";
   }
 
-  Text += pGraph->Var + ": ";
   switch(numMode) {
-    case nM_Rect: Text += misc::complexRect(*pz, *(pz+1), Precision);
+    case nM_Rect: 
+        if (MarkerMode)//Delta marker
+        {
+            Text += misc::complexRect(*pz-ReferenceMarkerData[0], *(pz+1)-ReferenceMarkerData[1], Precision);
+        }
+        else//Default marker
+        {
+            Text += misc::complexRect(*pz, *(pz+1), Precision);
+        }
       break;
-    case nM_Deg: Text += misc::complexDeg(*pz, *(pz+1), Precision);
+    case nM_Deg: 
+        if (MarkerMode)//Delta marker
+        {
+            Text += misc::complexDeg(*pz-ReferenceMarkerData[0], *(pz+1)-ReferenceMarkerData[1], Precision);
+        }
+        else//Default marker
+        {
+            Text += misc::complexDeg(*pz, *(pz+1), Precision);
+        }
       break;
-    case nM_Rad: Text += misc::complexRad(*pz, *(pz+1), Precision);
+    case nM_Rad: 
+       if (MarkerMode)
+       {
+         Text += misc::complexRad(*pz-ReferenceMarkerData[0], *(pz+1)-ReferenceMarkerData[1], Precision);
+       }
+       else
+       {
+          Text += misc::complexRad(*pz, *(pz+1), Precision);
+       }
       break;
   }
 
@@ -456,9 +565,34 @@ bool Marker::moveUpDown(bool up)
   return true;
 }
 
+void Marker::setColor(QColor c)
+{
+  MarkerColor = c;
+}
+
+QColor Marker::getColor()
+{
+  return MarkerColor;
+}
+
+void Marker::setMarkerMode(int mode)
+{
+  MarkerMode = mode;
+}
+
+int Marker::getMarkerMode() const
+{
+  return MarkerMode;
+}
+
+
 // ---------------------------------------------------------------------
 void Marker::paint(ViewPainter *p, int x0, int y0)
 {
+  createText();
+
+// is a flag which indicates whether the marker was previously saved or not.
+
   // keep track of painter state
   p->Painter->save();
 
@@ -473,9 +607,13 @@ void Marker::paint(ViewPainter *p, int x0, int y0)
   x2_ = p->drawText(Text, x0+x1+3, y0+y1+3, &y2_);
   x2_ += int(6.0*p->Scale);
   y2_ += int(6.0*p->Scale);
-  if(!transparent) {
+  if(transparent) {
     p->eraseRect(x0+x1, y0+y1, x2_, y2_);
     p->drawText(Text, x0+x1+3, y0+y1+3);
+  }
+  else
+  {
+    p->drawRectD(x0+x1, y0+y1, x2_, y2_);
   }
   p->Painter->setWorldMatrix(wm);
   p->Painter->setWorldMatrixEnabled(false);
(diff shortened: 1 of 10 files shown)


andresmmera:
Uhm, Travis-CI crashed and left this message:
'*No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself*'
Does anyone know what is the origin of the problem? Is this related to the code?
On the other hand, I've noticed that AppVeyor is queuing every task but it seems to be quite stuck.

ra3xdh:
TravisCI crashes while running netlist test. To be certain, it cannot run test on `DC_SW_bsim4v30pMOS_Ids_Vds_prj` project. There is a problem while opening schematic with 3D-plots and markers. Please download Qucs test suite https://github.com/Qucs/qucs-test and try to open `testsuite/DC_SW_bsim4v30pMOS_Ids_Vds_prj/bsim4v30pMOS_Ids_Vds.sch` with your modified version of Qucs. It gives a `Wrong 'diagram' line format!` error and cannot open it. But `master` version operates on this schematic normally. So, there is a problem with the modified markers and 3D diagrams.

andresmmera:
Ok, thank you @ra3xdh, I've just found what's going on

guitorri:
I got this one rebased locally. I see you also performed a few Qt3Support fixes. I guess I will wait till #662 gets merged do rebase again.

guitorri:
Also, the color picker button can be the one from #672.

guitorri:
Is it absolutely necessary to create the [data set].Markers.dat file?
Shouldn't the diagram or marker objects be extended to accommodate such delta Markers?
I mean, this should be part of the schematic file somehow.
What about backward compatibility?
It might be interesting to have a visual feedback on the diagram indicating the delta. An arrow or a line. I need to thing about it.
Other than that, it works just fine. 👍

felix-salfelder:
#314 ... perhaps with the refactored marker code, this will be much easier.

extra marker files will definitely fall on our feet when refactoring the data sources.

in3otd:
I did not actually test the code but I do not like that the line connecting the marker box and the graph has the same color as the graph line. The colored box around the marker already gives a visual cue about which marker belongs to which line and IMHO everything that has the color of a graph line inside a diagram must come from the data. Then one can think of pathological cases where the marker line superposes/interferes with the actual waveform shape (for rectangular/triangular waves, etc.). Opinions?

Extra bonus: it may be nice to have the delta markers with *two* lines, one to each point used for computing the delta; this will make immediately clear where it is actually measured. I guess this is what you were referring to above, guitorri?
And maybe a delta marker box should be black or multi-color, since potentially referring to different graphs (as in the example above)

andresmmera:
>I did not actually test the code but I do not like that the line connecting the marker box and the graph has the same color as the graph line. The colored box around the marker already gives a visual cue about which marker belongs to which line and IMHO everything that has the color of a graph line inside a diagram must come from the data. Then one can think of pathological cases where the marker line superposes/interferes with the actual waveform shape (for rectangular/triangular waves, etc.). Opinions?

Good point. I didn't think about that case. Anyway, the default color can be changed. 

> Is it absolutely necessary to create the [data set].Markers.dat file? Shouldn't the diagram or marker objects be extended to accommodate such delta Markers?

Hum,... I didn't found another way to store the marker data, but I need to think about it. 
By the way, I noticed that the Diagram.cpp and other classes are quite entangled... or at least they seemed entangled to me. Usually, relying on third party libraries is not a good idea... but, *in the long term* it could be interesting to replace the painting routines by [the Qt charts module](https://doc.qt.io/qt-5/qt-charts-module.html) or even [matplotlib](http://matplotlib.org/) or [qcustomplot](http://qcustomplot.com/).

felix-salfelder:
> replace the painting routines

that (among others) was the intent of the new marker code and the graph_deque. need to check what state it is in right now.

(iirc, the 2d plots are already using qt routines for the drawing, which was a fairly big lift)

andresmmera:
Hello,

I'm working again on this branch. I totally agree that it is not convenient to create a separate file to store markers: I'm thinking about a practical case of sharing a schematic by just copying & pasting it as raw text... Using an additional file to save the markers would force to share the markers file too and it would be quite annoying...

I'll think about an alternative solution...

andresmmera:
I need to do some cleanup, but so far:
* I managed to get rid of those Marker.dat files
* The line between the trace and the marker box is always black
* It was added an extra line to the delta marker so as to give a visual clue of the reference marker
* Backwards compatibility is guaranteed

andresmmera:
**New appearance**
selection_003

###### **Backwards compatibiliy:**

**1. An old .dpl file can be loaded by this PR**

*Sheet created with the current Qucs code (master)*
selection_001

*Appearance of the previous file in this PR*
selection_002

**2. A .dpl file created with this PR can be loaded by the current Qucs code**
*Display sheet created with this PR*
selection_004

*Appearance of the previous sheet in the current Qucs (master)*
selection_005

in3otd:
thanks!
On my machine changing the Color and Border width does not change the appearance, does it work there?
I tried to remove the marker name assigning an empty string but an empty line is shown in the marker box, is this normal/expected?
`<picky mode on>`Can we have the previous brownish color as default instead of black?`<picky mode off>`

I got it to segfault a couple of times opening an exiting Data Display but it doesn't happen all the time; try to open and close the Data Display in the project below a couple of times - here it crashes usually the first or second time.
[markercrash.zip](https://github.com/Qucs/qucs/files/949034/markercrash.zip)

I also notice a message on the console 
```
Warning: QLayout: Attempting to add QLayout "" to SaveDialog "", which already has a layout
```
not sure if it was there before or it appeared with this PR.

andresmmera:
> On my machine changing the Color and Border width does not change the appearance, does it work there?

No, it was a bug... Now it is fixed

> I tried to remove the marker name assigning an empty string but an empty line is shown in the marker box, is this normal/expected?

I didn't think about that... I've just added a check to as to avoid empty names. Thank you

> <picky mode on>Can we have the previous brownish color as default instead of black?<picky mode off>

Of course, I've set the default marker color to 'Qt::darkMagenta'

> I got it to segfault a couple of times opening an exiting Data Display but it doesn't happen all the time; try to open and close the Data Display in the project below a couple of times - here it crashes usually the first or second time.
markercrash.zip

I cannot reproduce that segfault here, but I noticed that when opening that .dpl file, the console output is filled with warnings like this:
dangerous: (maybe) overwriting control token60.2409

These warnings also appear in the master and develop branches. Maybe they're related to the segfault... or not

> I also notice a message on the console
> Warning: QLayout: Attempting to add QLayout "" to SaveDialog "", which already has a layout
`not` sure if it was there before or it appeared with this PR.

Uhm, I've seen that this also happens in the develop branch (but not in master). The warning is thrown just when you pick a qucs project from the projects tab. In this sense, I've been tracking it down and it seems to appear at  2e9f7d1

in3otd:
thanks!
I can still see some segfault here; for the example above a `gdb` backtrace points to line 668 below, which I'm not sure is correct since the index `nVarPos` may point past the end of the vector but I did not really look at the whole code.
```
Thread 1 "qucs" hit Breakpoint 2, Marker::load (this=0x11d44c0, _s=...) at marker.cpp:668
668	    qDebug() << VarPos[nVarPos++];
(gdb) list
663	  do {
664	    j = n.indexOf('/', i);
665	    VarPos[nVarPos++] = n.mid(i,j-i).toDouble(&ok);
666	    if(!ok) return false;
667	    i = j+1;
668	    qDebug() << VarPos[nVarPos++];
669	  } while(j >= 0);
670	
671	  n  = s.section(' ',2,2);    // x1
672	  x1 = n.toInt(&ok);
```

Anyway, with another example the backtrace points to a call to `ActiveMarkers.clear()` in `diagram.cpp`

```
#0  0x0000000000423c5c in QBasicAtomicInt::deref() (this=0x0)
    at /usr/lib64/qt/include/QtCore/qatomic_x86_64.h:128
#1  0x0000000000423a1f in QString::~QString() (this=0x31ac9d0)
    at /usr/lib64/qt/include/QtCore/qstring.h:880
#2  0x0000000000609b3f in QMap<QString, std::vector<double, std::allocator<double> > >::freeData(QMapData*) (this=0x7fff3f7333b0, x=0x31ac620) at /usr/lib64/qt/include/QtCore/qmap.h:651
#3  0x000000000060864d in QMap<QString, std::vector<double, std::allocator<double> > >::~QMap() (this=0x7fff3f7333b0) at /usr/lib64/qt/include/QtCore/qmap.h:185
#4  0x000000000060870a in QMap<QString, std::vector<double, std::allocator<double> > >::clear() (this=0x31ac2f0) at /usr/lib64/qt/include/QtCore/qmap.h:446
#5  0x00000000005fc2ba in Diagram::paint(ViewPainter*) (this=0x31ac140, p=0x7fff3f733870)
    at diagram.cpp:141
#6  0x0000000000472829 in Schematic::drawContents(QPainter*, int, int, int, int) (this=
    0x31aa530, p=0x7fff3f733930) at schematic.cpp:427
#7  0x00007fd26c8c3fff in Q3ScrollView::drawContentsOffset(QPainter*, int, int, int, int, int, int) () at /usr/lib64/libQt3Support.so.4
#8  0x00007fd26c8c4257 in Q3ScrollView::viewportPaintEvent(QPaintEvent*) ()
    at /usr/lib64/libQt3Support.so.4
#9  0x00007fd26c8c5bc0 in Q3ScrollView::eventFilter(QObject*, QEvent*) ()
    at /usr/lib64/libQt3Support.so.4
#10 0x00007fd26dd20d76 in QCoreApplicationPrivate::sendThroughObjectEventFilters(QObject*, QEvent*) () at /usr/lib64/libQtCore.so.4
#11 0x00007fd26d09346c in QApplicationPrivate::notify_helper(QObject*, QEvent*) ()
    at /usr/lib64/libQtGui.so.4
```
last debug messages on terminal were
```
Debug: ---------MARKERS MAP---------- 
Debug: Re{y}, Im{y}, x_position, y_position, indep_var, Delta_Mode? 
Segmentation fault (core dumped)
```

The segfault does not happen very often, usually when switching between Data Display tabs with lot of graphs

qucs/qucs/diagrams/marker.cpp
```diff
@@ -363,18 +496,24 @@ void Marker::paint(ViewPainter *p, int x0, int y0)
   x2_ = p->drawText(Text, x0+x1+3, y0+y1+3, &y2_);
   x2_ += int(6.0*p->Scale);
   y2_ += int(6.0*p->Scale);
-  if(!transparent) {
```
in3otd: why the logic of 'transparent' was changed? Previously the marker border was always drawn and 'transparent' referred to the rectangle area, which was not transparent by default.
Now selecting 'transparent' removes the border and makes the rectangle area non-transparent.

andresmmera:
1. Concerning the nVarPos, if I replace 'nVarPos++' by 'nVarPos-1', qucs crashes.
I think that the segfault problem comes from here: https://github.com/Qucs/qucs/pull/407/commits/3374d90a2836b5286eb64a5eb57e4d2680222e83#diff-d6888ff24dbf42ac6c04ad6e983c02bcR665 By comparing this code with the master branch, I've noticed that I've accidentally added an extra autoincrement operator... This is equivalent to set nVarPos+=2 so... for sure this is the source of the problem...

2. Well, I cannot remember the reason why I've changed the transparency logic. Probably, I was doing some kind of test and... I've forgotten to undo the changes. Sorry for that :sweat:

in3otd:
doing some more testing, found a little bug:
- create a delta marker
- change the name of the reference marker
- double click on the delta marker... `qucs` crashes

I guess it looks/points to the old reference marker name and so something goes wrong

in3otd:
I also see that marker names with spaces do not work... I'll take a look at this.

in3otd:
I've fixed the problems above but some more general issues remain
- when changing a marker name, if there is a delta marker using it as a reference the reference name it shows is not updated
- similarly, if a marker used as reference is deleted this has no effect on the delta marker using it (I think it should actually change the marker type back to a normal marker.
- same if the marker used as reference is changed from normal to delta, this has no effect on the other marker(s) using it as reference

need to think how all this can be handled

andresmmera:
Hi,
@in3otd I've just pushed a fix for the issues you described.

https://drive.google.com/open?id=0BwDjrwGfd_z7WEdLNl81RXdVS0E

in3otd:
I've pushed some commits, one allows spaces in the Marker ID string: you may need to recreate/modify any existing schematic using Enhanced Markers as the marker ID is now enclosed in quotes in the schematic file.

Another thing I did is to keep the markers map in the diagram object only and not in every marker; I didn't see the need to have a copy in every marker and thought that only the markers parent needs to know all the markers characteristics.

There are some other points that in the long term may need to be addressed, e.g. now you can create delta markers between different graphs even if this does rarely make sense IMHO but nothing urgent at present.

andresmmera:
I've just pushed two commits. The first fixes a bug: the reference marker combobox was being activated after placing three or more markers

Then after rebasing against develop, I made minor changes to extend the delta marker feature to the recently added phasor and AC temporal diagrams

andresmmera:
>  e.g. now you can create delta markers between different graphs even if this does rarely make sense

The last commit fixes this

andresmmera:
In line with #701 it was added a impedance/admittance selection box for those markers pointing at a Smith chart diagram

andresmmera:
I've applied the fix https://github.com/Qucs/qucs/pull/701/commits/428b6870769580e486731a383410da40b6fc65b9 here...

guitorri:
Testing.

in3otd:
IIRC from my latest testing, it was working fine - I was modifying the code to make, IMHO, a little straightforward; now the markers info are kept in the Diagram, moving that to the Graphs will automatically remove the possibility to create cross-graph delta markers, without the need to check on which Graph the marker is, since every Graph will have separate marker info, only for their own markers. But these modification can be done later.

guitorri:
I checked out the `ImprovedMarkers` branch.
Openned an existing .sch
Added a marker to a diagram and saved.
It opens fine on 0.0.19, but there is an extra pair of quotes being saved on the `Rect` diagram:

```
[master]~/git/qucs/qucs-test/testsuite/TR_sawtooth-discreet_prj $ git diff .
diff --git a/testsuite/TR_sawtooth-discreet_prj/sawtooth-discreet.sch b/testsuite/TR_sawtooth-discreet_prj/sawtooth-discreet.sch
index 5576391..97627de 100644
--- a/testsuite/TR_sawtooth-discreet_prj/sawtooth-discreet.sch
+++ b/testsuite/TR_sawtooth-discreet_prj/sawtooth-discreet.sch
@@ -1,4 +1,4 @@
-<Qucs Schematic 0.0.17>
+<Qucs Schematic 0.0.19>
 <Properties>
   <View=-4,-29,786,547,1.15104,0,0>
   <Grid=10,10,1>
@@ -17,7 +17,7 @@
 </Symbol>
 <Components>
   <_BJT T3 1 320 210 -81 -53 0 2 "pnp" 1 "1e-16" 1 "1" 1 "1" 0 "0" 0 "0" 0 "0" 1 "0" 0 "0" 0 "1.5" 0 "0" 0 "2" 0 "100" 1 "1" 0 "0" 0 "0" 0 "0" 0 "0" 0 "0" 0 "0" 0 "0.75" 0 "0.33" 0 "0" 0 "0.75" 0 "0.33" 0 "1.0" 0 "0" 0 "0.75" 0 "0" 0 "0.5" 0 "0.0" 0 "0.0" 0 "0.0" 0 "0.0" 0 "0.0" 0 "26.85" 0 "0.0" 0 "1.0" 0 "1.0" 0 "0.0" 0 "1.0" 0 "1.0" 0 "0.0" 0 "0.0" 0 "3.0" 0 "1.11" 0 "26.85" 0 "1.0" 0>
-  <Vdc V1 1 50 80 18 -26 0 1 "15 V" 1>
+  <Vdc V1 1 50 80 18 -26 0 1 "15 V" 1 "0 Ohm" 0>
   <GND * 1 50 110 0 0 0 0>
   <R R1 1 130 150 15 -26 0 1 "220 Ohm" 1 "26.85" 0 "0.0" 0 "0.0" 0 "26.85" 0 "european" 0>
   <GND * 1 190 480 0 0 0 0>
@@ -67,10 +67,11 @@
   <320 280 320 280 "OUT" 340 250 0 "">
 </Wires>
 <Diagrams>
-  <Rect 490 347 248 137 3 #c0c0c0 1 00 1 0 1e-05 3e-05 1 -0.380103 2 4.78441 1 -1 1 1 315 0 225 "" "" "">
+  <Rect 490 347 248 137 3 #c0c0c0 1 00 1 0 1e-05 3e-05 1 -0.380103 2 4.78441 1 -1 1 1 315 0 225 "" "" "" "">
        <"OUT.Vt" #0000ff 0 3 0 0 0>
+         <Mkr 1.78845e-05 208 -156 3 0 0 "Marker0" #800080 1 0 "" 0#0#0#0#0 0 0>
   </Rect>
-  <Rect 490 465 249 85 3 #c0c0c0 1 00 1 0 1e-05 3e-05 1 -7.22877 20 20 1 -1 2 1 315 0 225 "" "" "">
+  <Rect 490 465 249 85 3 #c0c0c0 1 00 1 0 1e-05 3e-05 1 -7.22877 20 20 1 -1 2 1 315 0 225 "" "" "" "">
        <"UTOP.Vt" #ff0000 0 3 0 0 0>
   </Rect>
 </Diagrams>
```

Any idea why?

in3otd:
um, I guess it actually comes from [here](https://github.com/andresmmera/qucs/blame/ImprovedMarkers/qucs/qucs/diagrams/diagram.cpp#L1406)

guitorri:
You guys are more familiar with the marker code. 
Can you give it a try to rebase or merge develop into this?

Please document the latest marker line format in [qucs-help](https://github.com/Qucs/qucs-help/blob/develop/source/internal.rst).

felix-salfelder:
@guitorri need #370 first. #370 sanitizes the way graph data is stored/accessed/processed and how markers are attached to a graph. that particularly fixes  moving around markers.

Murmele:
Are there plans to write also the absolute values of the point where the delta mode marker is, to the label?

PR #412, opened 2016-01-21
Add support for using preferred number series with the ASCO optimizer + variables search string fix

in3otd:
The ASCO optimizer used by Qucs supports constraining optimization variables to assume discrete values in the [IEC 60063 preferred number series] (https://en.wikipedia.org/wiki/Preferred_number#E_series) but this option was not available in the optimization dialog (I guess the dialog was written before ASCO added support for this), so here is a simple modification to allow this selection.

I have also added an example using this feature of the optimizer. Note that for some reason the ASCO installed on my Linux system with Qucs 0.0.18 throws an error with this schematic, saying that the variable(s) is outside the allowed range. Recompiling ASCO from the sources, it works as expected. I guess this is due to some numerical precision issue, but did not have time to investigate further.

While trying to add the new `.sch` example file, I noticed that all the `.sch` files were ignored by `Git`, apparently due to [a rule] (https://github.com/Qucs/qucs/blob/master/.gitignore#L186) added to `.gitignore` some time ago.
I have modified the rule to not ignore the `.sch` in `examples/` and tried to clean up `.gitignore` a bit, but I realized we should probably think a bit more on how to handle this in general, as we have sometimes also `gitignore` files in subdirectories repeating the same rules...


in3otd:
I have added another commit, still related to ASCO, but not to the previous two. Since it's a quick fix requiring just one change I thought creating a new PR just for this was too much :grin: 

Some time ago a user reported strange results from a simulation involving the ASCO optimizer, in that the optimization ended immediately and with a big "cost function" value for no apparent reason.
(You can see this problem in [this schematic] (https://gist.github.com/in3otd/7f87e61ec2fb168ca064))

It turns out that this is due to the way ASCO works for extracting the Goals variables values from the text dataset. In short, for every variable Qucs prepares a files for ASCO describing how to extract the variable value from the dataset and in particular it tells ASCO to look for the string "`<indep VariableName`". The problem is that ASCO matches this also for lines like
```
<indep VariableName1 ...
<indep VariableName2 ...
<indep VariableNameEtc ...
```
so if there are other variables starting with the same name as a Goal variable, ASCO becomes completely confused and does not perform the optimization properly.

The fix is simple, just pass the correct search string to ASCO, which needs to have a space after the variable name, like "`<indep VariableName` " 

...so the fix is just really adding one space to the code, took much more time to actually *describe* the problem :grin:

in3otd:
rebased to get a working AppVeyor setup

in3otd:
added  the 0.0.19 label - at least for the 03fef76de07cdc8e66208a87a8a5442027f77dc2 bug fix, the other commits can go in develop if you prefer.

guitorri:
Merged into release-0.0.19 (0e279affc6d81ccd8b4aae862c23f0d7bcf63169). Thank you.

PR #413, opened 2016-01-22
Move AppVeyor build from windows CMD to MSYS2.

guitorri:
Setting up dependencies on Windows is simpler and faster with MSYS2. 
MSYS2 typically provides the latest packages, eg. gcc 5.2.0.
The previous build script (using windows cmd and CMake) can be added later by using a build matrix to enable parallel builds in distinct environments.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -2,126 +2,86 @@
 platform:
   - x64
 
+cache:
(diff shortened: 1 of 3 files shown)


guitorri:
I did some minor changes and a forced update.

PR #416, opened 2016-01-25
RF lumped components

andresmmera:
- [x] add INDQ, CAPQ, spiral inductor, planar circular loop inductor
- [x] new palette for RF passive devices
- [x] document INDQ, CAPQ
- [x] document spiral inductor, planar circular loop inductor

---
Hello, 

This PR adds two components to the "Lumped Components" palette. INDQ and CAPQ components mimic the S-param behaviour of inductances/capacitors whose quality factor is finite. So far, these models can only work with the S-param simulation. Regarding the implementation, the S matrix is derived from the ABCD matrix, which is easily calculated from the impedance of a series LR/CR circuit.

These components also include three frequency weighting profiles that determine the value of the series resistor versus frequency.

Regards

selection_012


in3otd:
:clap:  :clap:  ...I was thinking since long time that it would be good to add some RF models of real components...

Some general random comments:
- we may need to add a "RF components" (or similar) tab to the GUI, as I guess more models like this could/should be done in the future (what about L with parasitic Rs and Cp, C with parasitic Ls and Rp or a transmission-line model for coils?); the "lumped components" tab is becoming a bit too crowded
- ok for using the ABCD matrix and leaving the conversion to S and Y to Qucs, but writing directly the Y or S matrices for series components like these should give expressions almost as simple, I guess
- maybe the components description should start with "Inductor" or "Capacitor", "Q inductor" or "Q Capacitor" sound a bit weird to me.
- we do not have a single, uniform, coding style in Qucs but... you are using multiple styles :grin: (braces positions, indents, spaces around operator). (@guitorri, what about some Coding Conventions?)

guitorri:
@in3otd please open an issue so we can discuss about the coding convention. IMHO it should be something automated... I hate the idea of constantly nitpicking peoples commits for that.

@andresmmera please keep aside (gist perhaps) an sch example, so we can add later to qucs-test.

andresmmera:
@in3otd Hell, PROP_NO_VAL="Proportional to freq" should not be there...
I'm gonna apply your suggestions tomorrow (now, I need to sleep) (I *completely* agree with you). Thank you for taking a look to the code.

*we may need to add a "RF components" (or similar) tab to the GUI, as I guess more models like this could/should be done in the future (what about L with parasitic Rs and Cp, C with parasitic Ls and Rp or a transmission-line model for coils?); the "lumped components" tab is becoming a bit too crowded*

I've started coding models equivalents circuit of planar spiral inductances, MIM and interdigital capacitors. The procedure is much the same: EC -> ABCD matrix -> S/Y matrix. I'd like to implement this stuff in Verilog-A since it seems to be clearer. Unfortunately, it seems that frequency dependent parameters are not supported... 

Besides this, I've noticed that Murata provides quite accurate models for their capacitors and inductances. I totally ignore copyright issues but it would be so nice to bring them to Qucs. 
Take a look to http://ds.murata.co.jp/software/simsurfing/en-us/index.html?intcid5=com_xxx_xxx_cmn_nv_xxx# . If you select a part, a button on the top takes you to the EC netlist. It's quite cool

in3otd:
@andresmmera , honestly how `PROP_NO_VAL` is used is not completely clear to me; my guess is that it's used to assign a default value if the property is not found (e.g. for backward compatibility when expanding an existing component with new properties). In this case it may not make any difference...

Regarding noise in lossy components, in Qucs this is usually computing using Bosma's theorem, as described in the [Technical Papers] (http://qucs.sourceforge.net/tech/node12.html), so may be not too difficult even without using Verilog-A.

Can you try (in a future PR) to create a "RF components" tab and move there your new components and maybe also the Isolator, Circulator, Coupler, Hybrid, Amplifier (RF EDD too?), since these latter are rarely used, except for RF stuff, I think.

Yes, the Murata library is nice, but I think we cannot distribute their models, from a quick look at [their license] (http://www.murata.com/en-us/tool/download/mcsil/download). But at least the models are easy to find on the net...

@guitorri yes, the code formatting should be automated, we briefly discussed this [some time ago] (http://sourceforge.net/p/qucs/mailman/message/33005913/), I did at the time some quick check with `clang-format -style=LLVM` and it seemed to work well, except for large comment block where it sometimes messed up the alignment. Would be nice maybe to use this as a commit hook in git, not sure it will work well enough for leaving out any manual adjustment.

I too try to avoid nitpicking for the coding style, but sometimes after going back and forth several times between the Qucs GUI and the qucsator code during a long evening, another coding style variation is difficult to follow for me :grin:

andresmmera:
@Claudio. Indeed that string indicates the default value. The thing I had
that component in a sloppy branch and I've made a mistake while porting the
code. I'll correct all these issues during the evening

andresmmera:
I think I've corrected all the issues above. 
According to @in3otd's suggestion, I've replaced "Q Capacitor" by "Capacitor with Q". Since the beginning I've believed that "Capacitor with Q" is more accurate than "Q Capacitor", but... the thing is that the text is too long and it is not shown in the palette.

screenshot from 2016-01-26 17 58 41

andresmmera:
I think that something is wrong with TravisCI and AppVeyor:
Travis gives the following error:

Error: Failed to download resource "adms"
Download failed: http://superb-dca2.dl.sourceforge.net/project/mot-adms/adms-source/2.3/adms-2.3.4.tar.gz
The command "if [[ $OSX ]]; then brew tap guitorri/tap; brew install adms; fi" failed and exited with 1 during .
Your build has been stopped.

On the other hand, AppVeyor gives:

:: Retrieving packages ...
error: failed retrieving file 'pacman-mirrors-20150722-2-any.pkg.tar.xz' from downloads.sourceforge.net : Connection timed out after 10932 milliseconds 
warning: failed to retrieve some files
error: failed to commit transaction (download library error)
Errors occurred, no packages were upgraded.
Command exited with code 1

guitorri:
SourceForge must be down. We have to setup a cache or fallback mechanism. I will restart it for you.

in3otd:
...I did restart Appveyor already :)
Have (local) issues with Travis

andresmmera:
Hello, 

I've added the planar spiral inductor model to the new "RF passive components" palette. This time the equivalent circuit involves shunt capacitors, so the S-matrix becomes quite complex... Indeed, It involves lots of '*' operations, so it seems better to calculate the ABCD matrix and then use qucs::atos() to get the S matrix. You can take a look to the S-matrix [here](https://dl.dropboxusercontent.com/u/21294824/Smat%20vs%20ABCD)

selection_013

References:
 [1] I. Bahl, Fundamentals of RF and Microwave Transistor Amplifiers, John Wiley and Sons, 2009.  
 [2] Simple Accurate Expressions for Planar Spiral Inductances. S. S. Mohan. IEEE Journal of Solid State Circuits.   Vol 34, No 10, October 1999

andresmmera:
Now, a single turn planar inductor... (The equivalent circuit is the same as above)

selection_014

andresmmera:
Hello, 

I've just added a coil to the "RF passive components" palette. The equivalent circuit is:
selection_017


where R includes the skin effect and the hysteresis losses and C represents the series equivalent of the turn-to-turn capacitance.

selection_016

References:
 [1] Marian K. Kazimierczuk, “High-Frequency Magnetic Components”, John Wiley & Sons, Ltd. 2009 
 [2] Alberto Reatti and Marian K. Kazimierczuk, „Comparison of Various Methods for Calculating 
      the AC Resistance of Inductors”, IEEE Transactions on Magnetics, Vol. 38, No. 3, May 2002 
 [3] Gabriele Grandi, Marian K. Kazimierczuk, Antonio Massarini, and Ugo Reggiani, „Stray Capacitances of Single-Layer  Solenoid Air-Core Inductors”, IEEE Transactions on Industry Application, Vol. 35, No.5, September/October 1999
 [4] Peter Csurgai, Miklos Kuczmann, Comparison of various high-frequency models of RF chip inductors, 
 “Szécheny István” University, Lab. of Electromagnetic Field, Dept. of Telecommunications, Hungary

in3otd:
interesting... I never saw before the "improved solenoid formula" cited in [4] and would be nice to know what improvement it brings over the well-known Wheeler formula. Usually when looking for something even better than this latter, one uses the Lundin's formula and also this is for computing the inductance for the usual "current sheet" approximation, not for a real coil. So I wonder if we really need to include the "improved solenoid formula".
Then someone did a huge [study about the coils self-resonance] (http://www.g3ynh.info/zdocs/magnetics/appendix/self_res/self-res.pdf) where it's said that the parasitic capacitance formula from Grandi et al. is wrong, and it will be interesting to report this somewhere. (BTW, we need a way do have a proper documentation for every component, I'll open an issue on this ine day)
I find also strange that an article about RF chip inductors does not say anything about the variation of the inductance with frequency.

I quickly tried to do a simulation with this new coil model but `qucsator` crashed (`qucsator: matrix.cpp:818: qucs::matrix qucs::inverseGaussJordan(qucs::matrix): Assertion MaxPivot != 0 failed.`), did not have time to investigate. BTW, please make your test schematic available somewhere (using a `Gist` maybe)

`Coil_Radius` and `Cond_Radius` may be better called simply `R` and `r`, the property description tells what these are exactly. This is similar to what is done for the coaxial lines or bifilar lines.
And personally I prefer using the diameters; when I wind a coil, wires and coil formers are usually specified in terms of diameters and it's the quentity easily measured with a caliper :grin:

guitorri:
Right, please keep the examples and test schematics. 
We should not add more stuff to the core without checking for coverage of the new code.

andresmmera:
Hello, 

You can find the schematics here: https://gist.github.com/andresmmera/ce4a5123f60d6544216a

@in3otd Thank you very much for your help. I'm gonna take a look to that report. I'll try to find better references too.

guitorri:
I did a rebase and conflict resolution of this one.
I could not do force push to your branch. 
You can either enable "Allow edits from maintainers" (see right hand side of your PR), or I can push it to my fork and you pull and do the force push yourself. 
We should avoid open/close/reopen PRs. It messes up the numbering. 😄

andresmmera:
@guitorri I've just allowed edits from maintainers... I think I'd better do this for the other PRs, just in case...
I'll try to work on Qucs tomorrow...
Thanks again... :-)

guitorri:
Thank you. Branch updated.
I have yet to run your test schematics.
I think we should document also the circular and planar inductor.

andresmmera:
This reminds me that the new tapered line component should be documented too. PR#657

guitorri:
From Travis/Linux/Clang instance:

```
! LaTeX Error: File `capq' not found.

 ...                                              
                                                  
l.356 \includegraphics[width=2.5cm]{capq}
                                         

! LaTeX Error: File `indq' not found.

 ...                                              
                                                  
l.373 \includegraphics[width=2.5cm]{indq}
```

It builds on my Mac...

qucs-core/src/components/microstrip/spiralinductor.cpp
```diff
@@ -0,0 +1,147 @@
...
+ nr_double_t W = getPropertyDouble ("W");//Width
+ nr_double_t S = getPropertyDouble ("S");//Spacing between turns
+ substrate * subst = getSubstrate ();
+
+ nr_double_t Do = Di + 2.*N*W + (2.*N-1)*S;
+ nr_double_t chi = (Do - Di) / (Do + Di);//Fill ratio
+ nr_double_t a = (Di+Do)/4.;
+ nr_double_t Dav = .5*(Do+Di);
+  
+ nr_double_t h = subst->getPropertyDouble ("h");
```
guitorri: This variable seems to be unused.

andresmmera:
I've just updated the documentation and fixed the unused variable.
Once being there, I also added the DC model for the printed inductors.

qucs-core/src/components/indq.cpp
```diff
@@ -43,29 +43,27 @@ indq::indq () : circuit (2) {
...
-  calcABCDparams(frequency);
-  matrix Stmp = qucs::atos(ABCD, z0, z0);
-  setMatrixS(Stmp);
+ nr_double_t Rs = 0;
+ if ((f!=0) && (Q!=0))
+ {
+   nr_double_t Qf=Q;
+   if (!strcmp (getPropertyString ("Mode"), "Linear")) Qf*=frequency/f;
+   if (!strcmp (getPropertyString ("Mode"), "Quadratic"))Qf*=qucs::sqrt(frequency/f);
+   Rs = 1./(2.*pi*frequency*L*Qf);
```
in3otd: oops, the `Rs` formula is wrong here, you got it right in the previous commit comment, though :grin:

in3otd:
...so, where did we left off here? :grin:

some minor things:
- the *RF passive components* tab name may not need to be *passive*, strictly speaking the EDD/RFEDD can represent any transfer function
  - the *Gyrator* does not belong to this tab, I think
- after changing from the *ABCD* matrix intermediate representation to the S-parameters one, the `calcSP()` function is somewhat similar to the one in `inductor.cpp` and `capacitor.cpp` but not completely aligned, do you think it's worth to align them?
- the case `L=0` for the `INDQ` component is not handled for the AC analysis and the results are wrong (cfr. the code in `inductor.cpp`)
- would be nice to add the S-parameters noise analysis for the lossy components, using the Bosma's theorem, as mentioned above (see page 26 of `technical.pdf` and the code in `coaxline.cpp`)
- out of curiosity, why one of the "Q frequency profiles" of `INDQ` is called "quadratic" but the dependency is (correctly) a square root actually? Does some other simulator use this convention?

let me know if you lack time to address these points, I might be able to do that in the coming days/weeks

andresmmera:
> out of curiosity, why one of the "Q frequency profiles" of INDQ is called "quadratic" but the dependency is (correctly) a square root actually? Does some other simulator use this convention?

I was sure I took that from the ADS help page... but it clearly says 'proportional to sqrt(freq)'. Of course sqrt(f) is not a quadratic form so I'd better change that misleading label...

I'll try to address these issues ASAP, maybe at the weekend. Anyway, feel free to work on that if you wish :-)

in3otd:
I've pushed some commits that address the points above: I have done some test and everything seems fine here, of course further testing is welcome.
We will need to add also some schematics to `qucs-test` for these new components.

guitorri:
Is there something else to be done here? The tests are running ok. 
There are a few conflicts to be resolved. I tried the web editor, but did not commit as I was unsure if there is more working to be done.

in3otd:
ah, it was clean a few days ago... let me fix a couple of things in the docs which I just spotted, the rest is fine IMHO.

in3otd:
had to correct the impedance at DC calculations  used for AC and SP simulations.

Will push also an update to the `qucs-test` submodule once things are stable there...

PR #417, opened 2016-01-25
Fix docs build from tar

guitorri:
* includes to distdir some pdf figures needed by qucs-doc. 
* do not copy qucs-test/.git/ directory into distdir.


felix-salfelder:
thanks.

PR #423, opened 2016-01-27
Remove adms submodule

guitorri:


felix-salfelder:
> shouldn't configure fail saying that admsXml was not found ?

that sounds like a feasible workaround. in the long run, we should

* disable adms-related stuff if admsXml is not available (and still compile)
* turn adms-related stuff into an extension (build time first, then run-time)
* implement proper run time extension package
* remove all adms dependencies from the main package.

once admsXml stuff works independent to the main package, anybody can use/implement his favourite model compiler without polluting the code base.

NB: i should open an issue with fancy tick boxes, but i am in a hurry.

PR #427, opened 2016-01-28
Update authors and date in the About dialog

in3otd:
Some time has passed since the last update (creation) of the About dialog, so the release year had to be changed :grin: And also update authors' names there and in the AUTHORS file

Removed also the names from the translatable strings.

### DIFF
--- a/qucs/AUTHORS
+++ b/qucs/AUTHORS
@@ -17,12 +17,29 @@
 -- Boston, MA 02110-1301, USA.  
 --
 
-Maintainer:
-	Michael Margraf <michael.margraf@alumni.tu-berlin.de>
+Original developers:
+	Michael Margraf
+	Stefan Jahn
 
-Authors: (see THANKS file for details)
-	Michael Margraf <michael.margraf@alumni.tu-berlin.de>
-	Stefan Jahn <stefan@lkcc.org>
-	Jens Flucke <jens.flucke@gmx.de>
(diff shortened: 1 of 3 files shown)


qucs/AUTHORS
```diff
@@ -17,12 +17,29 @@
...
+Current developers:
+	Guilherme Brondani Torri
+	Mike Brinson
+	Richard Crozier
+	Bastien Roucaries
+	Frans Schreuder
+	Vadim Kuznetsov
+	Claudio Girardi
+	Felix Salfelder
+	Andr�s Mart�nez Mera
```
guitorri: Why are these symbols showing up here? Can we do something about it? I mean the file encoding or something.

felix-salfelder: perhaps, we should just use utf8 here...

andresmmera: ... Or even write (i, e) instead of (í, é). It is not so important.

in3otd: ...yep, I saw them but the strange thing is that if you then click on "View" (top right) GitHub displays the content properly, so they are not assuming the same encoding everywhere (?)

I have the terminal encoding to Latin1 here, so it looked fine to me :grin:
Switching to UTF-8 you get the characters above.

Should we use just 7-bit ASCII in text files like this?

in3otd:
for the moment used (i, e) instead of (í, é) in the AUTHORS file

PR #429, opened 2016-02-01
Cherry pick PR#206

guitorri:
- Apply patches from #206.
- Needed new PR to so Travis can pickup the updated qucs-test submodule.


guitorri:
Nope. Using golden references from OSX fail the numerical comprison on Linux. Need to look more closely...

PR #436, opened 2016-02-02
revive --with-mkadms

felix-salfelder:


guitorri:
Not sure what is going on here.
On windows `admsXml` is on the path but the config is failing to find it.
```
...
./configure --enable-maintainer-mode --with-mkadms=`which admsXml`"
configure: WARNING: unrecognized options: --enable-maintainer-mode
checking for gcc... gcc
...
checking for gperf... /usr/local/bin/gperf
checking for /usr/local/bin/admsXml... no
configure: error: need admsXml, but not found. consider --with-mkadms=<alternative admsXml>
```

Can you please rebase and add a commit to .appveryor? 
Please remove `--enable-maintainer-mode` and add `--disable-dependency-track`

guitorri:
I will try to reproduce this tomorrow on my Windows machine.

felix-salfelder:
thanks. I have no clue once again...

felix-salfelder:
apparently, this is how AC_PATH_PROG is meant to work. does not permit full paths (?!)

felix-salfelder:
this is not exactly what i expected. will have another look later today.

guitorri:
#438 seems good to me. Please have a look. 
The difference is that the binary DIR is given, not the full path to the admsXml executable. So it is not possible to use a renamed admsXml

felix-salfelder:
fixed (hopefully). see http://stackoverflow.com/questions/26034613/how-to-check-if-a-full-path-executable-is-correct-in-autoconf

PR #437, opened 2016-02-02
Travis include linux/clang/cmake

guitorri:


guitorri:
Nahhh. It is failing. Most likely the CMake tests for functions and compiler features are broken... I will take a look another day.

PR #439, opened 2016-02-03
Travis cmake

guitorri:


guitorri:
Nope. The qucs-core CMake seems to be broken on Linux for either gcc or clang.

PR #443, opened 2016-02-03
Add simple file selection for dat/dpl/m files in File Properties

in3otd:
I'm used to have several `.dat` files for every schematic and a few `*.dpl` too, so I wanted to have a way to select an existing `.dat` or `.dpl` (and `.m`) to use for the simulations using the mouse .
So I have added a small "file browser" for these files, with a button near every file type in the _File Properties_ dialog:

auxfilebrowse

The standard (system) file dialog cannot be used, since these files must be in the Project (schematic) directory and with a standard file dialog is not possible to avoid having the user select a file elsewhere.

Seems to work fine here but please double-check.

### DIFF
--- a/qucs/qucs/dialogs/settingsdialog.cpp
+++ b/qucs/qucs/dialogs/settingsdialog.cpp
@@ -17,6 +17,7 @@
 #include "settingsdialog.h"
 
 #include "node.h"
+#include "main.h"
 #include "qucs.h"
 #include "mnemo.h"
 #include "schematic.h"
@@ -35,7 +36,11 @@
 #include <QTabWidget>
 #include <QValidator>
 #include <QPushButton>
-
+#include <QFileSystemModel>
+#include <QTreeView>
+#include <QHeaderView>
+#include <QModelIndex>
+#include <QDebug>
 
 SettingsDialog::SettingsDialog(Schematic *Doc_)
     : QDialog(Doc_) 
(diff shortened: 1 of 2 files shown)


in3otd:
Can/should we put this in `0.0.19`? It would be quite useful, for me... :grin:

felix-salfelder:
"multiple .dat files" is going to be a problem, once .dat files are specific to a simulator backend.

i agree that something like this is probably useful, but the proposed way will only add extra headache to the proper simulator-backend handling.

i think, this sort of alternative 'data source' selection could well go to the options controlling each/single plot element for 0.0.19, but global backend-specific stuff must wait for >=0.0.20.

in3otd:
not sure I understand what's the impact of allowing multiple `.dat` files on the simulators backend development but this has been possible since "always"; here I'm just adding a "Browse" button to the already existing `.dat` file selection.

I usually have multiple `.dat` file for a single schematic (`.sch`), to keep the results of different variants/experiments and being able to plot them on a single diagram; the "Browse" button is to avoid having to type the name of an already existing file (and quickly remind me which `.dat` files already exist).
I thought this was a quite common thing but maybe not, since nobody implemented the "Browse" button before.

guitorri:
Sometimes I also have several `.dat` files. The code looks sane to me. I will merge this while we get our heads around the side effects of multiple simulator backends.

guitorri:
Oops, the PR was based on `master`...  
I just found out that it is now possible to [change the base branch in of a PR](https://github.com/blog/2224-change-the-base-branch-of-a-pull-request). Nice!

felix-salfelder:
> I just found out that it is now possible to change the base branch of a PR.

wow/about time. thanks for pointing that out!

in case my edit was not clear: yes, this is useful, except that it will probably break soon-ish. i already had the 'select data source' stuff implemented for the plots, but somehow, the code got ploughed under...

PR #448, opened 2016-02-05
Help button on the Edit Component Properties dialog

in3otd:
A simple idea to allow opening the online [Qucs Reference Manual](http://qucs.github.io/qucs-manual/0.0.19/html-en/index.html) (QRM) directly on the component description, when the Edit Component Properties dialog is open.
As an example, when the properties of a DIAC are being edited and the Help button is pressed

qucscompshelp

this will (try to) open the page http://qucs.github.io/qucs-manual/0.0.19/html-en/component_reference.html#compdiac

~At present this is **not working** since the web links are not right and some "details" need to be worked out:~
**Edit:** it works now, see latest comments.
- the current component links in the QRM are not directly linked to the component model or graphic, but are a simple counter of the component index
- at present the QRM contains a section for every component symbol, but we should probably have a section for every component model. E.g. we have a `resistor` and a `resistor_US`, which have different symbols but the same model. Same for `n-JFET` and `p-JFET` and several others.

Of course many improvements are possible, like
- linking the `F1` to the component help, even maybe when the component is just selected on the schematic
- point to the local copy of the QRM (which is not currently installed AFAIK)
- ...

### DIFF
--- a/qucs/qucs/components/componentdialog.cpp
+++ b/qucs/qucs/components/componentdialog.cpp
@@ -15,6 +15,10 @@
  *                                                                         *
  ***************************************************************************/
 
+#ifdef HAVE_CONFIG_H
+# include <config.h>
+#endif
+
 #include "componentdialog.h"
 #include "qucs.h"
 #include "schematic.h"
@@ -33,6 +37,8 @@
 #include <QComboBox>
 #include <QGroupBox>
 #include <QPushButton>
+#include <QUrl>
+#include <QDesktopServices>
 #include <QEvent>
 #include <QKeyEvent>
 #include <QDebug>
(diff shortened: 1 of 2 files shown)


in3otd:
@guitorri looking at the old commits, when you added the components data file generation (`qucs -doc`) there was a comment saying *it uses the bitmap file name to prefix the component data and property files* but this was later changed to use a simple numerical index as file prefix. Now I'll need a "unique key" to identify the component so I can open its help page and the component bitmap file name seems a natural choice for the key. Do you remember if there was any issue in using the component bitmap file name as key?

guitorri:
I think it is still the case that it is required to have one bitmap file per component.
I believe you cannot assign the same bitmap to multiple components.
Back then this did the trick for the qucs-doc proof of concept.
Maybe using the bitmap as key is not the best way. Perhaps the key should come from the implementation, where the component properties are defined.

in3otd:
> Perhaps the key should come from the implementation, where the component properties are defined.

yes, I think this would be better, in the long term.

Anyway, I realized that using the component bitmap file as key is not a practical solution; once a component is place on the schematic there is practically no way to know which was the bitmap associated with it. This makes sense, especially for the *MultiViewComponents*, since changing their properties can change the component graphic (e.g. like for the *Resistor* and *Resistor_US*. Also the component help text should not be related to the component symbol but to the underlying component model; e.g. for PNP and NPN BJTs we have different symbols bust just one model so IMHO it makes sense to have a single help page for both.

I saw that we now have components which are in more than one Category, like the *Current Probe* and *Voltage Probe*, which are in both *Lumped Components* and *Probes*. That's not an issue for the Qucs GUI but it complicates handling the components cross references in the help text for these components.

in3otd:
This should now be ready; works together with Qucs/qucs-manual#1.

PR #452, opened 2016-02-06
vams: remove non GPL headers

guitorri:
The constants.vams has no copyright notice.
The disciplines.vams is is copyrighted to Accellera.
These files are either provided (on demand) by admsXml or the user has to download them.


felix-salfelder:
oops. i wasn't aware that they are here as well `:/`.

guitorri:
`make dist` is failing in search of these files... I will push a fix.

Issue #453, opened 2016-02-06
Crash when changing background color of text document


in3otd:
When changing a text documents background color (using the Application Settings dialog) Qucs crashes.
To reproduce, open Qucs, create a new text document (File->New Text) and change the Application background color.

The problem comes from [here] (https://github.com/Qucs/qucs/blob/master/qucs/qucs/dialogs/qucssettingsdialog.cpp#L483), as now `TextDoc` is derived from `QPlainTextEdit`.

I'll put a fix together with the other stuff for #446 but feel free to push a separate hotfix if you would like to be sure that it will be in 0.0.19 :grin:

EDIT:
a `grep -r "inherits" --include \*.cpp .` shows also some other suspect things:
```
./qucs/dialogs/simmessage.cpp:  if(DocWidget->inherits("QTextEdit"))
./qucs/dialogs/simmessage.cpp:  if(!DocWidget->inherits("QTextEdit")) {
./qucs/dialogs/simmessage.cpp:  if(DocWidget->inherits("QTextEdit")) {
```

guitorri:
Confirmed on OSX. If you find a fix please inform.

in3otd:
...well, fix is just to use `inherits("QPlainTextEdit")` instead of `inherits("QTextEdit")`, as said it will be in the PR for #446, not sure I can finish it today.

Could you/someone please take a look if there is a similar bug related to the `inherits("QTextEdit")` in `simmessage.cpp` ? At a first glance it's about simulating VHDL/Verilog text files in Qucs, but I never did this, so I cannot check if it's a real issue.

BTW, there is a function [`QucsApp::isTextDocument`] (https://github.com/Qucs/qucs/blob/master/qucs/qucs/qucs.cpp#L2480-L2482) which is extensively used in `qucs.cpp` for this check, will be nice to reuse it around, so if we change the `TextDoc` inheritance another time just changing this function will be enough.
What will be the best way to do this? Declare it static? Or outside of a class, in `misc.cpp` ?

Issue #446, opened 2016-02-05
Change in Qucs Home requires restart.


guitorri:
When the user changes the "File > Application Settings > Locations > Qucs Home" a restart is required.
Can't we just close all the open projects/tabs and reload the list of projects from the new location?

guitorri:
Related: Pressing the '...' button to find the new Qucs Home launches a dialog that does not look native on Windows.

![image](https://cloud.githubusercontent.com/assets/1388949/12855759/8168301e-cc41-11e5-8cf1-97920f9f27b6.png)

in3otd:
@crobarcro I was looking at the Application Settings dialog and noticed that the "Add Path with SubFolders" button actually adds *just* the subfolders not also the selected directory itself. Is this intended?
Seems a bit confusing to me, since the button just above, "Add Path", adds the selected directory (as expected)

crobarcro:
Yes, this is a bug, it should also add the directory

PR #456, opened 2016-02-07
Fix for #446 and #453

in3otd:
lot of small stuff here,
- implemented changing the Qucs Home without restarting: it's a bit hackish but seems to work. Should have divided the "close all open files" and "prepare a new empty schematic and reload the projects tree" into two separate functions and maybe added a `QMessageBox` to inform the user.
- changed the non-native file dialogs (there were several). 
- corrected handling the "Task" and "Comment" colors which has never been properly saved/loaded. They seem actually unused, though :confused: 
- fixed the crash when changing background color of a Text Document, #453 
- corrected the behaviour of "Add Path With Subfolders", see #446 
- removed the remaining `inherits()` used to determine if a widget was a `TextDoc` or a `Schematic`, used a static function defined in `QucsApp` instead. Confirmed that also the simulation of a text document was crashing due to the wrong check.
- removed the deprecated Qt3 compatibility functions in  in `qucssettingsdialog.cpp`; could have been done a bit more cleanly but lacked time...

Tried to check everything, seems to work here, please double check there also, thanks.

### DIFF
--- a/qucs/qucs/dialogs/qucssettingsdialog.cpp
+++ b/qucs/qucs/dialogs/qucssettingsdialog.cpp
@@ -526,7 +526,7 @@ void QucsSettingsDialog::slotApply()
 
         while((w=App->DocumentTab->widget(No++)) != 0) {
           QWidget *vp;
-          if(w->inherits("QPlainTextEdit")) {
+          if(QucsApp::isTextDocument(w)) {
             vp = ((TextDoc*)w)->viewport();
           } else {
             vp = ((Schematic*)w)->viewport();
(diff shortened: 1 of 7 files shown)


in3otd:
? AppVeyor fails with
```
./.libs/lt-admsXml.c:315:19: error: '_P_WAIT' undeclared (first use in this function)
rval = _spawnv (_P_WAIT, lt_argv_zero, (const char * const *) newargz);
^
```
also #458 .

guitorri:
Related to unused QSettings, this is now #460

guitorri:
Thank you @in3otd.
The background color now works for the schematic *and* text documents. Bug or feature?
Perhaps we should allow two colors? Or just white for the text document...

Ignore AppVeyor for the moment.... they updated something that broke something else I don't yet understand...

in3otd:
> The background color now works for the schematic and text documents. Bug or feature?

well, since the original code tried to set the text documents background color to `QucsSettings.BGColor` (but didn't work), I made it actually work as intended. I think we may call it a feature :grin:

But we can certainly have two different background colors for the two document types, in 0.0.20 or later...

guitorri:
Fair enough. 
I build it locally on windows. We can skip the AppVeyor issue for the moment. 
Thank you!

PR #461, opened 2016-02-09
appveyor: set compiler path

guitorri:
Related to update appveyor/ci#637.


guitorri:
Not quite. An error is showing up after, related to Qt. 
I need to figure out what they did with the environment... it seems to be mixing up tools from different compilers on the PATH.

PR #469, opened 2016-02-12
try to make check in appveyor

felix-salfelder:
want to see what happens...

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -59,7 +59,7 @@ build_script:
 
   ## Prepare simple redistributable package (no qucs-doc)
   # Install to a destination
-  - bash -lc "exec 0</dev/null && cd $APPVEYOR_BUILD_FOLDER/qucs && make check || :"
+  - bash -lc "exec 0</dev/null && cd $APPVEYOR_BUILD_FOLDER/qucs && make check"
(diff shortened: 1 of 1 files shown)


in3otd:
don't know what happened, but AppVeyor seemed to pass but didn't actually include the last commit (see [here] (https://ci.appveyor.com/project/qucs/qucs/build/1.0.674)
I have restarted it and now the added `make check` is there, but it fails...

felix-salfelder:
strange. thanks for noticing at all...

felix-salfelder:
anyway, this is failing. should we consider this an issue for 0.0.19?

(if so: can somebody please fix it? through appveyor, this is no fun and takes ages.)

in3otd:
fail message is
```
make[3]: *** No rule to make target 'libqucsUnitTest.exe', needed by 'libqucsUnitTest.log'.  Stop.
make[3]: Leaving directory '/c/projects/qucs/qucs-core/tests'
Makefile:939: recipe for target 'check-TESTS' failed
```
which I think comes from [here] (https://github.com/Qucs/qucs/blob/master/qucs-core/tests/Makefile.am#L51-L55). But even if that was working, I think it's doing actually nothing if [gtest] (https://github.com/google/googletest) is not installed, so it will need additional stuff in `.appveyor.yml`
I think we can leave this out for 0.0.19, what do you think @guitorri ?

felix-salfelder:
looks irrelevant now

PR #470, opened 2016-02-13
fix #447 - Extract project name from path in OS-independent way

### LINKED ISSUE #447: Contents tab, project name takes a lot of space.
See below. On OSX it is not like that. Only the project name is shown, no path.

image

in3otd:
Fixes #447 

Now uses a `QDir` method to extract the project directory name; works on Linux and Windows here.

### DIFF
--- a/qucs/qucs/projectView.cpp
+++ b/qucs/qucs/projectView.cpp
@@ -51,21 +51,12 @@ ProjectView::~ProjectView()
 void
 ProjectView::setProjPath(const QString &path)
 {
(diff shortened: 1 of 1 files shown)


felix-salfelder:
seems to make a lot of sense (i only read the diff). should we include this into 0.0.19?

in3otd:
yes, I think it should go there, so the Windows users will see the project name nicely formatted in the dock as the Linux/OS X users do :grin:

PR #471, opened 2016-02-13
Tuner for Qucs rebased on master

tipofthesowrd:
Thank to Claudio my tuner module is now rebased on the current master.
Remaining 'major' issue is the misbehaviour of a user deletes a component from the schematic which was mentioned as 'tuneable'


in3otd:
the build on Travis on OS X failed with an [error] (https://travis-ci.org/Qucs/qucs/jobs/109011331#L4840-L4844). I guess you might need to `#include <cmath>` in `tuner.cpp`. Why this worked on Linux and WIndows without that I do not know :confused:

tipofthesowrd:
Done, added <cmath> and checks pass
@in3otd Probably under Linux or Windows cmath is included by some other means and not on Mac.
Doesn't matter, small fix

andresmmera:
I think Travis complains because I didn't add "tune.png RemoveElementTuning.png" to qucs/qucs/bitmaps/CMakelists.txt, line 33. Could you (or someone else) add this?

@guitorri Are you planning to add this feature in the next release (0.0.20)? ...some more revision/testing may be needed. Screencast: https://drive.google.com/open?id=0BwDjrwGfd_z7VlplN1kzMWRSNFk

in3otd:
I think those two `.png` files are missing in `qucs/qucs/bitmaps/Makefile.am` actually and the issue shows up only in the Linux builds since only there we do a `make distcheck` (i.e. with an out-of-tree build), IIRC. I'm taking a look at this.
We do not have (yet) any CI using `cmake`.

guitorri:
I haven't tried this yet (will do now). It is looking pretty good. Thank you for the screencast.

guitorri:
A few cosmetic things:
 - perhaps split the layout in two groupboxes 
- there should be a way to scroll left and right if the user has too many parameters to tune
 - the [license header](https://github.com/Qucs/qucs/wiki/Development#license-header) could be updated
- the buttons should have a min size, if the left edge of the dialog is dragged the buttons get bigger instead of making more room for the parameters

Frankly, if nobody can do it now, I would merge as such and we can improve it later.

I really want to make a release branch this weekend. You guys let me know if want to merge this.

tipofthesowrd:
Ah, I was wondering why the automated builds were failing since it was building fine locally
Maintainers are now able to edit + the image has been merged into the pull request

in3otd:
some random notes:
- didn't tuning from the Data Display work some time ago? When tuning a Schematic the first change correctly simulates and opens a Data Display but then it just opens the Simulation Messages window, telling "checker error, no actions defined: nothing to do".
- a property having a negative number value cannot be tuned, I get a "You selected a non-tunable property" message. And, by the way, I think that `checkProperty()` should be a (static) member of `tunerDialog()` (similarly `getScale()` and others)
- the strings [here](https://github.com/tipofthesowrd/qucs/blob/tuner_rebased/qucs/qucs/qucs_init.cpp#L577-L584) should be arguments to `tr()` to be translatable (and the comments may not be needed anymore :grin:)
- the known bug that causes Qucs to crash if a component being tuned is removed from the schematic is still there... may not be a common case but I don't like to have the application crashing

guitorri:
Fair enough. Work is needed. I will move it forward. I don't mind releasing again as soon as this is ready to be merged. We must speed up the release cycles anyway.

tipofthesowrd:
Dagnabbit, too bad for this release, although I agree.

- You're right, that was working
andresmmera did some awesome work on the tuner and it might have gotten lost along the way
- should be an easy fix
- those still there 😆 
- couldn't find a workaround for that one, help or hints are appreciated

in3otd:
yep, I agree this has been waiting for far too long to be merged...
Honestly, if it was not for the possibility of crashing Qucs it could be merged, IMO - as everything, it could be further polished elsewhere but nothing really urgent.

I also tried to see if there was a workaround for the "component removed" thing, but didn't see any...
A similar issue exist when removing an Equation variable whose value is being tuned; in this case the tuner silently starts to change the following equation instead.
Maybe we could simply make the schematic non-editable during tuning? (don't know how this could be done actually)

andresmmera:
> a property having a negative number value cannot be tuned, I get a "You selected a non-tunable property" message. And, by the way, I think that checkProperty() should be a (static) member of tunerDialog() (similarly getScale() and others)

I think the proper way to decide if a property is tunable or not would be to add an extra (boolean) field in every parameter of every component.

> I also tried to see if there was a workaround for the "component removed" thing, but didn't see any...

Wild guess,... would it be possible to throw a signal when the component is removed? That way the tuner could catch that and prevent Qucs from crashing.

tipofthesowrd:
> I also tried to see if there was a workaround for the "component removed" thing, but didn't see any...

I was working on the same thing today via a signal.
Currently the schematic.cpp can throw a signal when a component is removed but only to QucsApp?

`connect(this, SIGNAL(signalCompChange(Component*)),
    App_, SLOT(slotCompChanged(Component*)));`

Trying to send the signal directly to the tuner results a segmentation fault

`connect(this, SIGNAL(signalCompChange(Component*)),
    (const QObject*)App_, SLOT(slotCompChanged(Component*)));`

Still have to properly setup my debugger on my new ubuntu installation before I can find out what's goign on

guitorri:
Besides the current implementation, what should be the criteria for a tunable parameter in general? 
At first sight anything that can be represented as a slider (a range) should be a tunable parameter, no?

Yes, somehow, if the component is deleted, it triggers the corresponding tuning slider to be removed.

felix-salfelder:
> what should be the criteria for a tunable parameter in general?

perhaps it should be a parameter to begin with?

this pr looks really scary to me. we totally need a place for code like this.

andresmmera:
> what should be the criteria for a tunable parameter in general?

Well, you cannot tune port numbers, string parameters (such as combos or variable names), complex valued parameters (like Pxy in the RF-EDD component)... You cannot tune the 'coils' field in the coupled inductors component since it implies modifying other fields.
Also, I think there are some stuff it makes no sense to tune like simulation parameters.

I don't think there's a general rule for that... but many special cases

andresmmera:
> we totally need a place for code like this.

@felix-salfelder what do you mean? Should be the tuner kinda tool outside the Qucs code? I mean, like the other tool (filter design, transcalc, etc...)

felix-salfelder:
@andresmmera i don't know what the best solution is. if this is a tool like the others, it should be done like those others. maybe the interface needs an improvement, but then it should be improved first. adding extension code in random places (such as to the main executable) is a no-go wrt maintainability or future development.

what is missing in the tools interface?

tipofthesowrd:
@felix-salfelder fyi, I just compared with Microwave Office. When a tuned component is removed from the schematic it is removed from the tuner as well with the tuner dialog open.

In the end we need a way for the Qucs Schematic to signal a component has been removed.

From a feature perspective I would like to say this is on of the most needed features for Qucs (from an electronics / microwave point of view). And that is not just my personal preference but the most heard request from collegues / other microwave enthousiasts.

andresmmera:
Yes, this tool is not like the others in the sense it needs dynamic interaction with the main application. I think that @felix-salfelder means that it would be better if we could find a way to make the feature less intrusive in QucsApp (honestly, I don't see how). Maybe we could figure out a more elegant solution based on some kind of interprocess communication, but in the end I guess it works like the signal-slot mechanism

> In the end we need a way for the Qucs Schematic to signal a component has been removed.

Yes, and update the diagrams and the component values in real time.


> From a feature perspective I would like to say this is on of the most needed features for Qucs (from an electronics / microwave point of view). And that is not just my personal preference but the most heard request from collegues / other microwave enthousiasts.

I couldn't agree more! That's why I've tried to push it forward during summer :-)

felix-salfelder:
i didn't say it is not a nice feature.

> Maybe we could figure out a more elegant solution

start with moving the code to an appropriate place. there shouldn't be any application specific hacks in the core.

in3otd:
> what is missing in the tools interface?

the problem is that currently the Tools do not really have an interface, all of them, except the Matching Tools, are simply started from the GUI but do not get any input from it. OTOH, to be able to provide inputs to the Matching Tools, this is actually not a separate application but embedded in the main code. The tools that provide an output do this by simply loading a new schematic in the clipboard, to be manually pasted in the GUI by the user later. With the current (non-existent) interface, I don't think we can implement the Tuner in a less intrusive way.
We certainly need to better integrate the Tools, the fact that they are mostly completely independent applications limits their usefulness.

Also the interface with the simulator should likely be improved, currently also when Optimizing or Tuning a circuit (so not changing the circuit topology but just some components values) we can only pass a new netlist to simulate but there is no way to modify a component parameter and re-run a simulation without going thru the GUI schematic, so , every time a netlist must be generated, parsed by the simulator, the various internal data structures re-created, etc. Ideally we should be able to keep the simulator "open" and just send some kind of `.alter` command to modify the simulator internal structures, since the circuit topology is not changed.

ok, I went a bit off-topic...

felix-salfelder:
> [withhout interface] I don't think we can implement the Tuner in a less intrusive way.

if we merge it like this, it will be (much) harder to go forward. so we probably need some kind of interface?

> ok, I went a bit off-topic...

not really. let's sort what is needed to do this kind of feature. i would like to have interactive simulation in, maybe there is an overlap?

tipofthesowrd:
'[withhout interface] I don't think we can implement the Tuner in a less intrusive way.'

Correct, I found no other way to currently interface to either the schematic, netlist other than adding code directly. 

As for the launching of the tuner dialog (and all other pieces of code added to the core) I tried to write as unobtrusively as possible by copying existing code to launch the other tools.

I agree it is not as nice as possible, but currently Qucs doesn't support another way.

As for moving forward I leave the choice open. Either go for a full fledged interface (does this exist in Qt, comparable to a c# interface?), and push back this feature several releases.

Or attempt to improve the current implementation further and merge this sooner rather then later.

felix-salfelder:
> I found no other way.

i tend to believe you, can you explain the problem? can you maybe throw in what you need? perhaps there will be some work on the simulator driver interface one of these days, and tailoring is still possible.

generally speaking, adding features in the proposed unfinished way is mostly a maintenance problem. people drive by and drop some contribution the project is not up to speed with. we have plenty of those already... guess why plugins are not there yet.

in contrast, keeping unfinished branches around is not a big deal, and that's what branches are for.

andresmmera:
As said, IMHO, the problem is that the tuner needs to hijack the Qucs application for updating the schematic, calling qucsator and finally update results.

I think that we can generate the netlist and call qucsator from the tuner code, but, the thing is: can we modify the schematic/display sheet *dynamically* without intruding the main app code?

felix-salfelder:
> we modify the schematic/display sheet dynamically

if we (really) want that, there should be an interface to it.

but i can't see why we should go past changing parameters in the schematic. perhaps we should introduce an interface to that first...

in3otd:
(about tuning from the Data Display)
> You're right, that was working
andresmmera did some awesome work on the tuner and it might have gotten lost along the way

ops, sorry, tuning from the Data Display still works as it did some time ago; I was testing with a schematic which points to a Data Display which in turn has another schematic as default source (still called "Data Display" in the document settings dialog, we should maybe fox that too). So no bugs were introduced by anyone here :grin:. Still, would be nice to handle this case too, I'll take a look.

in3otd:
@tipofthesowrd I've noticed that now the Simulation Messages window is not closed automatically, when doing a normal simulation, which is a bit annoying. I saw the [comment](https://github.com/Qucs/qucs/pull/471/files#diff-07a46d9c7b31a19eccbb820ca0cd3aa8R2177) in your code, why closing the window as usual should be avoided?

tipofthesowrd:
@in3otd

When pulling a slider you can be running many simulations based on the step size, this causes the simulation window to open and close constantly and taking focus each time you tune or press a step button

in3otd:
ok, but do you think it's safe to close the Simulation Message window if not in `TuningMode` or the fact that it's now global can causes other issues?

tipofthesowrd:
@in3otd correct, when not in tuning mode there is no need to change this behaviour.
@felix-salfelder what the tuner would need as interfaces is pretty well described by @andresmmera 

From what I remember to describe the interfaces in more detail:

- [ ]  ability to open a dialog from the main window and overwrite default mouseactions
- [ ]  access to the schematic components 
- Detect when a user clicks a component + property via mouseactions
- Update value of a property 
- Detect when a component is removed

- [ ]  access to simulation engine
- Run a (many) simulation(s) with updated netlists each time (currently this works since we update the schematic)
- [ ] access to data display
- Run simulations from the datadisplay & update results

in3otd:
I've modified the code so that the Tuner works also when the schematic opens a Data Display after the simulation and the Data Display is not associated with that schematic.

in3otd:
@tipofthesowrd, did you have time to look further at using signals to handle components removal in the Tuner?

in3otd:
I've rebased this again on the latest `develop` branch and handled the component removal case by using a signal, as mentioned above; that seems to work.

**However**, while testing I've noticed another, fundamental, issue with the schematic *Undo*:
- the Tuner keeps pointer to the components properties being tuned
- when you modify something and then do an *Undo* the schematic is practically [rebuilt](https://github.com/Qucs/qucs/blob/develop/qucs/qucs/schematic_file.cpp#L1000), with new components instances, so the pointers kept by the Tuner are now invalid.
- so if one uses the *Undo* while tuning, Qucs will crash.

not sure how this should be solved.

felix-salfelder:
components properties are implementation details. these must be private, the current property dict must be removed anyway, and has been, in some refactored code.

why not (in tuner)

auto* C = schematic.find_stuff("component_that_needs_change");
C.set_parameter("some_parameter", "some_other_value");

?

in3otd:
yep, the current way of keeping direct pointers to component properties looks quite fragile.
I was wondering what could be used as "unique component ID" (the "component_that_needs_change" part), the component `Name`, something including the schematic name also (`schematicName:ComponentName`) or other.

Do you have a pointer to the refactored code mentioned above?

felix-salfelder:
not sure what you mean by "schematic name". is that the file that contains a subcircuit definition? if so, then, no. you need the reference to the *instance* of that subcircuit.

    .subckt foo 1 2
    V1 1 2 X
    ends

    foo1 1 2 foo
    foo2 1 2 foo

that gives `foo1.V1` and `foo2.V1`. in gnucap, `foo.V1` is the "long_label"....
 
thinking again: is this a good idea? perhaps it is more adequate to just change parameters.

    V1 1 2 {P}
    .param P=5
    .do_stuff
    .param P=10
    .do_it_again

We will end up with something structually more like 'V1 1 2 {P+3}' anyway, or never tune real circuits.

i don't know where the particular parameter cleanup is, maybe in the verilog netlister PR?. i started to rebase the refactoring for easier review. #757 is next.

mcknly:
What is the status of this feature? I can help test if necessary. I have a use for this functionality in my work, so I should be able to vet it pretty well.

in3otd:
it should work pretty well, as long as you do not modify the circuit (add/remove components, use Undo/Redo, etc.) during tuning. In the short term we should just set the schematic read-only during tuning, in the long-(ish) term a cleaner solution as discussed above will be better.
So, please help testing and report back how it works there, thanks.

mcknly:
Very good... I have it built and it seems the basic functionality is working so far. Will continue testing.

Issue #467, opened 2016-02-12
Selecting wire label change font on diagram


guitorri:
When selecting a wire label (I believe it was the last label I placed in the schematic). The fonts on diagrams become boldfaced.

Select first label (as expected):
![image](https://cloud.githubusercontent.com/assets/1388949/13007992/089dbafc-d194-11e5-8d91-c6c395cc1400.png)

Select second label (fonts on table and other diagrams become bold):
![image](https://cloud.githubusercontent.com/assets/1388949/13007997/171878ba-d194-11e5-9629-e651744ff842.png)

in3otd:
this seems to come from [here] (https://github.com/Qucs/qucs/blob/master/qucs/qucs/wirelabel.cpp#L116); indeed saving the font before and restoring after, as done [for the other Elements] (https://github.com/Qucs/qucs/blob/master/qucs/qucs/components/component.cpp#L189-L283) seems to fix the issue. Probably the issue is there since 312d8cc9234cca9cf58e665b0b529bcca68c47f0 but did not have time to bisect.
Will check a bit more and do a PR later.

in3otd:
oh, just found another (small) bug related to the wire labels highlighting: label several wires with the label "A", several others with "B". If you select just one of the labels name (A or B) you will correctly see all the other labels with the same name highlighted (bold), If you now select also the other label (i.e. both A and B, with Ctrl-LMB) only the last label name clicked will be highlighted, the first one will remain selected but not highlighted.Looking at the code this is not the intended behavior, both labels names should be highlighted.

Besides, I do not remember having seen this nice highlighting stuff documented anywhere (true for many other things...)

in3otd:
another related bug: label a wire and also a node (not connected to the wire, of course). Click on a label and you'll see that the other one will *not* be highlighted. Yet the wire and the node will be connected in the netlist (as expected).

guitorri:
Fixed in release-0.0.19 (356a7fc6b4d231ff6a719dde72728b42a3c80a73)

PR #472, opened 2016-02-13
Fix for #467

in3otd:
#~~(Partial)~~ fix for the issues in #467 

Things corrected so far:
- [x] when selecting a wire label the text on some diagrams become boldfaced
- [x] selecting multiple different labels only highlights one group
- [x] labels on nodes are never highlighted
- [x] highlighting not updated when label is edited

I have also slightly modified the highlighting behavior: now a label is also highlighted (boldfaced) when selected only if the same label is used elsewhere. I think this gives a better feedback if a label is used multiple times, without need to search for boldface labels in the entire schematic to understand if a label is unique or not.

~~Feel free to merge already if you would like to see these things already fixed in 0.0.19 .~~
EDIT: should be ready to merge now

### DIFF
--- a/qucs/qucs/wirelabel.cpp
+++ b/qucs/qucs/wirelabel.cpp
@@ -115,7 +115,7 @@ void WireLabel::paint(ViewPainter *p)
 //    highlightfill.setAlpha(50);
 //    p->fillRect(x1-1, y1-1, x2, y2, highlightfill);
     p->Painter->setPen(QPen(Qt::darkBlue,3));
-    newFont.setWeight (QFont::DemiBold);
+    newFont.setWeight (QFont::Bold);
   }
   else
   {
--- a/qucs/qucs/wirelabel.cpp
+++ b/qucs/qucs/wirelabel.cpp
@@ -106,22 +106,25 @@ bool WireLabel::getSelected(int x, int y)
 // ----------------------------------------------------------------
 void WireLabel::paint(ViewPainter *p)
 {
-  QFont font = p->Painter->font();
+  QFont f = p->Painter->font(); // save current font
+  QFont newFont = f;
+
   if (isHighlighted)
   {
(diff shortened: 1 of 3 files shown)


in3otd:
done also the node labels stuff.
If you would like to play with this open [this schematic](https://github.com/Qucs/qucs/files/129379/test_label_hilight.txt) (rename it in `.sch`)

qucs/qucs/wirelabel.cpp
```diff
@@ -106,22 +106,25 @@ bool WireLabel::getSelected(int x, int y)
...
+  QFont newFont = f;
+
   if (isHighlighted)
   {
 //    QColor highlightfill (Qt::blue);
 //    highlightfill.setAlpha(50);
 //    p->fillRect(x1-1, y1-1, x2, y2, highlightfill);
     p->Painter->setPen(QPen(Qt::darkBlue,3));
-    font.setWeight (QFont::DemiBold);
+    newFont.setWeight (QFont::DemiBold);
```
guitorri: Can we change for QFont::Bold?
On my Mac the font change is not visible. Probably related to http://stackoverflow.com/questions/14901933/is-it-possible-to-specify-more-than-just-regular-and-bold-versions-of-a-font

in3otd: ...I guess yes; let me check how it looks on Linux and Windows...

in3otd:
no difference between Bold and DemiBold here, look like Bold anyway.

bolddemibold

I have also corrected another small issue: when a label was edited the "highlighting" was not updated but the previous state was kept, until the label was clicked again.

guitorri:
After you patch on Mac OSX (without it the font only becomes blue, but not **Bold**):

image
image

guitorri:
Merged into release-0.0.19 (356a7fc6b4d231ff6a719dde72728b42a3c80a73)

Issue #476, opened 2016-02-15
Finalize release 0.0.19, todo list


guitorri:
- [x] update NEWS.md
  - [x] doc OCTAVEBINDIR -> QUCS_OCTAVE 
- [x] ~~provide Windows binary package for freeHDL (other dependencies have packages)~~
- [x] ~~test optional packages~~ 
- [x] check issue related to verilog-a loader reported by Mike (#478)
- [x] update strings/ push to transifex
- [x] create release candidate
- [x] tag to https://github.com/Qucs/qucs-help
- [x]  update Windows packaging about OCTAVEBINDIR -> QUCS_OCTAVE 
- [x] include freeHDL (one last time) into Windows binary package.

---

To do:
- [x] Merge `release-0.0.19` into `master` and `develop`. #621 #642 
- [x] Tag `qucs-0.0.19` into `master`. 
- [x] Move the base of all PRs to `develop`
- [x] Delete `release-0.0.19`
- [x] Sync master branch on SF repository one last time...
- [x] Produce source and binary packages, upload o SF
- [x] update website, landing and download pages
- [x] Announce qucs-devel, qucs-help, update SF news
- [x] update Wikipedia

ra3xdh:
> check issue related to verilog-a loader reported by Mike

@guitorri  , There was no blocking issue with Verilog-A. There was capacitor polarity issue in my Verilog-A synthesizer that will be not included in release. I will correct code in my branch soon. The `master` branch is not affected. You can mark this point as completed and proceed.

guitorri:
Hi @ra3xdh, I was testing Mike model against master. It was crashing on me while loading the .png (or dragging the component into the schematic). It was late, I will check again to see what I did.

guitorri:
@ra3xdh see #478 for the issue I saw.

felix-salfelder:
> provide Windows binary package for freeHDL (other dependencies have packages)

are you sure? i mean, is freeHDL unmaintained?

guitorri:
freeHDL is dead upstream.

guitorri:
by the way, we should deprecate freeHDL and switch to GDHL...

felix-salfelder:
> freeHDL is dead upstream.

do we really need it for the release? i mean, there's a lot of things to wait for, for example debian has no adms package yet...

> and switch to GDHL...

probably makes sense.

> [further up] test optional packages

which optional packages?

guitorri:
[...]

PR #480, opened 2016-02-18
more qucs/NEWS.md

felix-salfelder:
trying to help with `qucs/NEWS.md` (#476)

this file mentions `qucsator`. however, there is no qucsator in `qucs`... i was tempted to move it to toplevel, but then we would have no NEWS.md for qucs anymore `:(`.

### DIFF
--- a/qucs/NEWS.md
+++ b/qucs/NEWS.md
@@ -11,6 +11,8 @@ Port
 
  * Ported Component member variables Lines, Arcs, Rects, Ellips, Texts from Qt3PtrList to QList
 
+ * Now using Qt to draw plots
+
 
 Miscelaneous
 ------------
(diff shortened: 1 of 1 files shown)


guitorri:
Thank you. I have more things lined up here: https://gist.github.com/guitorri/5ce296f510980ec64a45

I am tempted to rename it to Changelog.md, after [this](http://keepachangelog.com/)

I made the "mistake" of unifying all the changes for the package in one NEWS.md on 0.0.18 (before `make dist` we used to flatten all dirs into qucs/qucs/ so it would become the toplevel NEWS. Each subdir has a NEWS, but we are not maintaining them. I don't know.

felix-salfelder:
gnu standard is to have both (which might be a good idea).

http://www.gnu.org/prep/standards/html_node/Change-Logs.html#Change-Logs
http://www.gnu.org/prep/standards/html_node/NEWS-File.html#NEWS-File

> I am tempted to rename it to Changelog.md, after this

maybe except for the 1st section, which reads much like NEWS. (YMMV)

> Each subdir has a NEWS, but we are not maintaining them. I don't know.

does not seem to make much of a difference. currently everything is (still?) in one repo, so one set of news/changelog is reasonable.

guitorri:
Merged/modified locally. I will push soon the updated NEWS.md.

PR #481, opened 2016-02-18
Correct S-parameters renormalization formula

in3otd:
There was a post to the `qucs-help` maling list some time ago about some "strange" results using the [S-parameters renormalization formula found in the technical documentation](http://qucs.sourceforge.net/tech/node98.html#SECTION001611000000000000000).

Indeed a term there was missing and unfortunately it was also missing in the [actual implementation] (https://github.com/Qucs/qucs/blob/989c77243bfd1ee0e78fb9c29384a5b71843f91e/qucs-core/src/math/matrix.cpp#L855-L900) in `qucsator`. There was also a a sign error (in the denominator of r<sub>i</sub>) in the [source code documentation] (
http://qucs.github.io/qucs-doxygen/qucs-core/namespacequcs.html#a398c66cb517cc637f7df2bd9085dda93).

To verify this, you can open [this schematic](https://github.com/Qucs/qucs/files/136841/stos_bug.zip) to check the results (had to zip it to have it accepted by GitHub)

And since it's a bug, I propose to squeeze this fix into 0.0.19...

### DIFF
--- a/qucs-core/src/math/matrix.cpp
+++ b/qucs-core/src/math/matrix.cpp
@@ -51,7 +51,8 @@
       Impedance Environment"
       in Scropion Application note
       Anritsu
-      online(2007/07/30) http://www.eu.anritsu.com/files/11410-00284B.pdf
+      online (2016-02-21) 
+      https://dl.cdn-anritsu.com/en-us/test-measurement/files/Application-Notes/Application-Note/11410-00284B.pdf
 
   [5] Conversions between S, Z, Y, H, ABCD, and T parameters
       which are valid for complex source and load impedances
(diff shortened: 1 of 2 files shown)


in3otd:
yes, `a` and `r` are diagonal; `z0` and `zref` there are both `qucs::vector` and `sqrt (z0 / zref)` is supposed give the vector where the operations are applied element-wise (`sqrt (z0`<sub>i</sub>` / zref`<sub>i</sub>)).
I didn't check what `inverse()` does, indeed makes little sense to invert a diagonal matrix as a dense one... Should we have a specific "diagonal_matrix" class ? :)
Or in the meantime, just do the diagonal matrix inversion locally?

felix-salfelder:
thanks. just figured out, that my grep for `operator/` missed vector.h because it's in the wrong place `:/` (see my EDIT).

about inverting diagonals: how about elementwise invert and then put on the diagonal? (still inefficient, but better). much better, but less readable would be this.

    matrix& multiply_columns(matrix&, const vector);
    matrix& multiply_rows(matrix&, const vector);

different matrix types for different storage makes a lot of sense, but we should use a library for that, not reinvent the wheel. (postpone.)

felix-salfelder:
the ` * 2 * zref ` change in the code seems correct. i have checked the case where `0<=r<=1` and `z0+zref` is positive (which is probably intended).

in3otd:
ok, will invert element-wise instead of using `invert()`.

Should we move also `qucs-core/src/matvec.{h,cpp}` and `qucs-core/src/fourier.{h,cpp}` to `qucs-core/src/math/`? In this PR or another targeting `develop`?

felix-salfelder:
please only include the changes that you feel comforatble with *now*. my extra suggestions are due to my "while i am at it"-attitude ...

in3otd:
ok, I have updated the S-parameters conversion stuff

PR #482, opened 2016-02-20
Match dialog fixes

in3otd:
Small fixes for the Tools->Matching Circuit dialog:
- wrong results show for the polar/rectangular conversion:
  - to reproduce: open dialog, uncheck "calculate two-port matching", enter a non-zero imaginary part for the "Reflexion coefficient", change the Input Format from *real/imag* to *mag/deg* and then back again to *real/imag*. The result shown is not the same as the initial one.
- wrong conversion when polar format is used (similar to the previous one, but different root cause):
  - to reproduce: open dialog, uncheck "calculate two-port matching", change the Input Format to *mag/deg* and enter a non-zero angle part for the "Reflexion coefficient", e.g.10 degrees. The impedance will change even if the reflection coefficient magnitude is fixed at zero.

plus other small "cosmetic" fixes, see commits messages for details.

### DIFF
--- a/qucs/qucs/dialogs/matchdialog.cpp
+++ b/qucs/qucs/dialogs/matchdialog.cpp
@@ -97,8 +97,8 @@ MatchDialog::MatchDialog(QWidget *parent)
   h1->addWidget(FormatLabel);
   FormatCombo = new QComboBox();
   h1->addWidget(FormatCombo);
-  FormatCombo->insertItem(tr("real/imag"));
-  FormatCombo->insertItem(tr("mag/deg"));
+  FormatCombo->insertItem(0, tr("real/imag"));
+  FormatCombo->insertItem(1, tr("mag/deg"));
   connect(FormatCombo, SIGNAL(activated(int)), SLOT(slotChangeMode(int)));
   h1->addStretch(5);
   SParLayout->addLayout(h1);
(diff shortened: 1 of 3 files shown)


qucs/qucs/dialogs/matchdialog.cpp
```diff
@@ -238,40 +252,51 @@ void MatchDialog::slotSetTwoPort(bool on)
...
-    S12sLabel->setEnabled(true);
-    S22sLabel->setEnabled(true);
-    S12degEdit->setEnabled(true);
-    S22degEdit->setEnabled(true);
-    S12uLabel->setEnabled(true);
-    S22uLabel->setEnabled(true);
-    Port2Label->setEnabled(true);
-    Ref2Edit->setEnabled(true);
-    Ohm2Label->setEnabled(true);
+    S12magEdit->setVisible(true);
```
felix-salfelder: can you put these in a seperate function?

    void something(bool x)
    {
      S12magEdit->setVisible(x);
      // the other 13 lines
    }

better yet: use a container

    for(x in list){ x->setEnabled(true); }

(just a thought)

in3otd:
ok, put that stuff in a separate function.

I have push-forced since I removed two deprecated Qt3-support function that I previously forgot - now added to the related commit.

in3otd:
didn't see a neat way to use a loop, so I just did 4 separate functions...

While I was at it, I have also added some little "setters" to access the internal widgets.

Issue #478, opened 2016-02-16
Reloading verilog-a symbol/bitmap crashes application.


guitorri:
Steps to reproduce:

- Extract and open the attached project (see below).
- `Project > Load Verilog-A module...`, Ok for the selected Verilog-a symbol
- Pick and place from `Components > verilog-a user devices` (in the dock), should work normally
- (repeat) `Project > Load Verilog-A module`...
- Press `Change icon`
- Select another png icon, ok
- Pick and place from `Components > verilog-a user devices` --> CRASH

Traceback:

```
Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   org.qucs                      	0x000000010bcc53a0 QucsApp::slotSelectComponent(QListWidgetItem*) + 3440 (qucs.cpp:884)
1   org.qucs                      	0x000000010bd4f194 QucsApp::qt_static_metacall(QObject*, QMetaObject::Call, int, void**) + 548 (moc_qucs.cxx:286)
2   QtCore                        	0x000000010d7dbb4e QMetaObject::activate(QObject*, QMetaObject const*, int, void**) + 1566
3   QtGui                         	0x000000010cf35e9e QListWidget::itemPressed(QListWidgetItem*) + 46
4   QtCore                        	0x000000010d7dbb4e QMetaObject::activate(QObject*, QMetaObject const*, int, void**) + 1566
5   QtGui                         	0x000000010cec2e56 QAbstractItemView::pressed(QModelIndex const&) + 38
6   QtGui                         	0x000000010cec78b9 QAbstractItemView::mousePressEvent(QMouseEvent*) + 1305
7   QtGui                         	0x000000010ca11940 QWidget::event(QEvent*) + 2816
8   QtGui                         	0x000000010cdaed5c QFrame::event(QEvent*) + 44
```  

Further issues:
- [x] cannot unload symbols

---
Edit: mention the attached project below

guitorri:
This test project, by @MikeBrinson,  can be used to reproduce the issue [RC_prj.zip](https://github.com/Qucs/qucs/files/133353/RC_prj.zip)

ra3xdh:
Yes, I can reproduce this. Time-domain simulation issues reported by Mike are not caused by this bug. I have fixed Verilog-A synthesizer at my branch and it seems time-domain simulation issues are gone.

in3otd:
[...]

PR #491, opened 2016-03-12
fix issue #478: crash on reload verilog symbols

guitorri:
* Use a mutable interator to remove previously registered modules.
  Deleting the items of a non-mutable iterator results in a crash.

* Take care of clearing the icon list when module are unloaded.


qucs-test
```diff
@@ -1 +1 @@
-Subproject commit 4ca2a329eb1966b9216286ebda3e54efc1047371
+Subproject commit 9b4234615f4437d6ef17a26850bb9f00f99523bf
```
guitorri: This should not be here... I will remove this.

Issue #498, opened 2016-03-25
Fast zooming (Ctrl+Mouse wheels) doesn't work.


arhiv6:
In Qucs at fast zooming (using Ctrl + Mouse wheel) is not working properly. If you change the zoom is not fast - it works fine. But if you zoom in quickly, the scale is reset to the initial value - that is it does not work. Tested on Ubuntu15.10 32 and 64bits, version 0.0.18 and 0.0.19.
Video: [https://youtu.be/yOsCOftvIBM](https://youtu.be/yOsCOftvIBM)

in3otd:
I have tried to reproduce the issue (Slackware 32 b, Qucs 0.0.19), but I did not see this problem. Do you see a similar behavior also with panning (just the mouse wheel and Shift+Mouse wheel)?

arhiv6:
No, Mouse wheel and Shift+Mouse wheel working normal.

andresmmera:
I've noticed that behaviour some time ago... but in the end I can live with it so I've never reported it :-)
Here i'm using Ubuntu 14.04 LTS 64b

in3otd:
ok... can you try to insert a line 
```
qDebug() << delta;
```
[here](https://github.com/Qucs/qucs/blob/master/qucs/qucs/schematic.cpp#L1882) and recompile, looking then at the numbers printed on the terminal while zooming, just to check if some strange step value is passed to the routine. I always get just `60` or `-60`.

andresmmera:
Uhm,... I've followed your instructions and after doing some random fast zooming (CTRL+Wheel) I can see many 60 and -60, but also 120, -120 and -180

andresmmera:
I can confirm that it shows (-60, 60) when you scroll slowly. However, as soon as you scroll faster, the effect described above arises and the log shows some 120 and 180.

andresmmera:
I've reduced the zooming step by replacing  "int delta = Event->delta() >> 1; " by   "double delta = Event->delta()*0.7;"  at line 1881 and it seems to work here...

in3otd:
thanks; could you please try changing the lines [here](https://github.com/Qucs/qucs/blob/master/qucs/qucs/schematic.cpp#L1894-L1895) to
```C++
if(delta < 0) Scaling = 1/1.1;
else Scaling = 1.1;
```
and check whether the zoom issue is gone

andresmmera:
Yep, this way the problem disappeared too :-)

in3otd:
[...]

PR #499, opened 2016-03-26
Fix schematic zoom for fast wheel changes

in3otd:
Fix for #498. Some mice return delta values which are multiples of 120 when the wheel is turned fast. This caused the zoom factor to be calculated wrongly; now the zoom factor is scaled properly if a value higher (or lower) than 120 is returned.

The PR is to the `develop` branch, I think it could also be safely applied to the `relase-0.0.19` branch.


andresmmera:
Tried here. It works :+1:

in3otd:
@andresmmera , could you please do a final check of the code (if needed) and merge to `release-0.0.19`, thanks!

PR #501, opened 2016-03-28
Fix lib lang search

guitorri:


guitorri:
@in3otd the first patch should fix the issue you observed on the rc1 win64 .zip package.

There is/was a similar issue with the translations.

in3otd:
Merged into `release-0.0.19` (e1f97da2771b19fbe29d813abc91db53b0fdedd8)

in3otd:
BTW, #377 had also a small fix for an hard-coded path (see https://github.com/Qucs/qucs/pull/377/commits/32c691af73a9acdc945864dcac9f8e59eff0228a), it may make sense to merge that into `0.0.19` also.

PR #504, opened 2016-04-03
fixed 2 bugs in wire editing

alysogorov:
1. cross was painted wrong at scroll/zoom, fixed.
2. wire was painted wrong if corner changed(right button press while
   wire creating).

### DIFF
--- a/qucs/qucs/mouseactions.cpp
+++ b/qucs/qucs/mouseactions.cpp
@@ -311,6 +311,43 @@ void MouseActions::MMoveElement(Schematic *Doc, QMouseEvent *Event)
   Doc->viewport()->update();
 }
 
+/**
+ * @brief draws wire aiming cross on Document view
+ * @param Doc - pointer to Schematics object
+ * @param fx  - document x-coordinate of center
+ * @param fy  - document x-coordinate of center
+ */
+static void paintAim(Schematic *Doc, int fx,int fy){
+	//let we reserve couple of points at the edges of lines for some aesthetics,
+	//and visual check that our calculations has fit the widget window.
+	const int ldelta = 2;
+
+	//left and upper edges of our lines
+	int lx0 = DOC_X_POS(Doc->contentsX()+ldelta);
+	int ly0 = DOC_Y_POS(Doc->contentsY()+ldelta);
+
+	//right and bottom edges
+	int lx1 = DOC_X_POS(Doc->contentsX()+Doc->viewport()->width()-1-ldelta);
+	int ly1 = DOC_Y_POS(Doc->contentsY()+Doc->viewport()->height()-1-ldelta);
+
+	//post line paint events
+	Doc->PostPaintEvent (_Line, lx0, fy, lx1, fy);
+	Doc->PostPaintEvent (_Line, fx, ly0, fx, ly1);
+}
+
+//paint ghost line - horizontal
+static void paintGhostLineH(Schematic *Doc, int fx,int fy, int fxx){
+	Doc->PostPaintEvent (_Line, fx, fy-1, fxx, fy-1);
+	Doc->PostPaintEvent (_Line, fx, fy+1, fxx, fy+1);
+}
+
+//paint ghost line - vertical
+static void paintGhostLineV(Schematic *Doc, int fx, int fy, int fyy){
+	Doc->PostPaintEvent (_Line, fx-1, fy, fx-1, fyy);
+	Doc->PostPaintEvent (_Line, fx+1, fy, fx+1, fyy);
+}
+
+
 // -----------------------------------------------------------
 /**
  * @brief MouseActions::MMoveWire2 Paint wire as it is being drawn with mouse.
@@ -322,14 +359,18 @@ void MouseActions::MMoveWire2(Schematic *Doc, QMouseEvent *Event)
   MAx2  = DOC_X_POS(Event->pos().x());
   MAy2  = DOC_Y_POS(Event->pos().y());
   Doc->setOnGrid(MAx2, MAy2);
+  paintAim(Doc,MAx2,MAy2); //let we paint aim cross
 
+  //because cross slightly masks a wire, let we make wire thicker
+  //better to make it by increasing of pen, but here we cannot access
+  //pen
   if(MAx1 == 0) {
-    Doc->PostPaintEvent (_Line, MAx3, MAy3, MAx3, MAy2); // paint
-    Doc->PostPaintEvent (_Line, MAx3, MAy2, MAx2, MAy2); // paint
+	paintGhostLineV(Doc,MAx3,MAy3,MAy2);
+	paintGhostLineH(Doc,MAx3,MAy2,MAx2);
   }
   else {
-    Doc->PostPaintEvent (_Line, MAx3, MAy3, MAx2, MAy3); // paint
-    Doc->PostPaintEvent (_Line, MAx2, MAy3, MAx2, MAy2); // paint
+	paintGhostLineH(Doc,MAx3,MAy3,MAx2);
+	paintGhostLineV(Doc,MAx2,MAy3,MAy2);
   }
 
   QucsMain->MouseDoubleClickAction = &MouseActions::MDoubleClickWire2;
@@ -347,12 +388,9 @@ void MouseActions::MMoveWire1(Schematic *Doc, QMouseEvent *Event)
   MAx3 = DOC_X_POS(Event->pos().x());
   MAy3 = DOC_Y_POS(Event->pos().y());
   Doc->setOnGrid(MAx3, MAy3);
-
-  MAx2  = DOC_X_POS(Doc->viewport()->width());
-  MAy2  = DOC_Y_POS(Doc->viewport()->height());
-
-  Doc->PostPaintEvent (_Line, Doc->ViewX1, MAy3, MAx2, MAy3);
-  Doc->PostPaintEvent (_Line, MAx3, Doc->ViewY1, MAx3, MAy2);
+  paintAim(Doc,MAx3,MAy3);
+  MAx2 = DOC_X_POS(Doc->contentsX()+Doc->viewport()->width()-1-2);
+  MAx2 = DOC_Y_POS(Doc->contentsY()+Doc->viewport()->height()-1-2);
   Doc->viewport()->update();
 }
 
@@ -1303,8 +1341,10 @@ void MouseActions::MPressWire1(Schematic *Doc, QMouseEvent*, float fX, float fY)
   //Doc->PostPaintEvent (_DotLine);
   //Doc->PostPaintEvent (_NotRop);
   //if(drawn) {
-    Doc->PostPaintEvent (_Line, 0, MAy3, MAx2, MAy3); // erase old mouse cross
+#if 0  //ALYS - it draws some garbage, not deleted because of possible questions
+	Doc->PostPaintEvent (_Line, 0, MAy3, MAx2, MAy3); // erase old mouse cross
     Doc->PostPaintEvent (_Line, MAx3, 0, MAx3, MAy2);
+#endif
   //}
   //drawn = false;
 
@@ -1312,6 +1352,10 @@ void MouseActions::MPressWire1(Schematic *Doc, QMouseEvent*, float fX, float fY)
   MAx3 = int(fX);
   MAy3 = int(fY);
   Doc->setOnGrid(MAx3, MAy3);
+
+//ALYS - draw aiming cross
+  paintAim(Doc,MAx3, MAy3);
+//#######################
 
   formerAction = 0; // keep wire action active after first wire finished
   QucsMain->MouseMoveAction = &MouseActions::MMoveWire2;
@@ -1366,8 +1410,10 @@ void MouseActions::MPressWire2(Schematic *Doc, QMouseEvent *Event, float fX, flo
       }
     }
 
-    Doc->viewport()->update();
-    drawn = false;
+    //ALYS: excessive update. end of function does it.
+	//Doc->viewport()->update();
+
+	drawn = false;
     if(set1 | set2) Doc->setChanged(true, true);
     MAx3 = MAx2;
     MAy3 = MAy2;
@@ -1375,7 +1421,12 @@ void MouseActions::MPressWire2(Schematic *Doc, QMouseEvent *Event, float fX, flo
 
    /// \todo document right mouse button changes the wire corner
   case Qt::RightButton :
-    if(MAx1 == 0) {
+
+#if 0
+	//ALYS - old code preserved because isn't clear - what it was???
+	//looks like deletion via painting.
+	//i'll delete it after possible clarification from team
+	if(MAx1 == 0) {
       Doc->PostPaintEvent (_Line, MAx3, MAy3, MAx3, MAy2); // erase old
       Doc->PostPaintEvent (_Line, MAx3, MAy2, MAx2, MAy2); // erase old
     }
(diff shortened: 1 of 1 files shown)


qucs/qucs/mouseactions.cpp
```diff
@@ -319,6 +356,8 @@ void MouseActions::MMoveElement(Schematic *Doc, QMouseEvent *Event)
  */
 void MouseActions::MMoveWire2(Schematic *Doc, QMouseEvent *Event)
 {
+//ALYS - i saved old code inside #if 0 block
```
felix-salfelder: please don't do that. we use git.
this way makes review harder, due to lack of an meaningful diff.

alysogorov: i can delete this PR, remove code if #if 0 blocks and create new PR.
is it good?
ps. i see in code some old commented lines, do not know who is author. May be delete them too?

felix-salfelder: > is it good?

no. you can add more commits to this PR. just push them to the branch.

you may even edit your local commits and force push. this is what you want to do just before a merge. please post a note such as "rebased".

> do not know who is author

use `git blame`

> May be delete them too?

only if they are close by/related. otherwise use a dedicated PR for cleanup

alysogorov: Felix, "git revert" and commit without #if 0 ...could save my PR? :)
also i see another one small bug, not solved in my current code. 
so i'll do another one commit.

felix-salfelder: a branch is merely a pointer to a commit. your PR lives as long as you dont delete the branch.

you should make yourself familiar with the `git reset` and `git rebase` commands, as well as with forced (non-forward) push. git might be a bit confusing for starters, but all of this makes perfect sense (promised).

alysogorov:
the reason of such weird sequence of commits is next:
I reverted the very first commit, and fixed it by removing #if 0 #else #endif blocks.
after commit the system marked this commit with red cross.
after investigation i found that i've created that my branch not from develop, but from master branch.
then i rebased my branch from master to develop without any changes in code.
so correct commit(for people's critics) is 1e96014.

felix-salfelder:
ok perfect. now (i assume that you intend to include only one commit with this PR), type

    git checkout <your_branch>
    git rebase -i develop

select the commits that you need and squash/fix them into just one (there's instructions at the bottom of the rebase control file).

then `git push <your_repo> +<your_branch>` and you are done (it will show up here).

note that there are *lots* of ways to achieve this (clean up your commits). this is my favourite, YMMV -- i'm not trying to force you into my way of doing things.

hint1: with an explicit/manual backup before the rebase, you might have less trouble reverting a failed attempt.
hint2: hint1 is not necessary, see `git-reflog(1)`. read carefully.

felix-salfelder:
4fd691a looks okay. i have no clue how to clarify the questions related to what the old code did.

i sugggest to leave it like that and merge, if nobody objects.

in3otd:
quickly tried and it does what it says; doesn't seem to break anything but I may try harder in the following days :grin:

@alysogorov if you feel motivated to do some further work on this, I'll have a couple of suggestions:
- as in the original code, if you zoom while in "wire mode" the cross cursor jumps around the mouse pointer, it goes to the upper left when zooming in and back to lower right when zooming out. Would be nice to fix that also
- I see you draw a double line for the wire being drawn to avoid having the cross cursor covering the wire; you could maybe define a new "paint event" [here](https://github.com/Qucs/qucs/blob/master/qucs/qucs/schematic.h#L64) so you could use a wider pen instead (as you noted in the comments)  in `Schematic::drawContents()` [here](https://github.com/Qucs/qucs/blob/master/qucs/qucs/schematic.h#L64)

alysogorov:
Ok, in3otd. I' ll look into the code, to solve this issues.

alysogorov:
i've found the reason of this problems. problem is quite common. this custom events for drawing, you mentioned in link,  are created from specific mouse handlers, and then are putted into the queue. during repaint they are painted, and queue has been CLEANED. because scroll and zoom handlers do not know about all this crosses and wires, they do not add this primitives to the queue. They just repaint all the scene with new offset and zoom. But because queue already has not this crosses and wires they are not painted.
so any usage of this events and cleaned queue will have the same bug - all such paintings won't be visible at scroll and zoom(if you do not move the mouse).
the easiest way is to restrict usage of this events(i don't inspect where it is used else), better to get rid of it, and use permanent object(hidden/visible) for example for cross. mice handlers must only set new coordinates of this object, hide or show it.
alex

in3otd:
(added to 0.0.19 milestone, as I see no reason to postpone this PR)

guitorri:
For the record I never undertook the what the old code did. I still don't get the `MAx1,2,3` variables. 

In any case keep in mind that this code, the rendering queue and direct calls to paint should be ported and eventually removed (if we follow the route in #563). So no pressure in removing the commented out code.

It did not break anything here either. 

in3otd, I cannot reproduce the jumping you mention.

guitorri:
I see. I think it has to do with grid snapping. If I set the grid to 1 or 50 the jumps get smaller or larger. As the zoom moves the canvas w.r.t the cursor the cross tries to snap. Maybe turn off the snapping during zoom?  By the way, the grid snapping seems to snap to grid/2 intervals not exactly to the grid points.

alysogorov:
this grid/2 alignment looks like it tries to align to grid BEFORE zoom. so aligns to "old grid"

guitorri:
Changed base to `release-0.0.19`. 
@alysogorov or @in3otd are you going to have time to look into this issue in the next day or two? Else I will try to fix the double line and leave the cursor jumping for later.

in3otd:
@guitorri , sorry, in the next day or two I won't have time, likely.

The cursor jumping can be certainly left for later, I proposed to take a look at that since alysogorov spent already some time on the code and was hoping he may have some ideas... Same also for the double line, it works, could maybe be improved.
I would already be happy to thoroughly verify that this PR has no side effects, the drawing code looks a bit obscure sometimes.

alysogorov:
Hi, to improve this double line, as in3otd said, we just need to add new "paintStyle" to enum -
typedef enum {_NotRop, _Rect, _Line, _Ellipse, _Arc, _DotLine, _Translate, _Scale}PE;
and paint not double, but single line  in functions
static void paintGhostLineH(Schematic *Doc, int fx,int fy, int fxx){
	Doc->PostPaintEvent (_Line, fx, fy-1, fxx, fy-1);
	Doc->PostPaintEvent (_Line, fx, fy+1, fxx, fy+1);
}

//paint ghost line - vertical
static void paintGhostLineV(Schematic *Doc, int fx, int fy, int fyy){
	Doc->PostPaintEvent (_Line, fx-1, fy, fx-1, fyy);
	Doc->PostPaintEvent (_Line, fx+1, fy, fx+1, fyy);
}

so change this functions to ->
//if we added the _FatLine kind.
//some hint - I use prefix "f" for formal function parameters. for years :)
static void paintGhostLineH(Schematic *Doc, int fx,int fy, int fxx){
	Doc->PostPaintEvent (_FatLine, fx, fy, fxx, fy);
}

//paint ghost line - vertical
static void paintGhostLineV(Schematic *Doc, int fx, int fy, int fyy){
	Doc->PostPaintEvent (_FatLine, fx, fy, fx, fyy);
}

and add of course, painting of this style(I do not remember now - where).

I can do it of course, but because of long mine absence, i do not know in which state the actual git is. also i do not know which kind of thick line you prefer, semi transparent or something.
alys.

as i said - all this postponed drawings in queue look not good, and it must be changed to objects, which understand their shape and draw it during repaint. so all the scene must be a set of objects with specific draw(..) method, not graphics primitives in queue. imo.

alys.

nvdl:
@alysogorov: I had to do some fixes just to make it work at the point. The issue I was having was that there are function pointers and variables pointing to previous points that are updated as user interacts with the schematic. My thought is to use some state variables to know what is happening at a specific moment and decide rather than keep on updating the variables and pointers that point to the "current" to be called function.

You can give this PR a look:
https://github.com/Qucs/qucs/pull/332

We may cleanup this together; looks like it needs writing from scratch.

alysogorov:
yes, nvdl, imo, the code there is quite complex, and has "side effects", and could be rewritten somehow. That's why i tried to minimally changed it in my bug fix.
Just now i cannot look deep into your code, but in couple of days, i hope to have more free time and look.
Seems it must be rebased, because cannot pass checks?
alys.

nvdl:
It needs to be converted to a state conscious class that knows what drawing, moving, editing etc. operation is going on rather than assigning different values to the function pointers and variables for clicked points. Most of the code was written in "printf" to see what was the next function to be called. Moreover, some signals do not make it to the "mouse handler" code that are needed in schematic drawing especially the keyboard presses. You can have a look at the patch blob when free.

alysogorov:
i'll look, but i doubt in such "state conscious class". imho it is better to have some generic interface with functions, kinda 
class genericActor {
  void onLBclick(int fx, int fy){};
  void onMBclick(...){};
  void onRBclick(...){};
  void onScroll(...){};
  ...
} 
and implementations of this interface for different modes of editing (add wire, delete node, add node...cope/paste... etc).
and mouse handler just must call appropriate mentioned current actor functions.
kinda _current_actor.onLBclick(miceX,miceY).
actor must not know about mouse, he just reacts on events at point (x,y).

then you need to change current actor(or push/pop it onto actors stack) depending on current editing mode...
it is my very first idea, because i did not see deep in current code and yours.
alex.

alysogorov:
ohh. i forgot to mention that functions in GenericActor are virtual :)
but actor does not paint something, there must be interface, for example, 
class genericCursor{
private:
int _x, int _y; //store coordinates
public:
  void setPos(int fx, int fy){_x=fx; _y=fy;}
  virtual void paint(){}; //to be redefined - paint somehow the actual cursor at his position
}
and there are actual cursor classes:

class NewWireCursor: public genericCursor{
 virtual void paint() override{..here we paint the cross line...}
}

when we change editing mode, we change currentActor and currentCursor...
of course idea is very basic, look into code needed
alys

nvdl:
I agree with the idea. Do we need multiple actors at the same time or in a chain?
In a move operation, if wire frame mode is used, can it be one or multiple actors; one for move, another for wire-framing? With chaining, new functions can be added later by adding new classes and inserting those into the chain.

Do we need cursor classes or is one enough? It changes the cursor via image or formula based on the active actor.

alysogorov:
1. better to say not about cursor, but about ActorLook. so actor implements behavior, ActorLook implements - how actor looks for user.
there could be more complex looks of actor for example - the rectangle - for actor "selectComponents"(when you select region).

of cource better to have pair - actor-actorLook, but some easy actors can have the same look(kinda simple cross) or use looks of another actors... but simple idea - actor and his special look also works well.
if actor implements some complex behavior - for example "select compjnents" - when button is not pressed - look draws simple cross, when LB pressed - actor fixes it as "base point" and calls his look's function - setBasePoint(x,y), look also fixes it and begins to draw rectangle, when mouse LB released - actor says to look - clearBasePoint(), then look again looks like simple cross. so look can change state, being managed by his actor...
of course there is a plenty of ways how to implement it.
i see that different editor modes must have their actors, and this actors must have their special look.
if you want do add new editor mode - you just need to add new actor(implementing - what to do) and his look(implementing - how to look).
where are chains there?
there could be "stack" - when one editor mode somehow calls another one, and continues when the called mode ends.. but i'm not sure there are this possibility in editor.

guitorri:
Before you start building on top of legacy code, keep in mind that the graphics will eventually be ported, see #563.

In the new framework the graphical items draw themselves and can keep track of its position and handle mouse coordinates transformations and whatnot. With the graphics out of the way, it I believe that with a simple state-machine can handle the different mouse interaction modes. 

For a toy example, have a loot at this: https://github.com/ClemensFMN/QucsQT5PoC
For a more elaborate example (forked from QUCS), this one: https://github.com/Caneda/Caneda

guitorri:
I don't see any risk of side effects.
As in3otd said, there is no reason to postpone this. Thanks!

nvdl:
@guitorri: Can the "QucsQT5PoC" be used as base code? I don't see any license.

guitorri:
Strictly speaking, I think we cannot just grab code. However the author was a contributor of QUCS at the time. You can see that he based some of the classes to mirror QUCS. So maybe it counts as derivative? In case you want to use it doesn't hurt to contact Clemmens.

guitorri:
If you are going to jump into the graphics port, you are welcome, but let us try to talk and coordinate a bit the efforts.

nvdl:
Have been away for a while from the discussions. Needed is:
Qt5 porting.
Isolation of graphics stuff from main stuff.
Ease of new symbols design.

We can discuss how to divide the functions across these phases.
Schematic drawing (wires, symbols etc.).
Netlisting (schematic to netlist for a simulator).
Simulation.
Plotting (I saw that there are some plotting issues, why not use external plotting APIs).
Data sinks (Octave or other bench support).

Can we make another thread for this discussion?

ra3xdh:
> Ease of new symbols design
The most of components symbols should be moved outside C++ code. Only dynamic components such as Subcircuit, LibComp, Verilog-A, SpiceComp, and Simulations should remain be written C++. Other components should be moved to XML. It will be needed to extent an existing LibComp device to allow handling of such components. It will not require C++ skills to design new components. I have took such experiments with XML components definition. Probably I will share my findings soon.

guitorri:
Please, shall we finish the Qt4 port first (drop Qt3Support), before going Qt5? This is solely rooted in the graphics. If we find a way to port the diagrams, the exercise in #563 may prove feasible enought to be followed thru. We use that as a reference and split the workload. Focus is key, I propose we finish up the graphics port more than anything else.

Isolation of model and view can be outlined first by introducing namespaces, them we split things into smaller files, with logical structure. 

Moving components away form C++ is something I like. less code to compile and maintain. I also like XML, but it open the discussion of what format. Shall we formalize the current format, adopt the one from Caneda? Craft yet another format?

I think we should discuss these thing if possible in the mailing list, so the discussion is not tied to an already closed issue.

felix-salfelder:
> The most of components symbols should be moved outside C++ code

nothing must be *moved*.

it must be made possible to use different sorts of components, THEN one can add xml-backed components. (and also the other components).

all components must become plugins, first the legacy components, and then new ones. dont create yet another qucs fork just for the choice of a file format.

the SYMBOL class has already been drafted, just implement a supplementary SYMBOL_FROM_A_CELL_BACKED_BY_XML (or whetever) class that inherits from it. put it into `contrib` (no review required). i will put the geda symbols into contrib. these approaches are not exclusive, so lets not pretend. and lets not start a bikeshed about which backends we prefer personally.

nvdl:
@felix-salfelder: You haven't changed in one year!
I assume, by moving he meant moving functionality away from recompilations all the time.

felix-salfelder:
> You haven't changed in one year!

should I?

> by moving he meant moving functionality away from recompilations all the time.

no, what he meant is "change the existing code" and "replace it for something different". this is the most evil approach. we need to *add to* the current code. and pave the way for future contributions, by *demonstrating* that code for different approaches can coexist.

(but i am only a software architect, not a circuit designer, so who knows...)

nvdl:
> should I?

Yes, use a bit of bitter sugar.

ra3xdh:
@nvdl , @felix-salfelder  Yes, I meant by "moving outside C++" that old C++-hardcoded components should be replaced by an XML library. Old C++ devices then should be dropped. Of course, this new implementation should be backward compatible with older Qucs versions. It may be achieved by tweaking of an existing `LibComp` component. Look at Caneda project. They have almost all components defined in libraries, but not in C++.

felix-salfelder:
> Old C++ devices then should be dropped

that should be "old C++ deviced should not be loaded by default". lets not repeat the mistakes.

note that you *will* have to write C++ code to pull in the contents of your XML files. i am just aking you to share the base class with the current "old" models.

nvdl:
I gravitate towards anything that will not need compilation.
Python classes are OK from my side if the Netlister (C++) can read them.
Every component is a ".py" file with SPICE model string(s) in it.

Can we come up with a generic container (generic enough for the current components) that has a graphical view, is grid friendly and has snap-able nodes. That can be the mother of all components.

We can use models in different formats if convenient; XML, Python etc.

Needed is a SVG file with nodes specifications and a SPICE model or models.
What else is needed for a component?

Maybe it is already done if one looks at the way the code base is brewing in so many pots.
Or possibly this is what XML library is about.

> that should be "old C++ deviced should not be loaded by default". lets not repeat the mistakes.

Agree, until the transition is complete.

alysogorov:
guys, i very love my thread, but this conversation must be moved somewhere, else content could die somehow.

PR #507, opened 2016-04-14
Enabled loading of unknown components

### LINKED ISSUE #506: Handle removed Verilog-A components when loading an schematic.
Schematics containing any of the removed Verilog-A components cannot be opened. The components are unknown...

The dialog below helps, but perhaps we should add a button to skip the missing component. So user can try to recover the schematic (use another component, of the Verilog-A loader ). 

image

ra3xdh:
This fixes #506. If Qucs find unknown component, a new dialog window with Yes/No buttons will be shown. User can press Yes and load schematic anyway, and press No and abort loading of schematic. Unknown components will be replaced by placeholders. I am using Subcircuits as placeholders. Maybe it's better to create some special placeholder device, which will be available for this purpose only? Should this PR will be for `release-0.0.19`?

### DIFF
--- a/qucs/qucs/components/component.cpp
+++ b/qucs/qucs/components/component.cpp
@@ -1571,16 +1571,24 @@ Component* getComponentFromName(QString& Line, Schematic* p)
 
   if(!c) {
     /// \todo enable user to load partial schematic, skip unknown components
-      QMessageBox* msg = new QMessageBox(QMessageBox::Warning,QObject::tr("Warning"),
-                                         QObject::tr("Format Error:\nUnknown component!\n"
-                                                     "%1\n\n"
-                                                     "Do you want to load schematic anyway?\n"
-                                                     "Unknown components will be replaced \n"
-                                                     "by dummy subcircuit placeholders.").arg(cstr),
-                                         QMessageBox::Yes|QMessageBox::No);
-      int r = msg->exec();
-      if (r == QMessageBox::Yes) c = new Subcircuit();
-      else return 0;
+      if (QucsMain!=0) {
+          QMessageBox* msg = new QMessageBox(QMessageBox::Warning,QObject::tr("Warning"),
+                                             QObject::tr("Format Error:\nUnknown component!\n"
+                                                         "%1\n\n"
+                                                         "Do you want to load schematic anyway?\n"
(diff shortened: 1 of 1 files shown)


felix-salfelder:
i vaguely remember that qucs is also supposed to run from the command line (especially for testing). what does `QMessageBox` do in that case?

i consider this as kluge to facilitate loading netlists that exist for unfortunate reasons. however, this now works for *all* components. we should check consistency with a possible long-term solution, before we merge...

what would a long term solution be? to me the translation

    nonexistant_component args => unknown component="nonexistant_component" args

makes some sense (and we should add an "unknown"/"catchall" symbol for that). if such symbol is availble, it should also implement how the user is to be bugged about its use and when (only for simulation, maybe).

i think this (or a variant) should go into 0.0.19. even as a post-release hotfix.

in3otd:
> i vaguely remember that qucs is also supposed to run from the command line (especially for testing). what does QMessageBox do in that case?

...it just shows the `QMessageBox` dialog, waiting for user input... as in the standard GUI mode, but without the GUI being actually open.

@felix-salfelder , do you think this PR can go in `release-0.0.19` as temporary "solution"?

felix-salfelder:
> dialog, waiting for user input.

nothing we should do in batch mode. unfortunately it was already there (line 1574), but it was only meant to catch critical errors. i would have preferred `exit(1)`.

> do you think this PR can go

it "solves" one problem and introduces 2 new ones. lets not merge this yet.

in the long run, the parser should be allowed to throw, but certainly (and right now) it should not use QWidgets and it should certainly not ask questions. it makes testing even harder...

maybe the least collateral-damage hotfix will be to just parse stuff as good as possible and display a warning in the console....?

ra3xdh:
Dialog showing logic may be the following:
* If Qucs GUI is available then show dialog and ask user to load unknown component or stop schematic loading;
* If Qucs GUI is not available then print error message into console and terminate application.

ra3xdh:
I have added error messages for CLI mode and some improvements for subcircuit placeholders. Now Qucs stops and prints an error message if unknown components are found.

guitorri:
I think that letting user load the partial schematic is a valid fix for the moment.

PR #473, opened 2016-02-13
Fix qucshelp

guitorri:

PR #508, opened 2016-04-17
Instantiate QApplication sooner so applicationDirPath() is valid

in3otd:
...now is my turn to realize i did a sloppy review :unamused: 

In #501 the path for the language files was changed to be relative to the application path, but if a `QApplication()` object is not instantiated before calling `QCoreApplication::applicationDirPath()` this latter actually returns an empty path, so the language files are not found (unless you run the application from the `prefix` directory).
This was already fixed for `qucshelp` in #473, but after #501 all the other tools needed to be updated .

I have checked this to work but please double/triple check...


guitorri:
All applications seem to find the translations here. Thanks!

PR #509, opened 2016-04-25
Power combining (develop)

andresmmera:
Hello there,

This PR is aimed to replace [PR #346 ](https://github.com/Qucs/qucs/pull/346/files). 

In addition to the existing features, it was added the LC equivalent of the Wilkinson power combiner (both single-stage and multistage).

On the other hand, it was added a panel on the right side of the window so as to display a preview of the current topology.
selection_054

selection_055


in3otd:
nice work! I'm doing some testing, here are some random comments:
- general:
  - do not disable window resizing, someone somewhere on some platform *will* need to resize it. I saw the comments in the related commit but, still, what's the advantage for the user in disabling resize? Can the aspect ratio of the SVG images be fixed ? (I seem to remember a similar problem with the Active Filter tool images)
  -  disable "number of outputs" (visible but not active) for configurations when only 2 outputs are supported
  - S-parameter simulation start/stop frequencies are wrong when the frequency units selected are not GHz. Or also when entering numbers < 1 in the frequency box.
  - bonus point: change length units for the lines to have always numbers between 1 and 999 (engineering notation). I guess is safe to use `misc::num2str()` for *everything* that needs to be human readable.
- if you select *Bagley* and then *Gysel* the *Number of outputs* shows 3, 5 and 7 but Gysel here supports 2 outputs only
- some Topologies generate the equation for `S11_dB`, some don't; I think it's better to have it always (maybe also for all the `Snn` terms)

- Wilkinson Multistage
  - Microstrip Implementation and Lumped Element**s** Implementation should probably be [radio buttons](http://doc.qt.io/qt-4.8/qradiobutton.html) even if the current implementation works fine
  - with Microstrips: the first microstrip has no W or L shown

- Tree
  - line parameters are not shown (I guess it's a choice for space reasons, right?)
  - microstrip implementation has something wrong, seems centered to half the design frequency

in3otd:
also you should set the main window icon
```
setWindowIcon(QPixmap(":/images/bitmaps/big.qucs.xpm"));
```
see e.g. the `QucsActiveFilter()` constructor.

andresmmera:
:+1: 
Ok, I'm gonna take a look.
Many thanks @in3otd !!

andresmmera:
I think I've almost fixed all the issues you pointed out. I'll probably push tomorrow.
Just a few comments:
> do not disable window resizing, someone somewhere on some platform will need to resize it. I saw the comments in the related commit but, still, what's the advantage for the user in disabling resize? Can the aspect ratio of the SVG images be fixed ? (I seem to remember a similar problem with the Active Filter tool images)

Well, I did so because at some point I thought it would be good to hide the substrate panel when the corresponding QCheckbox was unchecked. Somehow, that forced me to fix the window size. Anyway, I ended doing the same stuff the Active Filter tool does [(this)](https://github.com/Qucs/qucs/blob/master/qucs/qucs-activefilter/qucsactivefilter.cpp#L222)

> (Tree wilkinson combiner) line parameters are not shown (I guess it's a choice for space reasons, right?)

Yes, of course. Otherwise, it becomes a mesh...

> bonus point: change length units for the lines to have always numbers between 1 and 999 (engineering notation). I guess is safe to use misc::num2str() for everything that needs to be human readable.

Ok, done. The only inconvenient is some corner case. For example, at very high freqs the user may select 'mils' and the final result might be given in 'nm'... but I guess nobody is going to use this tool at 3 THz or so :-)

guitorri:
The microstrip implementations do not have a corresponding .svg figure right?
Do you think is worth adding?

guitorri:
The Travelling wave does not create the S11_dB Equation. (no big deal)

guitorri:
For different number of outputs the figures are also not there. I guess we can leave it as is, doesn't make sense to create figures for all combinations.  (Some day we will be able to use the schematic as a library so we can render the schematic templates and get a figure back...)

Let me know if you want to change something else.

andresmmera:
> The microstrip implementations do not have a corresponding .svg figure right?
Do you think is worth adding?

Honestly, I didn't really think about that. I just wanted to give a visual clue about the combiner structure. I can create the corresponding svg files for the microstrip implementation (hopefully) this weekend.

> The Travelling wave does not create the S11_dB Equation.

My bad. Didn't notice that...

> For different number of outputs the figures are also not there.

Yes, I did that because otherwise the schematic becomes a mesh when you select a big number of outputs. I know that the isolation between output ports (i.e., Sxy, x != y != 1, ) is an important figure, but I thought that it would be more convenient if the user adds the equation manually if he/she's interested on that.

> Let me know if you want to change something else.

I'll add the S11 equation for the TW and tree combiners and then I'll add the figures for the microstrip.

andresmmera:
I've just pushed two commits in order to address the issues you mentioned yesterday

andresmmera:
I don't know why Travis fails... I rerun it just in case, but the error is still there.
It says:

```
 rm -f TAGS ID GTAGS GRTAGS GSYMS GPATH tags
rm -f cscope.out cscope.in.out cscope.po.out cscope.files
make[2]: Leaving directory `/home/travis/build/Qucs/qucs/qucs-0.0.19/_build'
rm -f config.status config.cache config.log configure.lineno config.status.lineno
rm -f Makefile
./qucs/qucs/qucs_.cpp
make[1]: Leaving directory `/home/travis/build/Qucs/qucs/qucs-0.0.19/_build'
make: *** [distcheck] Error 1

```
Any guess?

felix-salfelder:
i don't know what triggers this. 6e3420e seems to fix it

andresmmera:
:+1: Thanks

I'll wait until you merge that and then I'll rerun Travis again.

felix-salfelder:
#722

andresmmera:
Cool!
Thanks @felix-salfelder

guitorri:
I rebased it locally. History is getting messy.
Can I force push to your branch? Please enable "Allow edit from maintainers" on the rigth.

andresmmera:
Feel free to do whatever you need.
The "Allow edit from maintainers" checkbox is already enabled

guitorri:
Thank you. I was giving the wrong branch name for update...

Issue #513, opened 2016-05-25
qucs crash when editing component properties


kraus1049:
### Reproducing way

```
1. Put "dc Voltage source"
2. Right click and choose "Edit Properties"
3. Press Enter key twice
```

qucs will fail into segmentation fault after step 3.

### qucs version

```
$ qucs --version
Qucs 0.0.19 (9d924ca)
```

### Environment

```
＄ cat /etc/lsb-release 
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=16.04
DISTRIB_CODENAME=xenial
DISTRIB_DESCRIPTION="Ubuntu 16.04 LTS"

$ arch
x86_64
```

andresmmera:
Hello, 

I've tracked down the seg fault. It seems that this [line](https://github.com/Qucs/qucs/blob/master/qucs/qucs/components/componentdialog.cpp#L723) should be:

  if ( row < prop->rowCount()-1) {

row is an integer ranging from 0 to prop->rowCount()-1, so Qucs tries to access a non-existing element...

prop->setCurrentItem(prop->item(row+1,0));

row+1 > prop->rowCount()-1 => here's the seg fault

Should I open a PR with the patch?

in3otd:
> Should I open a PR with the patch?

why shouldn't you? :grin:

andresmmera:
just checking :-)

in3otd:
Fixed in #514.

Thanks @kraus1049 for reporting the issue!

PR #514, opened 2016-05-25
Fixed issue #513

andresmmera:
Issue #513 

When editing the last property of a component, Qucs crashed after pressing
enter (if was a lineedit)

### DIFF
--- a/qucs/qucs/components/componentdialog.cpp
+++ b/qucs/qucs/components/componentdialog.cpp
@@ -692,15 +692,17 @@ void ComponentDialog::slotApplyChange(const QString& Text)
 void ComponentDialog::slotApplyProperty()
 {
   // pick selected row
-  QTableWidgetItem *item = prop->selectedItems()[0];
-
+  QTableWidgetItem *item = prop->currentItem();
+  
+  if(!item)
+    return;
+  
   int row = item->row();
 
   QString name  = prop->item(row, 0)->text();
   QString value = prop->item(row, 1)->text();
 
-  if(!item)
-    return;
+ 
 
   if (!ComboEdit->isHidden())   // take text from ComboBox ?
     edit->setText(ComboEdit->currentText());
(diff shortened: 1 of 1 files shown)


in3otd:
the fix is fine but there is another little detail I suggest to modify:
the crash was happening because the line
```
QTableWidgetItem *item = prop->selectedItems()[0];
```
tried to get the first item from an empty list, when the element past the last was erroneously "selected".
There was a check below, which was maybe intended to handle it
````
if(!item)
  return;
````
but it's useless, since `item` is already used before the check, and if it was `null` it would have caused a segfault already.

So I think it would be better to modify the beginning of the function as follows:
````
  // pick currently selected row
  QTableWidgetItem *item = prop->currentItem();

  if(!item)
    // should actually not happen...
    return;

  int row = item->row();
  ....
````
it should not change anything for the "normal" cases, but should avoid a crash in case the currently selected item does not exist, for some reason.

in3otd:
...maybe add a `Q_ASSERT(item);` before the `item` check, so people building with debug enabled will still have the application aborting if this happens...

in3otd:
> ...maybe add a `Q_ASSERT(item)`

uhm, no, that was a bad idea: there are a couple of components (at least) that have no editable properties (e.g. the current and voltage probes) and for them is then normal that the currently selected property does not exist, so the application should not abort.
...please remove the `Q_ASSERT()`... :grimacing:

andresmmera:
Ok, I've trusted your words almost without thinking 😂😂
No problem, I'll correct it back as soon as possible

Cheers

in3otd:
> Ok, I've trusted your words almost without thinking :joy::joy:

...that's bad, never trust any code that's not yours :grin:

andresmmera:
I've just done it...
I've squashed the previous commits, just in case

in3otd:
merged in `release-0.0.19` (c9d9ea51ce32e25691300fe9a15cdbce4873eb33)

I have reworded the commit message a bit and removed the no-op comment "Added Q_ASSERT/Removed Q_ASSERT()"...

Issue #512, opened 2016-05-22
unable to compile qucs-core


stonecome:
Hi,all
I am meeting an error when I compile qucs-0.0.19(master branch) according to the README file. It stops in the qucs-core compiliation. Then I make the qucs-core after commands(./bootstrap ./configure), it reports:
Projects/qucs/qucs-core# make
make  all-recursive
make[1]: Entering directory `/home/leiyc/Projects/qucs/qucs-core'
Making all in src
make[2]: Entering directory`/home/leiyc/Projects/qucs/qucs-core/src'
g++ -DHAVE_CONFIG_H -I. -I..  -I../src/math -I../src/components -I../src/components -I../src/interface -I.   -O2 -std=gnu++11 -pipe -fno-exceptions -fno-check-new -ldl -rdynamic -MT gperfappgen.o -MD -MP -MF .deps/gperfappgen.Tpo -c -o gperfappgen.o gperfappgen.cpp
mv -f .deps/gperfappgen.Tpo .deps/gperfappgen.Po
/bin/bash ../libtool --tag=CXX   --mode=link g++  -O2 -std=gnu++11 -pipe -fno-exceptions -fno-check-new -ldl -rdynamic   -o gperfappgen gperfappgen.o  -lm 
libtool: link: g++ -O2 -std=gnu++11 -pipe -fno-exceptions -fno-check-new -rdynamic -o gperfappgen gperfappgen.o  -ldl -lm
./gperfappgen  > gperfapphash.gph
/usr/bin/gperf -I -m 8 gperfapphash.gph | sed -e 's/{""},/{"",0},/g' > gperfapphash.cpp
make[2]: **\* No rule to make target `parse_citi.hpp', needed by`all'.  Stop.
make[2]: Leaving directory `/home/leiyc/Projects/qucs/qucs-core/src'
make[1]: *** [all-recursive] Error 1
make[1]: Leaving directory`/home/leiyc/Projects/qucs/qucs-core'
make: **\* [all] Error 2
The environment info of my compilation is that:
Debian 3.2.68-1+deb7u2 i686 GNU/Linux
gperf 3.0.3
automake (GNU automake) 1.10.3
gcc (Debian 4.7.2-5) 4.7.2
bison (GNU Bison) 2.5 
lex 2.5.35...
Am I missing something important? Any guess?

Thanks in advance
Best Regard

guitorri:
Seems related to #295 and #492.

stonecome:
[...]

PR #515, opened 2016-05-27
blacklist automake-1.10, like -1.11

felix-salfelder:
this is related to the ypp->hpp rules, which only newer automakes do
the way we need.

i assume that automake-1.10 causes #512. I DID NOT CHECK (sorry).

### DIFF
--- a/qucs-core/bootstrap
+++ b/qucs-core/bootstrap
@@ -24,23 +24,6 @@
 #
 
 AUTOMAKE=${AUTOMAKE:-automake}
-AUTOMAKE_VERSION_CHECK=${AUTOMAKE_VERSION_CHECK:-yes}
-
-if [ "$AUTOMAKE_VERSION_CHECK" = yes ]; then
-	AUTOMAKE_VERSION=$($AUTOMAKE --version | head -1 | cut -d' ' -f4)
-	case ${AUTOMAKE_VERSION} in
-		1.1[01]*)
-			cat >&2 <<__EOF__
-We experience trouble with your automake (version $AUTOMAKE_VERSION).
-Giving up.
-
-If you feel adventurous, please set AUTOMAKE_VERSION_CHECK=no in your
-environment and try again.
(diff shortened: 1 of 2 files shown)


in3otd:
would it make sense to specify the minimum version required as an [option](https://www.gnu.org/software/automake/manual/html_node/List-of-Automake-options.html#index-Option_002c-version)  to `AM_INIT_AUTOMAKE` in [`configure.ac`](https://github.com/Qucs/qucs/blob/release-0.0.19/qucs-core/configure.ac#L25) instead?. This will cover also case where one tries to use a version older than 1.10...

If the `automake` version installed is too old, the user will get a message like `error: require Automake 1.12, but have 1.9`. (Provided that [`bootstrap`](https://github.com/Qucs/qucs/blob/release-0.0.19/qucs-core/bootstrap#L63) there is properly modified to exit if a command fails (`|| exit 1`) as the top-level  [`bootstrap`](https://github.com/Qucs/qucs/blob/release-0.0.19/bootstrap#L16) does)

felix-salfelder:
that makes a lot of sense.

just noticed, i had pushed the branch to the wrong repo... i can't change that (github won't let me), sorry `:|`

guitorri:
Should this go into release-0.0.19?

felix-salfelder:
yes. (sorry for the mess)

in3otd:
thanks; what about adding `|| exit 1` to [`bootstrap`](https://github.com/Qucs/qucs/blob/release-0.0.19/qucs-core/bootstrap#L63) so that it clearly stops when the `automake` version check fails? Helps the distracted users and those with a misconfigured locale (because in this case perl prints a page of warnings about `perl: warning: Setting locale failed.` and this makes all the useful messages scroll away)
(may be good to exit on errors when running the other tools also)

guitorri:
Rebased to `release-0.0.19` in PR #574.

PR #518, opened 2016-06-04
fix makefile portability

ra3xdh:
This PR contains various fixes for Windows:

1. Fixed `va2cpp` make rules. MinGW32 make tries to execute commented line and Verilog-A modules cannot be compiled
2. Added Octave version checking for Octave >= 4.0.0. Octave>=4.0.0 has GUI enabled by default and cannot start inside Qucs. If version is >=4.0.0 `octave-cli` is forced to start.

### DIFF
--- a/qucs-core/src/components/verilog/va2cpp.makefile
+++ b/qucs-core/src/components/verilog/va2cpp.makefile
@@ -57,7 +57,7 @@ va2cpp: $(MODEL).cpp
 
 .va.cpp: $(XML_FILES)
 	@echo '# va2cpp - Creating C++ sources.'
-	# BUG: must catch stray spaces in ADMSXML, adding extra quotation
+# BUG: must catch stray spaces in ADMSXML, adding extra quotation
 	"$(ADMSXML)" $< \
                            -I "$(INC)"                    \
                            -e "$(INC)/qucsVersion.xml"    \


qucs/qucs/octave_window.cpp
```diff
@@ -119,7 +113,30 @@ bool OctaveWindow::startOctave()
   // append process PATH, othewise Octave does not find gnuplot
   QProcessEnvironment env = QProcessEnvironment::systemEnvironment();
   env.insert("PATH", env.value("PATH") + sep + QucsSettings.BinDir );
+#ifdef __MINGW32__
+  env.insert("PATH", env.value("PATH") + sep + QucsSettings.OctaveBinDir.canonicalPath() );
```
felix-salfelder: this specializes line 115 for a particular platform/compiler. line 115 is wrong already, better remove it.

`PATH` is an environment variable. should be taken from the environment. customizing the environment does not (should not, must not) require patching qucs here.

ra3xdh: New versions of Windows Octave need to load some DLLs from `octave_root/bin` directory. Otherwise Octave failed to start. Windows hasn't an equivalent of `LD_LIBRARY_PATH`. And I haven't found other way than `PATH` modification or changing `octave` process working directory. QProcessEnvironment exists only at process execution time and doesn't modify system environment.

But there may exist another way without version checking. We can simple test `octave-cli` executable. If `octave-cli` exists it will be forced to start instead of `octave`, and `octave` will be started otherwise. I will update my PR later today or tomorrow after checking this.

felix-salfelder: > I haven't found other way than `PATH`

Using the `PATH` environment variable is perfectly fine and correct. you do not need to change qucs for that. `QucsSettings.OctaveBinDir` does not make any sense, because that's what `PATH` is for. PATH is explained everywhere on the internet, first hit [here](http://superuser.com/questions/284342/what-are-path-and-other-environment-variables-and-how-can-i-set-or-use-them).

you likely fell for a misconception in line 115. bad ideas are viral, don't repeat that mistake...

ra3xdh:
I have replaced previous patch. New patch version doesn't require Octave version checking and `PATH` modification. It check if `ocatve-cli` available and starts it. Usual `octave` will be started otherwise.

felix-salfelder:
6464849 makes sense, fixes a portability issue i added some day.

bca32b6 still does not make a lot of sense, now i kind-of understand what you are trying to achieve.
if there's a need to run a different octave, use an override environment variable. this is already implemented for `QUCSATOR`, which overrides `qucsator`. i suggest to make it start with the letters `QUCS`, so its purpose is clear. `QUCS_OCTAVE` sounds best to me.

also, it seems, `QucsSettings.Octave*` is not reliable (surprise). someone tried to reinvent `PATH`, and failed. use `PATH` instead. see my messages attached to the commits you deleted.

ra3xdh:
I want to provide platform-independent way to launch Octave form Qucs-GUI without modification of system settings and environment variables. Octave >=4.0.0 fails to start from Qucs GUI on Windows. You can use Wine to reproduce this bug.  If `octave.exe` executable location is in path it can be successfully started. Other executable `octave-cli.exe` should not be in `PATH` to start without errors.

Introduction of something like `QUCS_OCTAVE` environment variable is bad idea. I will not introduce it.

I can remove bca32b6 if you add a note in documentation on proper environment variable setting as you see it. It is not a big deal. But you will explain every Windows user how to set system environment variables properly when you will get reports on Octave starting error.

felix-salfelder:
> Introduction of something like QUCS_OCTAVE environment variable is bad idea.

now, that's your personal preference. lets still get this fixed.

if you don't like environment variables, then only
- unhardcode `"octave"`, take from settings instead (line 93),
- ignore QUCS_OCTAVE environment variable (!?),
- use PATH (see how qucsator is called) when executing octave.

> But you will explain every Windows user how to set system environment variables properly when you will get reports on Octave starting error.

i prefer this to patching qucs everytime some platform starts to behave differently/inconsistently. you wouldn't want to modify qucs if it was already done right. so *do it right*. there will be other platforms and new issues in the future (promised!).

if you need a starter script for a deviant platform, implement it. that's part of the qucs package for that particular platform.

I will/might merge 6464849 later today, as time permits. looks independent enough.

ra3xdh:
I have excluded  bca32b6 from patch. You can proceed with merging now. Let's remain the proper solution for future. Adding octave location to system `PATH` may be temporary solution. Good solution will be to replace `OctaveBinDir` by something like `OctaveExecutable` in settings, but it will break compatibility with older Qucs versions.

felix-salfelder:
> `OctaveExecutable` [..] will break compatibility with older Qucs versions.

somehow i managed to not break compatibility, when i fixed qucsator invocation. try this:

- set octave = settings.octavebindir + "octave"
- if OctaveExecutable is set, then set octave = OctaveExecutable
- (likewise, if env(QUCS_OCTAVE) is set, then assign octave = env(QUCS_OCTAVE)).

this wont change behaviour, if Settings.OctaveExecutable is not set by default. "not set" is also the default for environment variables...

NB: you also need to remove the file check, it does not make sense. cf. 7eacbcec4e6254f67 and its parents.

in3otd:
> Added Octave version checking for Octave >= 4.0.0. Octave>=4.0.0 has GUI enabled by default and cannot start inside Qucs. If version is >=4.0.0 octave-cli is forced to start.

This seems to be a problem only in Windows (and Wine , apparently), in Linux, Octave 4.0.0 works fine (OS X I have no idea). The Octave code checks if `stdin` is a terminal and if it's not, it does not start the GUI, without needing to force this with any command-line switch.
The problem does likely not come from Octave itself, as if you do `echo "X=1" | octave` from a terminal it works as expected in both Windows and Linux - the result is shown in the terminal and the Octave GUI is not started. It may then be due to the `QProcess` implementation in Windows and maybe fixed in Qt5 (we already saw another Windows-only bug in `QProcess` which was fixed only in Qt5)

Issue #522, opened 2016-06-12
Exported graph image has the markers cut


in3otd:
a little thing, that I find somewhat annoying:
select a graph with markers, then export it as image (`File` -> `Export as image` then check `Export selected only`): the exported image is sized to fit the graph only, but the marker lines are visible.

Either we should:
-  not show the markers at all (not nice), or
- size the exported image to include the graphs (good), or
- allow the user to choose (even better).

Example:
- exported image with cropped markers (current behaviour)
  ![exported_graph_ko](https://cloud.githubusercontent.com/assets/9018179/15992548/235fc228-30cf-11e6-86e3-cf8526780f66.png)
- exported image as it should be
  ![exported_graph_ok](https://cloud.githubusercontent.com/assets/9018179/15992551/3496c348-30cf-11e6-8c73-53dd6749e099.png)

ra3xdh:
I confirm this issue. There also exists similar issue with wire labels. I will add fix soon. It's strange that I never met it before. I often use export of diagrams and schematics to images for articles preparation.

ra3xdh:
@in3otd, See #523 for fix.

guitorri:
Fixed with #523.

PR #523, opened 2016-06-13
Fix of #522

ra3xdh:
This PR contains fix for #522 and some refactoring at the `ImageWriter` class. @in3otd , please check on your side.


in3otd:
thanks, seems to work *almost* as expected.
If I have a Diagram with Markers and select only the Diagram, when printing the selection the marker lines in the Diagram area are still printed (as before) but the printed area is cropped to fit the Diagram only (as expected). Ideally it should be as for the schematic and wire labels, if the wire labels are not selected the labels lines are not visible.
I guess the issue with Diagrams comes from [here](https://github.com/Qucs/qucs/blob/master/qucs/qucs/schematic.cpp#L710) - the markers are always painted, even if not selected.

ra3xdh:
@in3otd I have divided `Diagram::paint()` into two methods: `Diagram::paintDiagram()` and `Diagram::paintMarkers()` This makes possible to paint diagram and markers independently. Check the last commit. Now markers will be not printed if they are not selected.

in3otd:
printing selected object seems to work well now, thanks.
There are still some details which are not fully right when printing the entire document (present also in the current `release-0.0.19` code):

- expand the [enclosed project](https://github.com/Qucs/qucs/files/380463/print_issues.zip)
- open the schematic
  - do *File*->*Export as image* then *Export*
  - the wire/node labels on the bottom are slightly cut (and there should be some margin anyway) in the exported image
- open the Data Display (do not move anything inside)
  - do *File*->*Export as image* then *Export*
  - the diagrams are cut on the left and right side in the exported image
  - if you move "something" in the Data Display then it's printed correctly, sometimes

ra3xdh:
I have corrected image size determination method. Previous one was unreliable in some cases. See the latest commit.

in3otd:
thanks, it's quite better, but not yet perfect, sorry...

In the example project linked above
- open the schematic
- move one wire label on the top side, above the components
- select a few components and the label above the components
- do *File->Export* as image then *Export*, check *Export selected only*
- the label above the components is cut (happens also if the label on the right or left, but not on the bottom...)

Besides, if printing the entire schematic, labels on the bottom, below any components, as in the example project linked above do not have any margin below them, while the margin is correctly present on the other sides.

ra3xdh:
Current implementation inherits label bounding box determination method from the legacy code `Schematic::sizeOfAll()` method. It seems it has a bug. I will try to dig into this code.

ra3xdh:
@in3otd , I have modified legacy code. Now label bounding box should be determined correctly. Here are two examples of exported images. The first uses "Export selected" and the second uses "Export All":
export1
export

in3otd:
thanks for the quick fix.

While testing I noticed one inconsistency: the bounding box calculations take into account labels only if [their associated wire is selected](https://github.com/ra3xdh/qucs/blob/107ea2f991d91638fae3301aca9a662213c13ec4/qucs/qucs/imagewriter.cpp#L313-L327) while the painting routine print selected labels [even if the associated wire is not selected](https://github.com/ra3xdh/qucs/blob/107ea2f991d91638fae3301aca9a662213c13ec4/qucs/qucs/schematic.cpp#L663-L669). The end result is that wire labels can be cut if the corresponding wire is not selected.
It's certainly a corner case, but I'll suggest following what the painting routine does and so include the selected labels in the bounding box calculations unconditionally.

Another details is that if you toggle the *Original size* checkbox while *Export selected only* is checked, the *Width*/*Height* lines show the full schematic size and not just the selected components area size.

I've pushed a PR to your branch with the small fixes for these things, please take a look.

ra3xdh:
@in3otd , Your code works as expected. If you have no other additions, this PR could be merged after CI will be passed.

guitorri:
Seems good to me as well. Merging.

Issue #330, opened 2015-07-23
review process reviewed


felix-salfelder:
the current review process has some issues. @guitorri has mentioned a WIP branch in #326. i call it "unstable". here http://nvie.com/posts/a-successful-git-branching-model/ it is "develop". will stick to "develop" here, old comments still reference "unstable".

from my experience, we should adopt some process that already works for other projects. here's my favourite
- everybody may open PRs, a PR is a request to merge code, unless stated otherwise
- an "issue" is a request to just discuss things.
- everybody may review PRs 
- a review can involve discussion about functionality and implementation details.
- a reviewer should read the code, filter out obvious shortcomings.
- a reviewer is entitled to request changes.
- a discussion that ends with at least two persons in agreement (typically the author and a person different from the author) is a positive review, if
- in addition some formal criteria are met
  - commit message
  - intermediate commits compile
  - (where is the full list?)
  - possible criterion: code coverage / no dead code
  - consistent style (maybe later)
- an official may now merge the PR into the `develop` branch (see below), (s)he does not have to.
- it makes sense to let the positive review sink in for a few seconds. but generaly the merge should happen within the next days after a positive review (if time permits)

the `develop` branch is the branch next to master, that will be merged into master by the release manager. `develop` does not have to follow master (sometimes master needs a backported bugfix).
_maybe_ there should be a `release` branch on top of `master`, where one can prepare releases, while `develop` is still being worked on.

that's all from me

EDIT1: added timing constraints
EDIT2: s/unstable/develop/
EDIT3: starting to spell out possible criteria

felix-salfelder:
@ra3xdh if you don't agree with the unstable branch model, can we discuss this here instead of randomly mess with each others issues?

in3otd:
I see two main, not completely related, subjects here:
- the PR approval process
- the code development model 

(a PR approval process does not necessarily imply a certain code development model and viceversa)

Regarding the first, the process above is similar to the (informal) one we had so far. Looks fine, when things go well, everybody agrees and the sun shines. But, as we saw recently, the difficult part is how to handle any disagreement.
- what to do with the enthusiastic contributor who happily approves any PR, because ey wants to see some fresh air coming into the project at all cost? Will eir approval always count?
- how to handle that grumpy supporter, who has been around long enough that ey thinks that the long-standing bugs have now actually become features and should not be touched, so ey rejects most PRs. Will eir disapproval be taken into account?
[...]

PR #524, opened 2016-06-14
set proper tex environment. particularly for out-of-tree build

felix-salfelder:
TODO: propagate to 65 other Makefiles
      (better: deduplicate)


in3otd:
started to look at this, then got carried away and ended up with #537, let's see how it works.

felix-salfelder:
distcheck is broken on travis/apple. `configure: error: please install latex2html` would that be possible, or shall i try to disable distcheck there?

felix-salfelder:
@in3otd, finally i got past travis. `qucs-0.0.19 archives ready for distribution`. can you please have a look?

theres no commit like 3911408, i don't know exactly what it was for.

in3otd:
@felix-salfelder , I saw your message, tried to find some time but unfortunately I have too many things going on at the moment....

(continues with general remarks about the development process at #330)

felix-salfelder:
> I have too many things going on at the moment....

that's not much of a problem. if nobody else cares about this particular PR, then we should maybe just leave it like that... for now.

(no, i will not merge my commits without review, as you suggest in #330.)

qucs-doc/qucstex.mk
```diff
@@ -0,0 +1,31 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this package; see the file COPYING.  If not, write to
+# the Free Software Foundation, Inc., 51 Franklin Street - Fifth Floor,
+# Boston, MA 02110-1301, USA.
+#
+
+TEX_ENV = export TEXINPUTS=$(srcdir):$(srcdir)/..:./..:
+
+.tex.pdf:
+	$(TEX_ENV); $(PDFLATEX) $<
```
guitorri: I had to add `-shell-escape` here (using MacTex here):
`$(TEX_ENV); $(PDFLATEX) -shell-escape $<`

Only in the first pass seems to fix the following error:
  `Package pgf Warning: Plot data file 'activefilter.pgf-plot.table' not found.`

Reference : http://texample.net/tikz/examples/gnuplot-basics/

With that I can build in-tree.

guitorri:
In-tree works fine here. Out-of-three it stops with:

```
[...]
epstopdf ($Id: epstopdf.pl 36129 2015-01-24 00:30:11Z karl $) 2.23
!!! Error: Cannot open pics/rc_ac_circuit.eps: No such file or directory

! Package pdftex.def Error: File `pics/rc_ac_circuit-eps-converted-to.pdf' not 
found.
[...]
```

It seems the generated files are still in-tree, or cannot be found:

```
[TEX_ENV_mac]~/git/qucs/qucs-doc/tmp $ find .. -name 'rc_ac_circuit.eps'
../tutorial/textmode/pics/rc_ac_circuit.eps
```

felix-salfelder:
thanks guilherme! will try to fix that.

i am stuck here

    error: 'zpoly' undefined near line 74 column 8
    error: called from
        ../../technical/stabgear.m at line 74 column 6
    Makefile:577: recipe for target 'stabgear.eps' failed
    make[1]: *** [stabgear.eps] Error 1
    make[1]: Leaving directory '/home/felix/extern/qucs/qucs-doc/build/technical'

any clue/hint?

guitorri:
There is a `qucs-doc/technical/zpoly.m`, maybe tweak octave search path?

felix-salfelder:
d'uh, must have been due to out-of tree then. i've just looked in the wrong place ':|'

thanks alot

felix-salfelder:
> `Error: File `pics/rc_ac_circuit-eps-converted-to.pdf' not found`

fixed (hopefully).

qucs-doc/tutorial/textmode/content.tex
```diff
@@ -151,7 +151,7 @@
 
 \begin{figure}[h]
 \begin{center}
-  \includegraphics[angle=0,width=0.8\linewidth]{rc_ac_python.eps}
```
in3otd: removing the `.eps` extensions was done also in 110a4f4906a62da5b1eabbbe339402b19b04ae0d , which  you already merged; for some reason the change got lost (??).
It seems there are some recent versions `pdflatex` that can use `eps` file but also mine here cannot.

in3otd:
@felix-salfelder , regarding

> theres no commit like 3911408018a6f5403e4bc6bd0fa75cc9a0093926, i don't know exactly what it was for.

I added it after you merged my branch; as briefly described, it's just to make sure that the `pdf` files are created also if one simply does a `make install`, without running `make` before. This because you can compile and install `qucs` and `qucs-core` with just `make install` so I thought this should work also for `qucs-doc`. But it seems you are now cleaning up the `Makefile.am` files to properly use the makefile recursion, so `make install` should also automatically build the docs when needed in this case.

qucs-doc/tutorial/textmode/content.tex
felix-salfelder: > for some reason the change got lost 

indeed. i dont know what happened. but more changes are lost (such as 69a8725d0cc90f8e5c4f9a6e97687e745d2dd7cd)

sorry about that. will try to sort this out!

felix-salfelder:
noted. i did not understand either the purpose nor the function of the `pdf` target (and it messed up the build order quite a bit). i think we are better off without it.

felix-salfelder:
in3otd: i have now rebased everything to amend the mismerge, sorry about that.

now it seems to work, except on apple:

    Making check in qucs-doc
    make[1]: *** No rule to make target `check'.  Stop.

maybe due to `--disable-doc`. will have another look later.

in3otd:
I may have missed something, but if `--disable-doc` was actually passed, `make check` should not even enter in `qucs-doc`, or? Is [this] (https://github.com/Qucs/qucs/pull/524/commits/05f7a43593b8eab70c47f05bb86b517d6cf4c6f9#diff-354f30a63fb0907d4ad57269548329e3R162) right?

I guess some old doc targets no longer exist, please update the [README] (https://github.com/Qucs/qucs/blob/master/qucs-doc/README#L96-L124)

felix-salfelder:
> make check should not even enter in qucs-doc

i am not sure about that. strange enough, at home, i get a Makefile in `foo/qucs-doc/` with a proper `check` target after `mkdir foo; cd foo; ../configure --disable-doc`. whichever way, there's a bug i need to find...

> please update the README

good point. how much of this is reasonable? i think it makes more sense to just build everything (according to which tools/flags are present during configure), and do selective builds by `cd somewhere; make` as usual. (least surprise).

in3otd:
yep, it's strange that the `check` target doesn't even appear to exist on the Mac Travis build... BTW, won't it be better if `--disable-doc` didn't even touch the `qucs-doc` directory?? Currently it's the top level `Makefile` that decides not to enter in `qucs-doc`, but the directory is configured in any case. In the future, we may want to stop with an error if the programs needed for building the docs are not present. But this may not really belong to this PR...

Ok for the "manual selective building", not sure why there were explicit separate targets in the past... "it has always been that way". And BTW, I just saw that these targets were used [here] (https://github.com/Qucs/qucs/blob/develop/contrib/qucs-release.sh#L98-L117) so better put a note that now that's broken and/or ring @guitorri to let him know :grin:

felix-salfelder:
> BTW, won't it be better if --disable-doc didn't even touch the qucs-doc

this will break toplevel `make dist`. note how automake implements DIST_SUBDIRS as the collection of all *potential* SUBDIRS. arguably, it's just a convention that `make dist` is unaffected by configuration but now that's the expected (and documented, traditional, established...) behaviour. let's not mess with it.

> "it has always been that way"

it seems the transition to automake has never been completed. i will add `$alias: all` warppers unless guilherme is OK with removing that stuff (@guitorri: please let me know).

in3otd:
uhm, `make check` appears to work but configuring `qucs-doc` says [`please install latex2html`](https://travis-ci.org/Qucs/qucs/jobs/158843011#L1120-L1121)
(and this is an example of why I'd like `configure` in `qucs-doc` to be skipped if `--disable-docs` is passed, so one can avoid having to scroll back 3000 lines to find the error. But I understand this will cause other issues)

felix-salfelder:
ok, we have a problem. the proper solution would be (imo) to disable html generation in that case. it's just not implemented. still we need other packages for the docbuild -- although the checks are missing or broken.

how about we make `--disable-doc` the default? that's roughly the original behaviour. and we need some proper way to pass "please install latex2html" (or *any* configure error) to toplevel (i.e. avoid those 3000 lines).

in3otd:
ok on `--disable-doc` as default, but this will not solve the issue we have with OS X on Travis, or?

(looking at at the `automake` docs [here](https://www.gnu.org/software/automake/manual/html_node/Unconfigured-Subdirectories.html#Unconfigured-Subdirectories) it skipping configuring `qucs-doc` yet still distribute it seems possible, but not recommended for the novice...)

in3otd:
I think the build on OS X is now failing because you (correctly) made `configure` in `qucs-doc` fail if `latex2html` was not found, in https://github.com/Qucs/qucs/pull/524/commits/ead0bdbf9a19cae143c1fa483025b24754bea2a0 . Quickest ~~fix~~ hack would be to remove the `AC_MSG_ERROR()`...

felix-salfelder:
is that true? if --disable-doc is the default, then configuration fails, but qucs-doc makefile is never used, because qucs-doc will not be in toplevel SUBDIRS. (that was the plan)

actually, AM_CONDITIONAL instead of AC_MSG_ERROR would be the correct solution. do we really need it (now)?

in3otd:
my assumption is that the `Your build exited with 1.` message in the OS X Travis was due to the failing `configure` in `qucs-doc`, but I may be wrong. Try removing the `AC_MSG_ERROR` there and see what happens...

I'm not an expert of the autotools, but as a user I'd expect that if I run a `./configure` and I'm missing a tool to properly build what `make` will build I will be notified at configure time, not at compile time with possibly some obscure error. Conversely, if I ask not to build the docs (or if that is the default) it makes no sense to notify of missing stuff that won't be needed to successfully build the configuration requested.
Again, this is likely a bit beyond the (original) scope of this PR and not really that urgent.

felix-salfelder:
> Try removing the AC_MSG_ERROR

did that. makes some sense, don't know it it's required. note that 0c7eeeb implements your suggestion.

> scope of this PR and not really that urgent.

right. i only want to see travis pass and postpone some details.

guitorri:
I don't know why PDF had individual targets. It was like that and I never bothered to change. 
It does makes sense to have top `make all`. One can always  needed `cd` and `make` into a particular directory if needed. 
I don't think I ever got to build the `html` targets. I believe only the technical manual is available in the website.

felix-salfelder:
ok thanks Guilherme. then this is finished so far as i'm concerned.

in3otd:
some random notes:
- I've built the docs in and out-of-tree, played with `distcheck` and friends and everything seems to work here.
- the others `configure.ac` in `qucs` and `qucs-core` output an empty line before the `...configured successfully.` message (see [here](https://github.com/Qucs/qucs/blob/master/qucs/configure.ac#L850-L868) and [here](https://github.com/Qucs/qucs/blob/master/qucs-core/configure.ac#L930-L932))
- [`qucs-doc/qucstex.mk`](https://github.com/Qucs/qucs/pull/524/commits/574ec9222408eae96062dbd634a594c10bf62ac4) is placed under the GPLv3 license, but the rest of the code is GPL2+. IANAL but my understanding is that this is not allowed: either we move all the code to GPLv3 or everything has to be GPLv2(+).
- where is [`MORE_TEXINFO`] (https://github.com/Qucs/qucs/pull/524/files#diff-f454448e3a434c74c2b9b5c3179c8730R28) used/needed ?
- there seems to be something fishy in [`activefilter/Makefile.am`](https://github.com/felix-salfelder/qucs/blob/07cc44af7cf5b4698b22f6e1124ca7baae5a5631/qucs-doc/report/activefilter/Makefile.am): why is the `.tex.dvi` rule [rewritten](https://github.com/felix-salfelder/qucs/blob/07cc44af7cf5b4698b22f6e1124ca7baae5a5631/qucs-doc/report/activefilter/Makefile.am#L46-L51)? What does [this line](https://github.com/felix-salfelder/qucs/blob/07cc44af7cf5b4698b22f6e1124ca7baae5a5631/qucs-doc/report/activefilter/Makefile.am#L51) do there? I will appreciate some more words about the [*insane hack*](https://github.com/felix-salfelder/qucs/blob/07cc44af7cf5b4698b22f6e1124ca7baae5a5631/qucs-doc/report/activefilter/Makefile.am#L60-L61), to help understanding why it's needed. In any case, `activefilter.pdf` is not generated correctly, it's missing the proper references and bibliography.
- not really related, but spotted while checking this PR: the author in the footnote/reference in the [MESFET report](https://github.com/Qucs/qucs/blob/master/qucs-doc/report/mesfet/content.tex#L661) should be [Yannis Tsividis](http://www.ee.columbia.edu/yannis-tsividis) (just one person, with the correct spelling, not two)

felix-salfelder:
in3otd, thanks for your thorough notes. i think i have addressed all of them.

guitorri:
I tried `make distchek` in and out-of-tree. Both give:
```
./config.guess
./config.sub
./install-sh
./missing
make: *** [distcheck] Error 1
```

felix-salfelder:
interesting. which directory?

i'm wondering how travis is getting past this...

in3otd:
uhm, strange, I did run `make distchek` in and out-of-tree here and tried again also now and it works fine.

Not sure why `distclean` complains there, indeed those files are in the `tar.gz` created by `dist` and the GNU Coding Standards say "`distclean`  *Delete all files in the current directory (or created by this makefile) that are created by configuring or building the program*". The files above were created by the `boostrap` script, running `autoconf`, before configuring and building, so I think it's right to have them in the distribution.

guitorri:
Strange indeed. I did a `git clean -xdf` and rebuild. The error is gone. Thank you for checking and sorry for the noise. Building from the tarball seems fine, parallel build (-j8) also worked here. Looks good to me.

felix-salfelder:
`\o/`

thanks for the help to both of you!

in3otd:
ok, so this should be ok to merge. @felix-salfelder, I guess/hope you were waiting for the end of the review to squash some commits...

in3otd:
BTW @guitorri , parallel build almost never works here, I usually get errors when compiling in `qucs-core/src/components/verilog` without the "fix" mentioned in #545 .

felix-salfelder:
> to squash some commits...

working on it. in3otd, shall i squash your commits as well?

guitorri:
@in3otd I was referring to qucs-doc parallel build. It is been a wile I don't try to build qucs-core.

in3otd:
> working on it. in3otd, shall i squash your commits as well?

thanks, just wanted to be sure :grin:

feel free to squash my commits also; I already marked some I wanted to squash, you can squash more if you prefer.

in3otd:
ok, @guitorri, I was hoping you found a way to cleanly build all in parallel :grin: `qucs-doc` builds fine in parallel also here.

felix-salfelder:
strange things happen on apple since the cleanup. seem completely unrelated to the changes made.

    clang: error: no such file or directory: 'componentdialog.moc.cpp'
    clang: error: no input files
    make[3]: *** [componentdialog.moc.o] Error 1
    make[2]: *** [check-recursive] Error 1
    make[1]: *** [check-recursive] Error 1
    make: *** [check-recursive] Error 1

    The command "if [ "$OSX" = true ]; then make check; else make distcheck DISTCHECK_CONFIGURE_FLAGS="${DISTCHECK_CONFIGURE_FLAGS}"; fi" exited with 2.

guitorri:
Could be a glitch on the installation of Qt for OSX. 

Look up during the configure:

```
checking for Qt OS X framework... found in /Library/Frameworks
checking for lrelease-qt4... no
checking for lrelease... no
checking for lupdate-qt4... no
checking for lupdate... no
checking for moc-qt4... no
checking for moc... no
checking for uic-qt4... no
checking for uic... no
checking for rcc... no
configure: WARNING:
  The qucs package needs the 'Qt Meta Object Compiler' to compile properly.
  Though Qt itself may be properly installed including headers and libraries
  the 'moc' program is missing.  Possibly you need to install the full
  development package of Qt.

```

I will try to re-spin Travis to see what happens.

felix-salfelder:
thanks Guilherme. haven't looked close enough...

anyway: that should be MSG_ERROR... (new PR, soon)

guitorri:
Is there something else (cleanup, squashing, rebase) you want to do here? If not, please merge. You can safely ignore the Travis OSX glitch (it is not building `qucs-doc` anyway). I tested locally and things are working fine.

felix-salfelder:
thanks Guilherme.

will merge, after another CI run, to be sure..

guitorri:
Thank you! 
Shall we remove the `TEX_ENV` branch (and `autoblacklist`) from the repo?

felix-salfelder:
sure. TEX_ENV was a mistake anyway, pushed too fast. now deleted.

Issue #521, opened 2016-06-12
octave invocation is hardcoded, needs a fix.


felix-salfelder:
see #518 for the details.

guitorri:
Fixed with PR #530.

PR #525, opened 2016-06-15
Introduce a setting to hold Octave executable path

ra3xdh:
This may be solution for #521. I have modified Octave location setting (using QSettings) to hold executable location instead of Octave directory location. This new setting is fully compatible with previous method. The logic of first Qucs start is following:
1. At first start Qucs tries to load `OctaveExecutable` setting. If is found it loads it.
2. If this setting is not found, Qucs tries to load `OctaveBinDir` setting. If it is found, `OctaveExecutable` will be set as `OctaveBinDir+octave`. 
3. If `OctaveBinDir` is not found (clean installation), `OctaveExecutable` will be initialized by default value.

Also  `OCTAVEBINDIR` environment variable and related hardcoded paths were removed and related GUI controls were corrected.

### DIFF
--- a/qucs/qucs/dialogs/qucssettingsdialog.cpp
+++ b/qucs/qucs/dialogs/qucssettingsdialog.cpp
@@ -426,7 +426,7 @@ QucsSettingsDialog::QucsSettingsDialog(QucsApp *parent)
     homeEdit->setText(QucsSettings.QucsHomeDir.canonicalPath());
     admsXmlEdit->setText(QucsSettings.AdmsXmlBinDir.canonicalPath());
     ascoEdit->setText(QucsSettings.AscoBinDir.canonicalPath());
-    octaveEdit->setText(QucsSettings.OctaveBinDir.canonicalPath());
+    octaveEdit->setText(QucsSettings.OctaveExecutable);
 
 
     resize(300, 200);
@@ -617,7 +617,7 @@ void QucsSettingsDialog::slotApply()
     /*! Update QucsSettings, tool paths */
     QucsSettings.AdmsXmlBinDir = admsXmlEdit->text();
     QucsSettings.AscoBinDir = ascoEdit->text();
-    QucsSettings.OctaveBinDir = octaveEdit->text();
+    QucsSettings.OctaveExecutable = octaveEdit->text();
 
     QucsSettings.IgnoreFutureVersion = checkLoadFromFutureVersions->isChecked();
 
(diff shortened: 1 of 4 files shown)


felix-salfelder:
`OctaveExecutable` makes some sense (as a future replacement for OctveBinDir). it should serve the same purpose as `QUCS_OCTAVE` that I proposed earlier. `QUCS_OCTAVE` should become an override, but OctaveExecutable first needs to be implemented correctly.

this

    foreach (QString dir, path_lst) {
         [..]
    }

only reimplements `PATH`. it is done by exec() which is underneath the QtProcess. please remove.

this

              if (progOctave.exists()) {
                  found = true;
                  break;
              }

does not work, unless progOctave contains the full path to octave, which might not even be known at this stage. see how qucsator is invoked, and fix the octave invocation accordingly.

see #518 and #521.

ra3xdh:
Removed file checking and `OCTAVEBINDIR` replaced by `QUCS_OCTAVE`.

felix-salfelder:
thank you. this looks much better (will compile and test later today...)

2 minor remarks
- i don't think we need *two* additional QucsSettings variables. plain `"Octave"` should be sufficiently descriptive.
- in qucssettingsdialog.cpp:914, `"Select the octave executable location"`. the "location" is optional. by omitting it, it will be resolved using `PATH`. i suggest to put `"Select the octave executable"`.

ra3xdh:
> i don't think we need two additional QucsSettings variables

Two separate fields of QucsSettings for Octave are needed to prevent overwriting of `OctaveExecutable` after `saveSettings()` is invoked. Alternative solution will be to test `QUCS_OCTAVE` variable right in `OctaveWindow::startOctave()` If it is set, tkae Octave path from `QUCS_OCTAVE`. If it's not set, get Octave path from `QucsSettings`. 

> in qucssettingsdialog.cpp:914

Yes, I will correct this caption later.

felix-salfelder:
you are right about saveSettings. i'm wondering if it is good or bad to overwrite it the way you describe. it is certainly not what one would expect...

i'd prefer a less cluttered QucsSettings, but i don't have a good idea. just leave it like that.

> Yes, I will correct this caption later. 

thanks, i can also do that just before the merge...

ra3xdh:
Caption at open dialog is corrected, all commits are squashed.

guitorri:
Ok, this patch is included in #530. Closing

Issue #526, opened 2016-06-16
Equation component: Incorrect Property name when using Enter to accept edit


in3otd:
while checking #514 I noticed a bizarre behaviour of the Equation component:
- place an Equation component on the schematic
- double click the Equation to open its Edit Properties dialog
- the selection is now on the first Line Edit (the Name of the property): type something (e.g. *Name*)
- press **Tab** to move to the second Line Edit (the Value of the property): type something (e.g. *Value*)
- press **Enter** to save the modified property...
- now the modified property shows up with name=***Value*** and Value=*Value*... the entered *Name* has disappeared. :confused: 

I'll take a look...

ra3xdh:
Fixed via #527

PR #527, opened 2016-06-17
Fix issue 526

in3otd:
- the first commit is a fix for #526: apparently a copy/paste error was made during the conversion to Qt4 in e4d02f60bfab9b7c15804467b3dae83a23baf6e6
- the other commits are just "nice to have" or cosmetic stuff, feel free to skip and/or put in `develop`
  - one is for making "Enter" always advance to the next property, also for properties whose values is selected using a ComboBox. Until now pressing Enter for these properties did nothing.
 - one for a (better?) alignment of buttons
 - last one fixes a small sync issue between a property visibility status and the corresponding checkbox

### DIFF
--- a/qucs/qucs/components/componentdialog.cpp
+++ b/qucs/qucs/components/componentdialog.cpp
@@ -34,6 +34,8 @@
 #include <QComboBox>
 #include <QGroupBox>
 #include <QPushButton>
+#include <QEvent>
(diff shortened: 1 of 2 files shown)


ra3xdh:
I have tested this solution and found no problems. This PR could be merged.

PR #529, opened 2016-06-19
Fix crash in Octave window is Page Down was pressed with empty history

in3otd:
While working on #527 I noticed that if Page Down was pressed in the Octave command window edit line before any command was entered (so the command history was empty) Qucs crashed.
This is due to the command history pointer being advanced without checking if there were actually any command in the command history.


ra3xdh:
This is trivial and could be merged.

PR #530, opened 2016-06-19
Octave fix

felix-salfelder:
turn error message into something useful. this includes the review for 3c605ce (#525).

### DIFF
--- a/qucs/qucs/octave_window.cpp
+++ b/qucs/qucs/octave_window.cpp
@@ -117,7 +117,7 @@ bool OctaveWindow::startOctave()
   octProcess.waitForStarted();
 
   if(octProcess.state()!=QProcess::Running) {
-    output->setText(tr("ERROR: Failed to execute " + Program));
+    output->setText(tr("ERROR: Failed to execute ") + Program);
     return false;
   }
 
--- a/qucs/qucs/octave_window.cpp
+++ b/qucs/qucs/octave_window.cpp
@@ -117,7 +117,7 @@ bool OctaveWindow::startOctave()
   octProcess.waitForStarted();
 
   if(octProcess.state()!=QProcess::Running) {
-    output->setText(tr("ERROR: Failed to execute ") + Program);
+    output->setText(tr("ERROR: Failed to execute \"%1\"").arg(Program));
     return false;
   }
 
--- a/qucs/qucs/octave_window.cpp
+++ b/qucs/qucs/octave_window.cpp
@@ -84,25 +84,15 @@ bool OctaveWindow::startOctave()
   if(octProcess.state()==QProcess::Running)
     return true;
 
-  QString OctavePath=QucsSettings.OctaveBinDir.canonicalPath();
-
-
   QString Program;
   QStringList Arguments;
 
-  OctavePath = QDir::toNativeSeparators(OctavePath+"/"+"octave"+QString(executableSuffix));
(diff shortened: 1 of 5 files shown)


ra3xdh:
This looks good. I also had similar plans to make this error message more informative. This PR could be merged after CI checks will be passed.

qucs/qucs/octave_window.cpp
```diff
@@ -117,7 +117,7 @@ bool OctaveWindow::startOctave()
   octProcess.waitForStarted();
 
   if(octProcess.state()!=QProcess::Running) {
-    output->setText(tr("ERROR: Failed to execute " + Program));
+    output->setText(tr("ERROR: Failed to execute ") + Program);
     return false;
```
in3otd: wouldn't be better to use [`QString::arg()`](http://doc.qt.io/qt-4.8/internationalization.html#use-qstring-arg-for-dynamic-text) so this will work well also for languages where the program name needs to be in another position (I do not have an example, but I suppose this may happen)?

felix-salfelder: certainly, you are right. i will/might fix it some day. don't hold your breath.

is this an imperfection or a bug?

in3otd: well, the user may just get a ungrammatical error message, for some languages; the Qucs GUI will still do its main purpose, of course. You can call it an issue, instead of a bug :grin:
In any case, I guess there are similar "issues" in other parts of the code around, we could as well open a dedicated issue to track that.

in3otd:
so now the user has to know which Octave to start... `octave` or `octave-cli`? It only slightly better than tell the user to rename `octave-cli` to `octave`... We should make things simpler for the user not more complex.
FYI, `octave` and `octave-cli` use a different graphic backend (at least on Linux) because the CLI version has no graphics capabilities by itself and relies on `gnuplot`, as the pre-GUI versions; I suspect the CLI version is more limited in the graphics output then. IMHO, we should do as ra3xdh did at the beginning, check the `octave` version and pass the `--no-gui` if needed.
As said in #518 this might not be needed anymore once we switch to Qt5 (...)

Not in this PR, but still about making thing simpler, all the external applications paths should be set using a "wizard" at the first Qucs startup, so it can be checked if the user choice is valid. Same wizard needs to handle the config file backward compatibility stuff, converting the old setting into the new ones and cleaning up the config. Not that I like the "wizards" a lot, but this is how things are done nowadays.

Having to set the executable name for octave but only the application directory for the other external applications in the same Application Settings tab is a bit weird and maybe slightly confusing.

felix-salfelder:
> a wizard

there is a `contrib`,  directory for things like that. maybe we should introduce `qucs/contrib` for stuff related to the gui.

anyway: this (platform configuration setup) is something a *startup script* should be doing. whether or not it invokes a wizard. a startup script should be shipped with a binary package for platform in question with the particular issues. platform dependent startup quirks could/should be kept in `contrib` as well.

the simplest startup wizard i can think of is `$EDITOR ~/.qucs/qucsrc`, we should not plough it under, just for some (temporary?)`-cli` suffix madness on only *one* platform in combination with *one* particular tool version.

> executable name for octave but only the application directory [..] confusing.

maybe the cleanup should be an issue and milestoned for something <= 0.0.20?

guitorri:
It fixes the hardcoded path. Merging.

PR #534, opened 2016-06-26
fix gperf check

felix-salfelder:
3rd argument is the error return value...

### DIFF
--- a/qucs-core/configure.ac
+++ b/qucs-core/configure.ac
@@ -577,9 +577,9 @@ fi
 unset osversion
 
 dnl Check for additional programs in PATH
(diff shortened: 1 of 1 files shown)


in3otd:
thanks; as mentioned there is a [similar problem with git] (https://github.com/Qucs/qucs/blob/master/qucs-core/configure.ac#L919) also. Is not as bad, since it just causes a `git: command not found` error message but it's confusing for the user. Could you please add a commit for that also?

in3otd:
merged in `release-0.0.19` (7da63d0327f90d700513e552565272c074c62092)

PR #535, opened 2016-06-29
Fix build failure and add more build dependencies in the README file

rubund:
Without the #include fix, the build of qucs fails on Debian which has GCC/G++  5.4.


felix-salfelder:
see #536. please close this PR, if you agree.

rubund:
looks good. wasn't aware of the 0.0.19 branch.

PR #537, opened 2016-07-09
(WIP) enable building qucs-doc out of tree

in3otd:
...expanded #524 and fixed a few things here and there; now `qucs-doc` can be built out of tree, on my machine. Marked as "WIP", since I expect it will still need some refinement to make Travis happy.


qucs-doc/report/EKV26/Makefile.am
```diff
@@ -29,6 +29,8 @@ EXTRA_DIST = $(SOURCES)
 SOURCES = EKV26.tex content.tex
 TARGETS = EKV26.pdf
 
+export TEXINPUTS=$(srcdir):$(srcdir)/..:
```
felix-salfelder: commit 641ce37e88f5c515288739e2983e71dfdb112590 already did it right, why do you deviate from it?
global variables are a bad idea, and ideas are prone to cargo cult.

and (sorry for being pedantic, but) please include my commit in your PR if you write "...expanded #524".

qucs-doc/tutorial/textmode/Makefile.am
```diff
@@ -31,7 +31,14 @@ TARGETS = textmode.pdf
 
 export TEXINPUTS=$(srcdir):$(srcdir)/..:
 
-tutorial: $(TARGETS)
+textmode.pdf: $(SOURCES)
+
+pics-tutorial:
```
felix-salfelder: probably this should be marked phony, i.e. `.PHONY:  pics-tutorial`.

NB: do you mean tutorial-recursive? iirc newer autotools (will?) support custom recursive targets.

qucs-doc/tutorial/activefilter/Makefile.am
```diff
@@ -34,12 +34,7 @@ export TEXINPUTS=$(srcdir):$(srcdir)/..:
 
 activefilter.pdf: $(SOURCES)
 
-pics-tutorial:
```
felix-salfelder: did I just comment on an unneeded target (oops). please ignore, but squash.

qucs-doc/report/EKV26/Makefile.am
```diff
@@ -44,9 +46,9 @@ SUFFIXES = .dvi .tex .pdf .ps
 	$(LATEX) -interaction=batchmode $<
 
 .tex.pdf:
```
felix-salfelder: this rule (maybe others, too) should not be states so many times. that's what `include` is for.

as you need to touch all occurences of this rule in your commit, you could as well simply put `include common_tex.mk` (or something like that)... cf my comment in 641ce37e88f5c515288739e2983e71dfdb112590.

qucs-doc/report/EKV26/Makefile.am
in3otd: I did differently because the `Makefile.am` is less cluttered in this way; no need to have a variable repeated for every invocation of `pdflatex` (and for the `.tex.dvi` rules, even if they are not used at the moment, I think).
Why exporting `TEXINPUTS` here is bad? To me it seems almost the same as you did, every recipe line invoking `LaTeX` will be executed in its own shell,and the environment will have `TEXINPUTS` set automatically instead of having to specify it explicitly.

Your commit is not in this PR because I did not actually start from it (to be pedantic, I should have written "...expanded from the idea in #524"). I can certainly add it.

qucs-doc/tutorial/textmode/Makefile.am
in3otd: yep, likely this (and others) should be `.PHONY`...

qucs-doc/report/EKV26/Makefile.am
in3otd: ah, now it's (slightly) clear what you had in mind for *deduplicate* there...

qucs-doc/report/EKV26/Makefile.am
felix-salfelder: > less cluttered in this way

no. global variables are *more* cluttered. seriously.

>  every invocation of pdflatex

the number of invocations wll reduce drastically with 'include'.

> Why exporting TEXINPUTS here is bad?

exporting is bad. what happens here is (semantically) argument passing. it's not your fault that latex does not support `-I`. but that's what it is. look at how a compiler is invoked: `$(CC) $(CCFLAGS)`, every time, over and over again.

argument passing is a *local* thing, so it should happen in the very same line that runs latex.

(hope that's more clear now...?)

qucs-doc/tutorial/textmode/Makefile.am
in3otd: forgot to comment on the `-recursive`, did not think about that, but could likely be useful. But I'm not too familiar with the auto-stuff, so this will be maybe for another PR, after fixing all the other things around...

qucs-doc/report/EKV26/Makefile.am
in3otd: yes, now is clearer. Fully agree that what we would like to do is argument passing but as you said unfortunately we can't do that.

The C compiler is invoked as `$(CC) $(CCFLAGS)` by `make`, you do not have to explicitly write a rule for this in `Makefile.am`. I'd like to write a line like `tutorial_SOURCES = tutorial.tex` (etc.) and having automake automatically  prepared the command line needed, but that's not possible.  Using the `export` avoid typing things explicitly, as said, in the same way as you do not type `$(CC) $(CCFLAGS)`  in `Makefile.am`.

> no. global variables are more cluttered. seriously.

Care to explain, seriously? There is less text that does not add much to understanding what the rule is doing, so there is less clutter in `Makefile.am`. The shell in which the command is run has the same number of environment variables in both cases. The end effect is the same.

> exporting is bad.

yes, you said that already, but did not explain why. Repeating without an explanation sounds like the cargo cult you were mentioning. I tried to explain why I think it's not bad; please explain which bad things will happen using the `export` here.

To me all this, until now, looks like a difference of style only.

felix-salfelder: > we can't do that [passing include flags to latex]

yes, we *can do* that. just unfortunately, we have to pass through the environment, because latex does not support `-I`. this is exactly what happens in the current version, except the author didn't know about that colon trick.

> having automake automatically prepared  ..  but that's not possible

anything is *possible*. using an include file certainly is, and that's practically equivalent.

> understanding what the rule is doing, 

yes. if the rule is "run program with arguments", it should mention 1) the program 2) the arguments, for my understanding.

> yes, you said that already

sorry i meant to emphasize the "exporting" in "exporting TEXINPUTS (globally)" [is bad], since this is not so much about "TEXINPUTS"...

the (hopefully better) explanation is: argument passing is *local*, so it should happen locally. a local inline export exactly implements argument passing in this [latex] case (i.e. is functionally equivalent). in contrast, a global export is 1) not visible where it should and 2) may even leak to the wrong commands.

> To me all this, until now, looks like a difference of style only.

maybe. the wiki articles on programming/coding style i found, do not mention variable scopes at all. style seems to be about indentation and whitespace, also 'naming conventions'. i think this discussion is more about semantics.

but there's some relief [here](http://c2.com/cgi/wiki?GlobalVariablesAreBad)

> For experienced programmers, there is little temptation. Using a local or the SingletonPattern is usually easy, even if it does take a few extra seconds to type. After you get burned a couple of times, you will forget the notion that globals would make things easier.

felix-salfelder:
i have merged these commits into PR #524.

in3otd:
ok, thanks - unfortunately I didn't have time to work on this PR further.
I have just pushed one last commit which I left out earlier, as not really needed for fixing the docs building: currently `make install` also compiles the code, if needed, while it does not generate the documentation. I have aligned the behavior so now `make install` generates the docs also, if needed.

Issue #538, opened 2016-07-11
"File -> Example" does nothing


pidgeon777:
I suppose that by clicking on:

File -> Examples

it would open a window to browse the examples provided within the software package. Unfortunately, nothing happens by doing that.

![image](https://cloud.githubusercontent.com/assets/13739892/16742311/6ba25e30-47a7-11e6-9987-1a650a7df9d2.png)

I'm using Qucs v0.0.19 (qucs-win64-bin portable version), running it on a Windows 7 x64 machine.

in3otd:
did you have previously another (older) version of Qucs installed?
If it was not properly uninstalled, you might still have a spurious `QUCSDIR` environment variable pointing to the old installation directory; if you open a command window and type `echo %QUCSDIR%` do you see a path to the old Qucs ?

pidgeon777:
I never installed Qucs previously. I ran the command you suggested, and here is the result:

![image](https://cloud.githubusercontent.com/assets/13739892/16750904/6b4f9dfc-47d5-11e6-94fb-6e94be256c42.png)

andresmmera:
I can reproduce this bug in Fedora 24. It seems the examples folder is missing:

> vfs-open:

 /usr/local/share/qucs/docs/examples/: error opening location: Error when getting information for file '/usr/local/share/qucs/docs/examples': No such file or directory

[andresmmera@localhost qucs]$ ls /usr/local/share/qucs/docs/
cs  de  en  es  fr  pt  ru  uk

Then, I went to qucs/qucs/examples folder and I did:
./autogen.sh
./configure
sudo make install

At this point, the examples folder was created at the aforementioned folder, but... the thing is that there's only a text file (README) which contains the following:

>  This distribution contains no examples, please copy them manually from the qucs git repository:
git clone git://git.code.sf.net/p/qucs/git qucs

felix-salfelder:
there are examples in `$toplevel/examples` plus a `Makefile.am` with my name in it. maybe i once tried?

there are compressed files involved. we should unpack them first. also, i can't find any license statements.

in3otd:
yep, this has nothing to do with `QUCSDIR`, at first I thought it was due to the path problems we saw recently...

The `examples` directory is "simply" missing from the binary distribution; apparently the examples were usually "manually" added by the [release script] (https://github.com/Qucs/qucs/blob/master/contrib/qucs-release.sh).

pidgeon777:
[...]

PR #544, opened 2016-07-12
install examples to examplesdir=${pkgdatadir}/examples

felix-salfelder:
maybe #538 needs yet another fix.

### DIFF
--- a/examples/Makefile.am
+++ b/examples/Makefile.am
@@ -22,8 +22,8 @@
 # Boston, MA 02110-1301, USA.
 #
 
-pkgdocsdir= ${docdir}
-dist_pkgdocs_DATA= \
+examplesdir = ${pkgdatadir}/examples
+dist_examples_DATA = \
     active_bp.sch\
     active_lp.sch\
     active_mixer.sch\
--- a/qucs/qucs/main.cpp
+++ b/qucs/qucs/main.cpp
@@ -678,7 +678,7 @@ int main(int argc, char *argv[])
 	  QucsSettings.LibDir =      QucsDir.canonicalPath() + "/share/qucs/library/";
   }
   QucsSettings.OctaveDir =   QucsDir.canonicalPath() + "/share/qucs/octave/";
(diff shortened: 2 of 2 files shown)


in3otd:
also the [hardcoded path](https://github.com/Qucs/qucs/blob/master/qucs/qucs/main.cpp#L681) (...) in `main.cpp` should be changed accordingly

in3otd:
I have just noticed that the recently added example `filter_optimization.sch` is not listed in the `Makefile.am`, I forgot to add it there also...

Should also the compressed examples be copied, so at least the user sees they are there? There is a subdirectory also, `external_interface/` which is currently not copied.

felix-salfelder:
> not listed in the Makefile.am

generally wildcarding is a bad idea (and discouraged). this is a place where it might not hurt much.
i think it can/should be implemented using `install-local` and `dist-local` targets (or something like that).

    install-local:
        $(INSTALL) -d $(DESTDIR)/$(pkgexamplesdir)
        $(INSTALL) *.sch $(DESTDIR)/$(pkgexamplesdir)

(did not try or check.)

> he compressed examples be copied, 

the compressed examples are really bad. there should be no binary files in the repo or in a tarball.

now we could uncompress them, which will inflate a working copy. to avoid that, we could move the (big?) examples to a seperate `qucs-gui-examples` repository.

if we cannot reproduce copyright/license information of the example files (does anyone care to waste time?), a seperate repo will come in handy...

guitorri:
I will move the compressed examples back to the website: http://qucs.sourceforge.net/examples.html

Most of them are already there.
At some point I think we should make a selection of representative things from the `qucs-test` repository and use as examples. `qucs-test` already contains majority of the examples.

Indeed, license information is a big question. In most cases there is no track record of authorship.

guitorri:
Ok. I moved the compressed examples to the website. The last 4 in *Complete projects* where missing in the website. I will later send another PR to remove all compressed examples, and move the examples directory inside `qucs/`.

Issue #541, opened 2016-07-12
"Move Up/Down" buttons greyed out in the equation property window


pidgeon777:
I created a list of equations:

![image](https://cloud.githubusercontent.com/assets/13739892/16751280/664685b6-47d8-11e6-8787-8d5eea335c2e.png)

It is impossible to modify the order of the displayed equations, because the "Move Up/Down" buttons are greyed out in the equation property window:

![image](https://cloud.githubusercontent.com/assets/13739892/16751249/2aed8e92-47d8-11e6-906e-08c4234b59e2.png)

I'm using Qucs v0.0.19 (qucs-win64-bin portable version), running it on a Windows 7 x64 machine.

in3otd:
Thanks for finding this bug! If you switch the application language to English then it will work as expected ...

The problem comes from [here](https://github.com/Qucs/qucs/blob/master/qucs/qucs/components/componentdialog.cpp#L588), the comparison should be with `tr("equation")` (But I think a translator could mess up this anyway, proposing a different translation here and in the Equation component description, or am I wrong?).

pidgeon777:
Probably it's as you're suggesting, maybe the translator has some effect even on the description of the equation... after the translation, the test in the code won't work anymore ("equation" get translated to "equazione" when using Italian language).

guitorri:
Fixed with #548.

Issue #543, opened 2016-07-12
Component search selects wrong MOS component


in3otd:
Steps to reproduce (in `release-0.0.19` and older):
- select the Components tab in the Main Dock on the left side
- type `MOS` in the search box at the bottom: eight MOS symbols will appear
- select the first symbol (`n-MOSFET`) and place it on the schematic
- note that the symbol placed is not the one for the standard `n-MOSFET` just selected but the one for the `n-MOSFET with substrate`. (the two have the same description in the Components tab, maybe not a coincidence...)

(similar issue exists with the `p-MOSFET`)

guitorri:
Fixed by #549

PR #383, opened 2015-09-26
cleanup Component::save and ::load

felix-salfelder:
this is a no-op in preparation of
- implement different schematic formats (verilog, XML, qucsch)
- implement symbol attributes properly

PR #547, opened 2016-07-23
Fix for #541 and #543

in3otd:
Fix for #541 and #543 .

- for #541: Up/Down buttons in the Component Properties dialog for reordering the equations in the Equation component are enabled looking at the component Model field and not at its description

- for #543 : now every component icon in the Components tab has associated information about the the component it represents, so this can be retrieved directly when the icon is selected. Previously the association was based only on the selected category and icon position in the IconView tab. This allows to show in the IconView icons of components belonging to different categories and easily retrieve the associated component.

### DIFF
--- a/qucs/qucs/components/componentdialog.cpp
+++ b/qucs/qucs/components/componentdialog.cpp
@@ -602,7 +602,8 @@ void ComponentDialog::slotSelectProperty(QTableWidgetItem *item)
     ButtAdd->setEnabled(true);
     ButtRem->setEnabled(true);
 
-    if (Comp->Description == "equation") {
+    // enable Up/Down buttons only for the Equation component
+    if (Comp->Model == "Eqn") {
       ButtUp->setEnabled(true);
       ButtDown->setEnabled(true);
     }
--- a/qucs/qucs/qucs.cpp
+++ b/qucs/qucs/qucs.cpp
@@ -621,7 +621,6 @@ void QucsApp::slotSetCompView (int index)
   editText->setHidden (true); // disable text edit of component property
 
   QList<Module *> Comps;
-  CompComps->clear ();   // clear the IconView
   if (CompChoose->count () <= 0) return;
 
   // was in "search mode" ?
@@ -629,9 +628,10 @@ void QucsApp::slotSetCompView (int index)
     if (index == 0) // user selected "Search results" item
       return;
     CompChoose->removeItem(0);
-    CompSearch->clear();
+    CompSearch->clear(); // clear the search box
     --index; // adjust requested index since item 0 was removed
   }
+  CompComps->clear ();   // clear the IconView
 
   // make sure the right index is selected
   //  (might have been called by a cleared search and not by user action)
--- a/qucs/qucs/qucs.cpp
+++ b/qucs/qucs/qucs.cpp
@@ -636,15 +636,20 @@ void QucsApp::slotSetCompView (int index)
   // make sure the right index is selected
   //  (might have been called by a cleared search and not by user action)
   CompChoose->setCurrentIndex(index);
+  int catIdx = index;
+  int compIdx;
+  iconCompInfoStruct iconCompInfo;
+  QVariant v;
   QString item = CompChoose->itemText (index);
 
   Comps = Category::getModules(item);
   QString Name;
+  pInfoFunc Infos = 0;
 
   // if something was registered dynamicaly, get and draw icons into dock
   if (item == QObject::tr("verilog-a user devices")) {
 
-    QListWidgetItem *icon;
+    compIdx = 0;
     QMapIterator<QString, QString> i(Module::vaComponents);
     while (i.hasNext()) {
       i.next();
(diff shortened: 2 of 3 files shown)


qucs/qucs/components/componentdialog.cpp
```diff
@@ -602,7 +602,8 @@ void ComponentDialog::slotSelectProperty(QTableWidgetItem *item)
     ButtAdd->setEnabled(true);
     ButtRem->setEnabled(true);
 
-    if (Comp->Description == "equation") {
+    // enable Up/Down buttons only for the Equation component
+    if (Comp->Model == "Eqn") {
```
felix-salfelder: please remove that. componentdialog must be cleaned from component code. this does not improve the situation.

felix-salfelder:
please merge  #383 first.

qucs/qucs/components/componentdialog.cpp
in3otd: please explain how this "does not improve the situation". The purpose of this change is to fix a bug in the current code. If you think it doesn't not actually fix it, please propose a better/different way or point out why it does not fix it.

in3otd:
yep, there are lot of things that should be merged in `develop` but this is targeted to `release-0.0.19`. What I would like to do is fix the open bugs in `release-0.0.19`, do a release and then go on with `develop`.

felix-salfelder:
feel free to merge #383 into the `release-*`.

this PR does not look like a bugfix, more like a major rearrangement. if it needs to be rearranged, clean up *first*. that's what #383 is about.

qucs/qucs/components/componentdialog.cpp
felix-salfelder: sure. first of all, move the equation code out of componentdialog. it's clearly in the wrong place, regardless of what it does. a better way would be to have the equation code in the equation "component".

the `Description` and `Model` strings are a misconception, related to the cause of this bug. switching one for the other is not an improvement.

if fixing (including reviewing the fix) this bug is harder than cleaning up, then certainly a better way is to clean up.

just put `Comp->doSomethingWithDialog(this);`  instead. choose a proper name.

in3otd:
#383 was targeted to `develop` and tagged `0.0.20`, why it should now go to `release-*`?
This PR does not touch any file modified in #383. If you have a simpler way to fix #541 and #543, please propose it. I'm not trying to clean up anything here, just to make the current code working as expected.

felix-salfelder:
> If you have a simpler way to fix

yes i do. see the inline comments.

> I'm not trying to clean up anything

exactly. but you should. and you should review and merge cleanup PRs, such as mine.

the continuation of #383 (cleanup) will conflict with the Model and Description string madness. it will replace it anyway. i never said that #383 was not meant for immediate merge. 0.0.19 was scheduled for... april last year?

in3otd:
> >   If you have a simpler way to fix

> yes i do. see the inline comments.

which inline comment do you mean? "merge #383" ? Could you please explain how merging #383 fixes #541?

> the continuation of #383 (cleanup) will conflict with the Model and Description string madness.

I can see that, even if "the continuation of #383" has not been proposed yet; at that point not only the single line modified in e24f2dd102bd4ac4b288a2e8f27dea37edf67fbd will have to change but much more. So what's the point of merging #383 before this PR? Lot of code will need to change anyway.

> I'm not trying to clean up anything

>> exactly. but you should. and you should review and merge cleanup PRs, such as mine.

I think I merged very little in develop, until now and I do not plan to merge much there until `release-0.0.19` is closed. Since my spare time is limited, I tend to merge first things that are urgent from the user point of view (bugfixes) and which I can easily see what they are doing and were they are going. Transitional refactoring (or "cleanup") PRs targeted to `develop` do not fall in this category.

felix-salfelder:
> not only the single line modified in  #383

the madness lies deeper. #383 adds
`QString Model = c->Model; // BUG: don't use names`
the intent is to reduce the abuse of those strings. your PR adds even more. (just dont.)

> hich inline comment do you mean? "merge #383" ? Could

github calls them "line notes". sorry about that.
the line note says "just put Comp->doSomethingWithDialog(this); instead. choose a proper name.".

clearer now?

qucs/qucs/qucs.cpp
```diff
@@ -834,44 +860,37 @@ void QucsApp::slotSelectComponent(QListWidgetItem *item)
...
-    //comp search is not empty, search all category to get a component in same name
-    QStringList cats = Category::getCategories();
-    int i = 0;
-    //search normal component
-    foreach (QString it, cats) {
-      Comps = Category::getModules (it);
-      foreach(Module *mod, Comps) {
-        if (mod->info) {
-          (*mod->info)(CompName, CompFile_cptr, false);
-          if (CompName == name) {
```
felix-salfelder: sorry, your patch is too big. i do not understand.

this seems to be the offending line. (right?). it looks like a deep linear search for something which is identified by a string (ugh)

what does it actually try to do and why? (and in particular: what is the context?)

what in your code replaces this search? (or is the search not necessary at all?)

qucs/qucs/qucs.cpp
```diff
@@ -834,44 +860,37 @@ void QucsApp::slotSelectComponent(QListWidgetItem *item)
...
+    QVariant v = CompComps->item(i)->data(Qt::UserRole);
+    iconCompInfoStruct iconCompInfo = v.value<iconCompInfoStruct>();
+    qDebug() << "variant :" << v << iconCompInfo.catIdx << iconCompInfo.compIdx;
+
+    Category* cat = Category::Categories.at(iconCompInfo.catIdx);
+    Module *mod = cat->Content.at(iconCompInfo.compIdx);
+    qDebug() << "mod->info" << mod->info;
+    qDebug() << "mod->infoVA" << mod->infoVA;
+    Infos = mod->info;
+    if (Infos) {
```
felix-salfelder: the subsequent block seems to infer `view->selElem` from `mod->info`

i don't understand how this is meant to work, but it does not seem to belong here at all.
can't you just put `mod->getElemInfo(view->selElem)` and implement `mod::getElementInfo` accordingly?

in3otd:
> clearer now?

oh, clearer that I missed the "inline comment"... I read just the latest comment at the end, while the "inline comments" grow "in place", so I missed it.

So, this PR contains fixes for two issues. This may be a source of confusion and I can split, if needed.
#541 is fixed by the first commit e24f2dd102bd4ac4b288a2e8f27dea37edf67fbd . I do not claim it to be a long-term solution, just a fix that works now, with the current code.

#543 is fixed by the other commits; I agree that d131d6923947bd91ba0d6589fb7b25152a43059c and 75fdea5e1efdabe28c4adbcd61d18a13f3e146ab are quit big, but I did not find a clean way to split them into smaller, self-contained, changes.

qucs/qucs/qucs.cpp
in3otd: yes, the problem was trying to identify a component via its description. This does not work when two components have the same description, like "n-MOSFET" for the 3-terminal MOSFET and the 3-terminal plus substrate connection MOSFET.
The context it the component search, which is activated by typing a string in the line input at the bottom of the Components tab on the left side of the GUI (see #543 description also)

Until now the icons in the Components tab did not directly identify a component; when a component icon was clicked the corresponding component was identified indirectly, looking at the currently selected component category in the Combo Box at the top and the icon index in the IconView pane.

What I do here is to put these information in the Icon data, so there is no need to search which component corresponds to the selected icon, now the icon itself holds these data and the (wrong) search for which component has the same description as the icon is no longer needed.

qucs/qucs/qucs.cpp
in3otd: I do not claim to fully understand it either :grin:

`mod->info` with the last parameter set to `true` instantiates a component, so it can be attached to the cursor and placed on the schematic. (`mod->infoVA` does the same for the Verilog-A components, no idea why it's handled differently)

felix-salfelder:
seems you rebased.

this kills all the line notes. should only rebase after the review is done.

yes, please split into different PRs and let's start from scratch.

in3otd:
yep, I rebased by mistake...

I'll split and open separate PRs

PR #548, opened 2016-07-24
Fix for #541 - use Model field to identify the Equation component

in3otd:
The Up/Down buttons in the Component Properties dialog for reordering the equations in the Equation component are now enabled looking at the component Model field and not at its textual description, so it works independently of the selected language for the GUI.

As said elsewhere, it certainly not future proof and there are currently a lot of other places where the component Model field is used to identify a particular component.

### DIFF
--- a/qucs/qucs/components/componentdialog.cpp
+++ b/qucs/qucs/components/componentdialog.cpp
@@ -602,7 +602,8 @@ void ComponentDialog::slotSelectProperty(QTableWidgetItem *item)
     ButtAdd->setEnabled(true);
     ButtRem->setEnabled(true);
 
-    if (Comp->Description == "equation") {
+    // enable Up/Down buttons only for the Equation component
+    if (Comp->Model == "Eqn") {
       ButtUp->setEnabled(true);
       ButtDown->setEnabled(true);
     }


felix-salfelder:
thanks for the seperate PR.

we had a lengthy discussion already. let me summarize
- the reason for the problem is the string comparison madness
- `Module` has been staged for removal, and that's work in progress
- i strongly suggest to merge that other PR first
- i suggested a solution in some line note of an outdated commit of yours.

mainly, putting Module here instead of Description is rearranging the deck chairs on the titanic. it does not work. if this approached worked, we would not talk about a fix, right?

in3otd:
yes, we had a lengthy discussion already, but still seems to me not much relevant to this PR.

My summary is:
- there is a bug (#541) in `release-0.0.19`
- this PR fixes that bug
- it uses `Module` one time more; if and when `Module` will disappear this also will need to be modified
- even if I merged #383 the fix for #541 will not change. If and when the following of #383 will be available (which AFAIK is not, yet) this will be implemented differently, together with a lot more code.
- removing `Description` and `Model` from the components in `release-0.0.19` seems a big change to me and certainly I do not know how to do that properly anyway.

Even shorter summary:
- this fix works now and does not make future refactorings any harder.

felix-salfelder:
> removing Description and Model from the components in release-0.0.19 seems a big change to me and certainly I do not know how to do that properly anyway.

we don't need to remove Model entitirely just to get it right in one place. and that was not what i was proposing.

> this fix works now and does not make future refactorings any harder.

yes it does. how much refactoring have you done?

in3otd:
> yes it does. how much refactoring have you done?

nothing related to `Description` and `Model` for sure. But what has the question to do with this PR making the refactoring harder?
The line I changed was intended to identify an `Equation` component, but wrongly used the `Description` field to do so. The refactoring will have to handle this case anyway in a different way, even without this PR. Unless if the refactoring is blindly done to reflect the current code and will not address fixing the bug.

You seems to have a better way to solve #541, great, just make a PR.

felix-salfelder:
> But what has the question to do with this PR making the refactoring harder?

fixing `Model` abuse means removing all occurences of `Model`. pl^H^Hfeel free to join in.

> You seems to have a better way to solve #541, great, just make a PR.

just backported the fix to #383. this PR also cleans up related problems.

guitorri:
One more line to change during refactor... Merging as it fixes the issue in question.

PR #549, opened 2016-07-24
Issue 543 fix

in3otd:
Fix for #543. 
Now every component icon in the Components tab has associated information about the the component it represents, so the component can be retrieved directly when the icon is selected. 

Previously the association was based only on the selected category and icon position in the IconView tab. If the icon position did not correspond to the component position in its category (as it was for the "Search Results" category) it was not possible to unequivocally retrieve the component corresponding to a particular icon. The components corresponding to the icons which were put in the "Search Results" view were identified only looking at their description and this did work with components having the same description (e.g. `n-MOSFET` is used as description for bot the 3-terminal MOS and the 4-terminal (with substrate) one.

The last 3 commits are small fixes for the "component search" functions, while I was at it.

### DIFF
--- a/qucs/qucs/qucs.cpp
+++ b/qucs/qucs/qucs.cpp
@@ -621,7 +621,6 @@ void QucsApp::slotSetCompView (int index)
   editText->setHidden (true); // disable text edit of component property
 
   QList<Module *> Comps;
-  CompComps->clear ();   // clear the IconView
   if (CompChoose->count () <= 0) return;
 
   // was in "search mode" ?
@@ -629,9 +628,10 @@ void QucsApp::slotSetCompView (int index)
     if (index == 0) // user selected "Search results" item
       return;
     CompChoose->removeItem(0);
-    CompSearch->clear();
+    CompSearch->clear(); // clear the search box
     --index; // adjust requested index since item 0 was removed
   }
+  CompComps->clear ();   // clear the IconView
 
   // make sure the right index is selected
   //  (might have been called by a cleared search and not by user action)
--- a/qucs/qucs/qucs.cpp
+++ b/qucs/qucs/qucs.cpp
@@ -636,15 +636,20 @@ void QucsApp::slotSetCompView (int index)
   // make sure the right index is selected
   //  (might have been called by a cleared search and not by user action)
   CompChoose->setCurrentIndex(index);
+  int catIdx = index;
+  int compIdx;
+  iconCompInfoStruct iconCompInfo;
+  QVariant v;
   QString item = CompChoose->itemText (index);
 
   Comps = Category::getModules(item);
   QString Name;
+  pInfoFunc Infos = 0;
 
   // if something was registered dynamicaly, get and draw icons into dock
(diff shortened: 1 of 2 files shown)


qucs/qucs/qucs.cpp
```diff
@@ -834,44 +860,37 @@ void QucsApp::slotSelectComponent(QListWidgetItem *item)
...
+    // in "search mode"
+    QVariant v = CompComps->item(i)->data(Qt::UserRole);
+    iconCompInfoStruct iconCompInfo = v.value<iconCompInfoStruct>();
+    qDebug() << "variant :" << v << iconCompInfo.catIdx << iconCompInfo.compIdx;
+
+    Category* cat = Category::Categories.at(iconCompInfo.catIdx);
+    Module *mod = cat->Content.at(iconCompInfo.compIdx);
+    qDebug() << "mod->info" << mod->info;
+    qDebug() << "mod->infoVA" << mod->infoVA;
+    Infos = mod->info;
```
felix-salfelder: see my line note on the same line in that outdated commit (how to find it?)

essentially, I suggested to not do that kind of quirk here, but in the component (module? component symbol?) instead.

in3otd: weird, GitHub still says you *commented on an outdated diff* but this time I did not rebase :wink: . I also used a different branch from the one used previously to avoid these issues... go figure!

I think you were referring to [this](https://github.com/Qucs/qucs/pull/547#discussion_r71971661). If this means modifying [module.cpp](https://github.com/Qucs/qucs/blob/release-0.0.19/qucs-core/src/module.cpp) I'm a bit reluctant to do that as I do not really understand its inner workings. That part I practically reused from the existing code.

felix-salfelder: (note to self: never again start line-comment threads...)

no need to modify anything. you just need to add the function you need, will not break any of the existing ones.

probably you will not even have to touch module.cpp. the base class implementation should be trivial (no-op), to you can safely put it in the header.

andresmmera:
I've just checked this PR: It works nicely as expected :+1: 
However, I've also noticed that the *selection mode* is turned off after using the searchbox. 

In my opinion, it would be great if the user could have access to the component properties without clicking in the arrow icon after placing the component. In order to activate the *selection tool* you could add:   
select->setChecked(true);

at the end of QucsApp::slotSelectComponent() (Qucs.cpp, line 905)

Hope it helps :)

screenshot from 2016-08-16 09-24-56

felix-salfelder:
please move the new code to where it belongs. some of the dialogue got lost due to the rebasing, but i bet it's still out there.

guitorri:
I cannot follow the conversation of what belongs to where. The code works here. Merging.

Andres, I am not sure I follow you. If you pick form the Components Tab, using search or not, the component stays selected (so you can place it multiple times). Only when you unselect (Esc or menu Edit or Toolbar) you enter selection mode. You want to clear the selection after each insertion?

andresmmera:
@guitorri Ok, I didn't realize that pressing Esc would do the job. Please ignore my message above...

PR #555, opened 2016-08-15
[WIP] Qucs Wideband Matching tool

andresmmera:
I made a complete mess with the git command... This time I doing the PR to the _develop_ branch. Sorry for the noise :S

Hello all, 

This PR is a tool for designing wideband matching networks between
complex, frequency dependent impedances. It is based on the grid search
algorithm described in [1], however the implementation differs from the
original where it was found convenient.

The workflow is the following:

1    The user enters the source/load impedance data and the frequency
band where a good matching condition is required. Whenever the
impedance depends on the frequency, it must be specified using a s1p
file.

2 Search mode selection: The tool comes with a set of default circuit
candidates for the best matching network, but the user can also specify
custom topologies (L/C/TL) using a text file. The response of the
default topologies was analytically calculated beforehand, so the
execution is faster than when using an arbitrary network.

3    For every candidate topology:
    3.1 The program sets a suitable starting point depending on the
impedance of the components at the mean frequency.
    3.2 Grid search. A hypercube is build over the current pivot and
then every vertex is evaluated. At the end of each iteration, the
hypercube is moved to the best vertex or, otherwise, it is
shrunk/expanded. When the stop condition is met, a ‘rough’ local optimum
should be achieved.

4    Those topologies which gave a good matching condition (say <10dB)
after the grid search are chosen to refine their results with a local
optimiser. At some point, I have linked Nlopt library and tested a few
derivative free optimizers such as Subplex, Praxis and Nelder-Mead. The
Subplex method claims to outperform NM when the dimension of the problem
is high, but, in general, a matching network should be kept as simple as
possible (so typically, dimension < 6). On the other hand, the Praxis
algorithm, as a variant of the Powell’s method, relies on the assumption
that the objective function follows a quadratic expression... but I am
not pretty sure that the matching network design problem meets this
condition. Finally, the Nelder-Mead algorithm is dead simple to
implement and the results seem to be good enough… so it was the one I
have chosen.

5  At the end of the program, the best matching network is copied to
clipboard in the same way as other Qucs tools do.

In order to verify the tool is working as expected, the user/developer
can execute a command line testbench (qucs-wideband-
matching/testbench/testbench)

For sure there is plenty of room for improvement, so comments are
welcome :)

[1] Broadband direct-coupled and RF matching networks. Thomas R.
Cuthbert, 1999
[2] Convergence properties of the Nelder-Mead simplex method in low
dimensions. J.F. Lagarias, J.A. Reeds. SIAM J. OPTIM, Vol 9, No. 1, pp.
112-147

selection_010
screenshot from 2016-08-14 16-39-00
selection_006


.gitignore
```diff
@@ -203,12 +203,61 @@ Debug
...
+/qucs-core/src/converter/parse_vcd.hpp
+/qucs-core/src/parse_citi.hpp
+/qucs-core/src/parse_csv.hpp
+/qucs-core/src/parse_dataset.hpp
+/qucs-core/src/parse_mdl.hpp
+/qucs-core/src/parse_netlist.hpp
+/qucs-core/src/parse_touchstone.hpp
+/qucs-core/src/parse_zvr.hpp
+/qucs/ar-lib
+/qucs/qucs-wideband-matching/CMakeCache.txt
```
guitorri: I am not sure these CMake files should be ignored as such.
If you do out-of-tree build they should not bother you.

qucs/qucs-wideband-matching/Makefile.am
```diff
@@ -0,0 +1,53 @@
+## Process this file with automake to produce Makefile.in
+#
+# qucs-widebandmatching/Makefile.am
+#
+# Automake input file.
+#
+# Copyright (C) 2006 Stefan Jahn <stefan@lkcc.org>
```
guitorri: Really?

andresmmera: lol, I have to update that stuff...

andresmmera:
I've just rebased and pushed two commits (some UI fixes and permissions stuff) 
selection_003

I've realized that this needs some extra work... I mean, it does the job but it may lead to a frustrating user experience... I reckon that it'd be better to leave this PR at the bottom of the list #651 or even discard it for 0.0.20 release (Notice that the milestone here is set to 0.0.21).

If possible, I'll try to tweak the code a little bit so as to support 2-port matching...

andresmmera:
Added the two-port matching feature

Here's a demo screencast:
https://drive.google.com/open?id=0BwDjrwGfd_z7NFcxOWJUaHdGTXc
I've removed 20 seconds of the video because during that time the UI is doing nothing...

roger-:
Any chance of this being merged soon?

andresmmera:
It's still WIP. I haven't worked on this for a while...

kmihaylov:
I'm interested in this tool. Since it is not yet merged, any instructions on how to compile myself? Which repo to fetch and etc?

I have to match a wideband (20 MHz wide) FM antenna with impedance varying a few hundred ohms. I guess that this tool may help?

andresmmera:
First you need to fetch this branch:
git checkout -b andresmmera-WidebandMatching_new develop
git pull https://github.com/andresmmera/qucs.git WidebandMatching_new

Then, compile Qucs as usual (see qucs README.md if needed):
    cd qucs
    ./bootstrap
    ./configure
    make
    sudo make install

Hope this tool fits your needs. I haven't worked on this since quite some time ago.

kmihaylov:
Thank you. I had to do some additional steps. Here is the sequence of commands I used.

I had to manually install ADMS first, so:
```
sudo apt-get install build-essential
sudo apt-get install automake libtool gperf flex bison
sudo apt-get install libxml2 libxml2-dev
sudo apt-get install libxml-libxml-perl
sudo apt-get install libgd-perl
git clone https://github.com/Qucs/ADMS.git
cd ADMS
./bootstrap.sh
./configure
sudo make install
sudo ldconfig
```

QUCS, since there was no package octave-epstk for ubuntu 18, I configured it with ```--disable-do```
```
sudo apt-get install build-essential
sudo apt-get install libqt4-dev libqt4-qt3support
sudo apt-get install automake libtool gperf flex bison
sudo apt-get install texlive-font-utils octave-epstk

git clone https://github.com/andresmmera/qucs.git
cd qucs
git submodule init
git submodule update
git checkout WidebandMatching_new
./bootstrap
./configure --disable-doc
make
sudo make install
```

Unfortunately the videos from your previous posts is missing. It would be nice if you upload it to youtube. Also the sample s1p files. I just installed QUCS and they will be of great help.

Thanks!

hoangkimnhut-itr:
where can i download the tool

andresmmera:
You would need to compile the source code. I didn't work on this again. 

If you want to design a broadband matching network you'd better use the Smith Chart. Try to make the Q circles as narrow as possible. Then put the network in a simulator and tune/optimize it.

Check out this tool:

https://www.rfdude.com/downloads

PR #561, opened 2016-09-11
[WIP] Use Appveyor to build 32 and 64 bit packages

in3otd:
just testing - trying to use Appveyor to build 32 and 64 bit binary packages

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,28 +1,36 @@
+environment:
+  matrix:
+    - MSYSTEM : MINGW64
+      MBITS: 64
+      MARCH: x86_64
+    - MSYSTEM : MINGW32
+      MBITS: 32
+      MARCH: i686
 
 platform:
   - x64
 
-cache:
-  - C:\msys64\var\cache\pacman\pkg\mingw-w64-x86_64-qt4* -> .appveyor.yml
+#cache:
+#  - C:\msys64\var\cache\pacman\pkg\mingw-w64-x86_64-qt4* -> .appveyor.yml
 
 install:
 
   # Take a look at the environment
   - set
 
-  - set "PATH=C:\msys64\usr\bin;C:\msys64\mingw64\bin;%PATH%"
+  - set "PATH=C:\msys64\usr\bin;C:\msys64\mingw%MBITS%\bin;%PATH%"
 
-  - C:\msys64\usr\bin\bash -lc ""
-  - C:\msys64\usr\bin\bash -lc "pacman --version"
-  - C:\msys64\usr\bin\bash -lc "pacman -Q"
+  - bash -lc ""
+  - bash -lc "pacman --version"
+  - bash -lc "pacman -Q"
   # Switch from SF to msys2.org (default, much faster)
-  - C:\msys64\usr\bin\bash -lc "pacman --noconfirm --sync pacman-mirrors"
-  - C:\msys64\usr\bin\bash -lc "pacman --noconfirm -S autoconf automake bison flex"
-  - C:\msys64\usr\bin\bash -lc "pacman --noconfirm -S mingw-w64-x86_64-qt4"
-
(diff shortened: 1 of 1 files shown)


in3otd:
the above Appveyor setup builds two Qucs binary packages for Windows:
- a 64-bit version: [qucs-win64.zip](https://ci.appveyor.com/api/projects/qucs/qucs/artifacts/qucs-win64.zip?job=Environment%3A%20MSYSTEM%3DMINGW64%2C%20MBITS%3D64%2C%20MARCH%3Dx86_64)
- a 32-bit version: 
[qucs-win32.zip](https://ci.appveyor.com/api/projects/qucs/qucs/artifacts/qucs-win32.zip?job=Environment%3A%20MSYSTEM%3DMINGW32%2C%20MBITS%3D32%2C%20MARCH%3DI686)

the above links point automatically to the latest successful packages built.

I have tried both versions on two machines where Qucs has never been used before and both were able to run without issues.

felix-salfelder:
looks sane, please merge as it suits.

guitorri:
Nice. Is it the intention to use these binaries also for distribution? If so, it might make sense to tag it after the branch|git hash|date. I guess Appveyour will build the binary for any branch, right?

in3otd:
yes, the intention is to use those binaries for distribution.
It works, but it's not yet really ready to be used: as you said the packages should likely contain the corresponding git hash or tag, should maybe built only for tagged commits or maybe for any commit in `release-*`, should include the GPL license text (and maybe also the `zlib` one). Need also to find a way to include the PDF docs... either retrieved from somewhere or built in AppVeyor (but this latter may be not so easy).

guitorri:
Somehow the compiler could be indicated (in all platforms actually). Either in the filename or the 'About' dialog  or in the `qucsator -h` pages. The compiler version is only relevant for loading Verilog-A modules, but still... 

Right setting up LaTex on appVeyor might not be easy. It is easier to pull the PDF files from somewhere.  If we do one commit to tag a release, and another to build the release package, we might be able to pull the release version of qucs-doc into the package. (Can we keep artifacts with GitHub?)

Can it keep multiple artifacts? 
You are building a portable (which is good). 
If we want the installer, we have to [run InnoSetup](https://github.com/Qucs/qucs/blob/release-0.0.19/contrib/windows/qucs-installer.sh) to package the thing into an executable.

guitorri:
I will merge this now, later we see how to include what is missing.
The link seems to be broken, but the artifacts can be found if one browses to [AppVeyor] page(https://ci.appveyor.com/project/qucs/qucs)

in3otd:
FYI, the links appear to work again now;
I guess they were not working since they are pointing to the ["latest successful artifact from all branches"](https://github.com/appveyor/ci/issues/403#issuecomment-180026470) and until this PR was merged the other PRs did not have any artifact defined, but I'm just guessing...

guitorri:
Makes sense, but the link would them expose the latest artifact? What happens when this patch gets applied to other branches? I am also guessing...

in3otd:
my understanding is that those links really give the latest available artifacts, from any branch. The linked discussion above shows also how to have links for artifacts from a particular branch only, but did not try to see how that actually works.

guitorri:
We can get anything we need:
https://www.appveyor.com/docs/packaging-artifacts/#permalink-to-the-last-successful-build-artifact

Just tried this to narrow it down to one branch:
https://ci.appveyor.com/api/projects/qucs/qucs/artifacts/qucs-win64.zip?branch=release-0.0.19?job=Environment%3A%20MSYSTEM%3DMINGW64%2C%20MBITS%3D64%2C%20MARCH%3Dx86_64

in3otd:
> Just tried this to narrow it down to one branch:

it works, if you change the `?` before `job` to `&` :grin:

guitorri:
Yes... Thanks! Fixed it above.

PR #568, opened 2016-10-06
error out if build essentials are not available

felix-salfelder:
... the build will fail anyway, see #524

### DIFF
--- a/qucs/configure.ac
+++ b/qucs/configure.ac
@@ -710,7 +710,7 @@ AC_PATH_PROGS(UIC, uic-qt4 uic, :, [$QT_DIR/bin:$PATH])
 AC_PATH_PROGS(RCC, rcc, :, [$QT_DIR/bin:$PATH])
 
 if test "$MOC" = ":"; then
-  AC_MSG_WARN([
+  AC_MSG_ERROR([
   The $PACKAGE package needs the 'Qt Meta Object Compiler' to compile properly.
   Though Qt itself may be properly installed including headers and libraries
   the 'moc' program is missing.  Possibly you need to install the full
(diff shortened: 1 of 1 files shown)


felix-salfelder:
as suspected, there's a problem with travis+apple. `:(`

don't know apple. need help.

guitorri:
The issue is around the installation of Qt. The Qt4 available with`brew`(package manager for OSX) does not contain the deprecated Qt3Support portion that we need. Hence, we are downloading the official Qt package from Digia.

Nothing was changed (on our side), but for some reason the package fails to install. I don't know if it is corrupt during download, if Travis has changed something related to disk space, permissions or something else... I restarted Travis again to see if gets unstuck. I will try to add a checksum for the package to check for corruption.

guitorri:
I think I found the reason for the regression on OSX. Travis now has multiple images for OSX. By default it is trying to run with the latest 10.11 at the moment:
```instance: a14bbde0-cc13-46a4-bf8d-0340646031bd:travis-ci-osx10.11-xcode7.3-1475610688```

Something was dropped in OSX 10.11 and the Qt 4.8.7 installer is no longuer works.
https://forum.qt.io/topic/60173/install-qt-4-8-7-on-a-mac-running-os-x-10-11-el-capitan

We can try to force the image to OSX 10.10 and see what happens.
https://docs.travis-ci.com/user/osx-ci-environment/#OS-X-Version

guitorri:
I don't see a point in rebasing this. It is now trowing an error as intended (Qt is not installing on OSX for his PR). With #573 Qt installation is back working. So, I will merge and see what happens.

PR #573, opened 2016-10-28
[CI] Qt install error, set OSX to version 10.10

guitorri:
The Qt 4.8.7 installer no longer works with OSX 10.11.
https://forum.qt.io/topic/60173/install-qt-4-8-7-on-a-mac-running-os-x-10-11-el-capitan

Force Travis to use OSX 10.10.
https://docs.travis-ci.com/user/osx-ci-environment/#OS-X-Version

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -102,6 +102,7 @@ install:
       brew link --force flex;
       brew install bison;
       brew link --force bison;
+      brew install python;
       pip install matplotlib;
     fi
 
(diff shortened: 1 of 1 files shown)


guitorri:
This should fix the regression in #568 and #524.
If this works, it might make sense to cherry pick it to develop

guitorri:
The Qt issue seems to be fixed, but the CI is timing out... need more work.

guitorri:
I rerun and it finished on time (50min allowed for OSX instances). If it happens again I will try to reduce the time it takes to setup the dependencies.

The Travis error above is because I accidentally pushed the branch also to origin (instead of my fork) and immediately deleted it. Travis tried to build and could not find the branch on the repository.

PR #580, opened 2016-11-03
Fix for issues #550 and #553

guitorri:
It was found that QucsLib tool didn't update neither the default
symbol nor the name/library of the selected component.

LoadDefaultSymbol() was removed and its code was placed where the
function was called.

The code for reloading the default symbol was moved to SymbolWidget
class.


guitorri:
Rebased and resolved .gitignore from #556.

PR #585, opened 2016-11-05
Fix Some GUI Issues (WIP)

nvdl:
Right click rotate when inserting a new component does not create ghosting.
Component movement after placement does not create ghosting.
Component is realigned to grid if previously misaligned using Ctrl key.
Selection box is broken in this commit and needs fixing.

### DIFF
--- a/qucs/qucs/mouseactions.cpp
+++ b/qucs/qucs/mouseactions.cpp
@@ -171,7 +171,7 @@ void MouseActions::editLabel(Schematic *Doc, WireLabel *pl)
 
   Doc->sizeOfAll(Doc->UsedX1, Doc->UsedY1, Doc->UsedX2, Doc->UsedY2);
   Doc->viewport()->update();
-  drawn = false;
+  //drawn = false;
   Doc->setChanged(true, true);
 }
 
@@ -214,7 +214,7 @@ void MouseActions::endElementMoving(Schematic *Doc, Q3PtrList<Element> *movEleme
 	Doc->insertNodeLabel((WireLabel*)pe);
 	break;
       case isMarker:
-	assert(dynamic_cast<Marker*>(pe));
+        assert(dynamic_cast<Marker*>(pe));
 	break;
     }
   }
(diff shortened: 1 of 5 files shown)


nvdl:
Work In P.

in3otd:
[rant mode on]
- please avoid mixing whitespace changes with functional changes: this increases the burden of the reviewers, which have to sort out significant and non-significant changes. It's ok if you use a different coding style for a new block of code you add or modify entirely, though it will be better if it uses the style of the surrounding code.
If you really cannot stand the existing (mixed) coding style:
  - try harder to accept it
  - if the above fails, please separate out the whitespace-only changes in an independent commit, so that it will be reviewed only once (while the functional changes *will* be read more than once)
`git rebase -i` and `git add -p` are your friends.

- why did you leave in leaving in old line of codes after commenting them out? We use a revision control system also to avoid things like these. A `diff` already tells what has been changed, no need to leave the old code in.

[rant mode off]

 apart from that, it seems fine so far, after a very quick test

If it's a work in progress, please add it in the title ("`[WIP]`") also (and remove it once you think it's finished).

nvdl:
> please separate out the whitespace-only changes in an independent commit, so that it will be reviewed only once (while the functional changes will be read more than once

Makes sense.

> try harder to accept it

Still trying to get used to organic growth.
We had a plan for code formatting. What happened to that?

in3otd:
> We had a plan for code formatting. What happened to that?

We still have a plan, I think. It's just postponed to after more important things have been taken care of. And I seem to remember we didn't reach an agreement on how/when do that. (reopen the discussion in the old thread, if you *really* like to :grin:)

> Keeping that for a while if something breaks (easier to spot). Will be removed in the end.

Thanks.

nvdl:
> We still have a plan, I think. It's just postponed to after more important things have been taken care of.

Yes, we may take that later. But I like the idea of one commit only dedicated to formatting. Those misplaced braces, spaces, comments etc. really distract me.

guitorri:
What is the situation here? I want to close the release.

nvdl:
Still in progress. Do not pull.

guitorri:
Moved base to `develop`.

PR #458, opened 2016-02-08
basic dynamic loading

felix-salfelder:
to be merged as soon as CI is happy. expect portability related readjustments....

PR #589, opened 2016-11-06
rearranging components a bit

felix-salfelder:
... so more components can be added later on (see e.g. qucs-nonfree PR#1)


felix-salfelder:
more info and background:
- legacy Components will derive from Symbol, which will derive from Element
- new components can be Symbols on their own
- toolbox on the left will display Elements
- will introduce a Command class, derived from Element (no, a command is not a component)
- need to cleanup Element::info public static members. they do not work

felix-salfelder:
weird Qt3 "smart" pointer list side effects. waiting for #662, will then rebase.

ra3xdh:
I am not very keen on C++ coded devices, but this PR could be useful. Some device may need custom processing on C++ level like Subcircuits and LibComp. An example of such device is my `SpiceLibComp` in Qucs-S. This PR doesn't conflict with my #659 . I can implement a special XML device and wrap it in a plugin. For example call it `SysLibComp`. Furthermore, I think non-free devices like BSIM, should not be hardcoded in C++. It's better to keep them in a special library.  I need only a legal way to show XML devices in left-side component palette. 

In current state, I cannot directly backport spice4qucs devices from Qucs-S using this approach. I am using the `Component::spice_netlist()` now to form a netlist entry. This is not implemented in mainline Qucs. I can backport only symbol view that has no sense without netlist. The backporting of spice4qucs devices will be possible, when more details in SPICE netlister will appear in #315 and #316.  Probably I will convert all my SPICE-specific symbols in an XML library and will use with new XML component.  

The following improvements could be done:
* It's possible to put in plugin only one component. It would be useful to put a group of devices into a single *.so library.  
* CMake support is missing.

felix-salfelder:
> level like Subcircuits and LibComp.

this PR will not touch subcircuits and LibComp, except that the code will move to a plugin.
that will happen as soon as the resistor is working the way it should. there are strange hacks in the way that need to be cleaned up. could you help with this?

> I cannot directly backport spice4qucs devices from Qucs-S using this approach. I am using the Component::spice_netlist() 

this is the component plugins PR, not the one for netlister(s).

> It's possible to put in plugin only one component. It would be useful to put a group of devices into a single *.so library.

a plugin is a shared object. what makes you think that it can only hold one component? please have a look at how the mechanism works, that will clarify things. note that the constructors of static instances are arbitrary C++ code. that code is executed at load time. it can be anything. *really* anything. the dynamic loader is introduced in #458, together with examples (tests).

> CMake support is missing.

isn't the autotools build system working as it should right now? i don't know how to fix cmake, maybe we could go with everything linked statically in cmake, until someone really wants that, steps in and fixes it. note that if plugin loading does not work for the cmake build (for some period of development) that will not be a regression.

need to get back to this PR. i guess we should sort out the preliminaries first.

PR #322, opened 2015-07-20
Add bsim610

guitorri:
# Add BSIM 6.1.0 to Qucs.
## Examples:

Qucs projects replicating of the benchmark tests released by the BSIM Group are available here:

https://github.com/guitorri/qucs-bsim6-examples

Most examples are within reasonable tolerances. The transient simulation needed a bit of help on the convergence, and some numerical damping is visible. Other than that, the results seem pretty good with respect to the published HSPICE data. 
## Results:

[idvd_nmos](https://github.com/guitorri/qucs-bsim6-examples/blob/master/idvd_nmos_prj/bench-idvd_nmos.pdf)
[idvd_pmos](https://github.com/guitorri/qucs-bsim6-examples/blob/master/idvd_pmos_prj/test_idvd_pmos.png)
[idvg_nmos](https://github.com/guitorri/qucs-bsim6-examples/blob/master/idvg_nmos_prj/test_idvg_nmos.png)
[inv_dc](https://github.com/guitorri/qucs-bsim6-examples/blob/master/inv_dc_prj/inv_dc.png)
[inverer_transient](https://github.com/guitorri/qucs-bsim6-examples/blob/master/inverter_transient_prj/test_inverter_transient.png)
## Notes:
- Why two models, NMOS and PMOS? 
  
  The original `bsim6.va` model makes use of parameters to define default values of other parameters. As follows: 
  
  ```
  parameter integer TYPE = `ntype  from [`ptype:`ntype] exclude 0;
  ...
  parameter real AIGC = (TYPE == `ntype) ? 1.36e-2 : 9.8e-3; 
  parameter real BIGC = (TYPE == `ntype) ? 1.71e-3 : 7.59e-4;
  ```
  
  Such dynamic change of default parameter values is not available in `qucsator`. Default values are stored in static structs. For the moment was just _easier_ to add two models with the default parameter values.
- Other issues:
  
  There other kinds of paramete dependency (up to 3 levels of dependecy), such as:
  
  ```
  parameter real NDEP  = 1e24;
  ...
  parameter real NDEPCV = NDEP;
  ```
  
  After the model is compiled, what should be the behavior when NDEP is changed? Should NDEPCV also be updated? For the same above reason this is is not yet supported. The default values are the ones available during compilation.
  - SPICE Model card:
    
    Using the following MOS modelcards in `qucsconv` should map to the BSIM 6.1.0 model.
  
  ```
  .model Tnmos NMOS LEVEL=6 VERSION=6.10 [...]
[...]

PR #595, opened 2016-11-06
BSIM6 and other XML fixes

guitorri:
These fixes are needed if one wants to compile the (patched) BSIM6 model. 
The patched BSIM6 model will be added to the qucs-nonfree repo later on.


guitorri:
Cherry-pick from #322.

PR #596, opened 2016-11-07
Fix regression on VHDL netlisting.

guitorri:
When netlisting for VHDL simulation a `generic();` entry
is being added without parameters which is an error.

Bug introduced in 8a30ad998bb6fe, after 0.0.18.


guitorri:
Self-merging this...

PR #597, opened 2016-11-09
Readme: Add info about branching strategy

guitorri:
[skip ci]

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,12 +20,24 @@ Qucs is an integrated circuit simulator which means you are able to setup a circ
     - <http://qucs.github.io/qucs-doxygen/qucs/index.html>
     - <http://qucs.github.io/qucs-doxygen/qucs-core/index.html>
   - Downloads: <http://sourceforge.net/projects/qucs/files/>
+  - Translation platform: <https://www.transifex.com/qucs/public/>
 
 ## Currently Supported Platforms
   - GNU/Linux
   - Darwin/OS X
   - FreeBSD
   - Windows
+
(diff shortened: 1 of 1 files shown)


in3otd:
seems good; to be nit-picking, I'll say that the sentence [here](https://github.com/Qucs/qucs/pull/597/commits/4d5ac95eb6dbd45e02e1db9cc95cd830f945f4a8#diff-04c6e90faac2675aa89e2176d2eec7d8L32) should probably be reworded to better match the added explanation above; both stable and development will be available via git. Similarly [here](https://github.com/Qucs/qucs/pull/597/commits/4d5ac95eb6dbd45e02e1db9cc95cd830f945f4a8#diff-04c6e90faac2675aa89e2176d2eec7d8R54), since the instructions that follow there will clone the stable master and not "latest code in development".

I was wondering if we need to add some explicit instruction to check out a branch other than master, probably not, as stackoverflow has all the explanations, but maybe point out that the instructions provided will clone the master branch.

guitorri:
Thanks. Makes sense. I will fix it. I actually wrote about branch switching and deleted. I will add the master as default and info about how to switch.

You know you can edit directly on my branch? We can try that another time.
I am now checking the 'allow edits from maintainers" on my PRs so maintainers can push to each others branches... seems neater than doing a PR on top of PR. I haven't tried yet.

in3otd:
thanks, look good to me.

Didn't know I could push to your PR branches directly, I'll try that one day :grin:

PR #599, opened 2016-11-13
Allow path settings to be overridden by env. vars

guitorri:
The `loadSettings()` must done earlier otherwise the
override by environment variables won't work.

In particular for these that are still stored in the settings.
 - AdmsXmlBinDir
 - AscoBinDir

This has to be simplified to make proper use of the PATH,
related to commens in issue #374.

### DIFF
--- a/qucs/qucs/main.cpp
+++ b/qucs/qucs/main.cpp
@@ -657,6 +657,11 @@ int main(int argc, char *argv[])
   QucsSettings.dx = w*3/4;
   QucsSettings.dy = h*3/4;
 
+  // load existing settings (if any)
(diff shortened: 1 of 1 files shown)


guitorri:
Sorry, I will self-merge (again...). 

This is needed to make a portable Windows version. I am writing a wrapper cmd script that sets QUCSDIR (and other override vars if needed) for the session. Similar to what `octave.bat` does. So avoiding the need to mess with environment variables. 
The installer still needs the QUCSDIR set I will try to work around it. QUCSDIR is a problem because it makes it hard to have two Qucs versions independent of each other.

PR #605, opened 2016-11-19
qucs: fix QucsHomeDir override by stored settings

guitorri:
Small regression introduced with a3eb8c3da3,
causing Qucs to always start at $HOME/.qucs/


guitorri:
Self merging again...

PR #606, opened 2016-11-19
Fix file extension handling in Import/Export data dialog

in3otd:
In the Import/Export data dialog a file named e.g. `LDMOS_7.5V_50mA.s2p` is currently not recognized as a `s2p` (Touchstone) file, since the file extension is taken as the part after the *first* dot, i.e. `5V_50mA.s2p`.
I have updated the code in the Import/Export data dialog to take as extension only the part after the *last* dot (i.e. use `QFileInfo::suffix()` instead of `QFileInfo::extension()` and `QFileInfo::completeBaseName()` instead of `QFileInfo::baseName()`).
I have also updated similar parts of the code where file extensions/base names appeared to be used incorrectly and updated the deprecated Qt3 functions with the Qt4 ones.

### DIFF
--- a/qucs/qucs/dialogs/importdialog.cpp
+++ b/qucs/qucs/dialogs/importdialog.cpp
@@ -16,7 +16,7 @@
  ***************************************************************************/
 #include <QLabel>
 #include <QLineEdit>
-#include <QTextEdit>
+#include <QPlainTextEdit>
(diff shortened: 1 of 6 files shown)


guitorri:
Looks sane to me. Thank you.

Issue #593, opened 2016-11-06
Cannot paste schematic from previous version(s)


in3otd:
In the GUI we can open schematics from previous versions but when trying to paste a schematic from a previous version a "Wrong document version" error pops up.

This comes from [here](https://github.com/Qucs/qucs/blob/release-0.0.19/qucs/qucs/schematic_file.cpp#L131) where the pasted schematic version needs to be the same as the current Qucs version.

To easily reproduce:
- open a schematic
- select all, copy and paste in a text document
- change the version in the first line to `0.0.18`
- select all, copy and try to paste in a schematic

(could maybe reuse the first [commit](https://github.com/Qucs/qucs/pull/340/commits/85bb09531dfe8ca0237ab17233feaccf144c12a1) of #340 to ease checking)

ra3xdh:
Yes, I know about this problem. I always use clipboard editing to allow copy-pasting between versions. This issue should not concern simple users, because they usually uses only single Qucs version. But it makes development and debugging a bit more difficult.

felix-salfelder:
perhaps, as long as loading old schematics is not fully tested, we should warn "loading old schematic, please be careful", instead? `return false` seems to be entirely pointless to me.

in3otd:
yep, I was sure this was already known, but wanted to have a remainder to fix it :grin:

Currently we load old schematics without any user warning, I think we should have the same behavior (emit warning or not) when pasting. When `0.0.19` will be out, all existing users' (vs. developers) schematics will become old.

guitorri:
There also something about loading future versions:
https://github.com/Qucs/qucs/blob/release-0.0.19/qucs/qucs/dialogs/qucssettingsdialog.cpp#L134

in3otd:
I was trying to understand the logic of the current code about loading from future versions: at present having it checked or unchecked makes practically no difference...

- when trying to load a file from a future version, the user is *always* asked if ey really wants to open the file (see the code [here](https://github.com/Qucs/qucs/blob/release-0.0.19/qucs/qucs/schematic_file.cpp#L868-L885)), regardless of the related setting
- also, currently we show all the `.sch` files in the tree view, also from future versions, regardless of the related setting status (see the code [here](https://github.com/Qucs/qucs/blob/release-0.0.19/qucs/qucs/projectView.cpp#L130-L138); `n` is negative for future versions but this case is not actually treated differently)

So, what I propose is to modify the current behavior as:
- if *load from future versions* is enabled:
[...]

PR #607, opened 2016-11-19
Allow pasting from future versions

in3otd:
Fix for #593, as described there, copied here for convenience:

- if *load from future versions* is enabled:
  - show future-version schematic in the tree view
  - open them without any warning message
  - paste from future versions without any warning message
- if *load from future versions* is **not** enabled:
  - do **not** show future-version schematic in the tree view
  - abort opening them (if forced thru the *Open* dialog) with an error message
  - abort pasting from future versions with an error message

### DIFF
--- a/qucs/qucs/components/libcomp.cpp
+++ b/qucs/qucs/components/libcomp.cpp
@@ -102,8 +102,9 @@ int LibComp::loadSection(const QString& Name, QString& Section,
 
   int Start, End = Section.indexOf(' ', 14);
   if(End < 15) return -3;
-  QString Line = Section.mid(14, End-14);
-  if(!misc::checkVersion(Line)) // wrong version number ?
+  QString Line = Section.mid(14, End-14); // extract version string
+  VersionTriplet LibVersion = VersionTriplet(Line);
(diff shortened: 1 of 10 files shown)


guitorri:
I rebased to fix a conflict, added a commit and pushed `-f` back to your branch.  
I hope it is ok, otherwise we keep bouncing messages and PRs around... What do you think?

in3otd:
> I hope it is ok

if you mean the `push -f` on "my" PR, that's fine with me.

Thanks for the added commit to refresh the project in case the setting is changed.

A small detail: I've just notices that [here](https://github.com/Qucs/qucs/pull/607/commits/19ea4772a29f97a18ad9a53c891d5382f6e5e7e0#diff-21d4c99ec9cdd653d28a62942ff16889R135) I should probably have done as [here](https://github.com/Qucs/qucs/pull/607/commits/19ea4772a29f97a18ad9a53c891d5382f6e5e7e0#diff-21d4c99ec9cdd653d28a62942ff16889R875). Does not make a difference in practice, I think.

guitorri:
It will create 2 strings to be translated, but the behavior is probably the same.

guitorri:
I will fix that.

PR #608, opened 2016-11-20
VHDL: Fix char* to QString conversion

guitorri:
This is causing VHDL simulation to fail.
For some reason the scale part is not converting correcty, generating
garbage in place of the time scale.

0.0.18    :    netCLEAR <= '1'; wait for 10 ns;
before fix:    netCLEAR <= '1'; wait for 10 1;

### DIFF
--- a/qucs/qucs/misc.cpp
+++ b/qucs/qucs/misc.cpp
@@ -327,7 +327,8 @@ bool misc::VHDL_Delay(QString& td, const QString& Name)
 bool misc::VHDL_Time(QString& t, const QString& Name)
 {
   char *p;
-  double Time = strtod(t.toLatin1(), &p);
+  QByteArray ba = t.toLatin1();
+  double Time = strtod(ba.data(), &p);
   while(*p == ' ') p++;
   for(;;) {
     if(Time >= 0.0) {
@@ -389,7 +390,8 @@ bool misc::Verilog_Delay(QString& td, const QString& Name)
 bool misc::Verilog_Time(QString& t, const QString& Name)
 {
   char *p;
-  double Time = strtod(t.toLatin1(), &p);
+  QByteArray ba = t.toLatin1();
+  double Time = strtod(ba.data(), &p);
   double factor = 1.0;
   while(*p == ' ') p++;
   for(;;) {


in3otd:
...you were too fast and forgot the `"fs"` case :grin:

do you have a test circuit for this? I'm curios to understand why this problem appeared...

guitorri:
Thank you. Somehow I deleted the first line... I will add it back.
I used the BCD example: http://qucs.sourceforge.net/examples/BCD_prj.tar.gz
It is not the first time I have issues with `char *` to `QString`. In this case it might be because the `char *` is not terminated by `\0`?

guitorri:
Added it back and squashed.

in3otd:
googling around I think I understood the root cause...

[here](https://github.com/Qucs/qucs/blob/release-0.0.19/qucs/qucs/misc.cpp#L330) the `t.toLatin1()` creates a `QByteArray` whose lifetime ends at the end of the expression. But `p` is now a pointing to someplace inside this `QByteArray` whose existence is not guaranteed after the line is executed. So the rest of the code is using a pointer to something which is not longer guaranteed to be valid. Using `gdb` and adding a `watch` to the location pointed by `p` shows that the memory area once used by the `QByteArray` is overwritten while creating the final string at the [end](https://github.com/Qucs/qucs/blob/release-0.0.19/qucs/qucs/misc.cpp#L348).

To avoid this, we can do it in two steps, storing the `QByteArray` first and doing the conversion to `double` afterwards, like 
```
QByteArray ba = t.toLatin1();
double Time = strtod(ba.data(), &p);
```

so that the `QByteArray` is not destroyed until the end of the function and `p` continues to point to valid data. Seems to work, here.

See [this discussion](http://stackoverflow.com/questions/5505221/converting-qstring-to-char) on Stack Overflow.

Potentially the same problem is present also in [`Verilog_Time()'](https://github.com/Qucs/qucs/blob/release-0.0.19/qucs/qucs/misc.cpp#L389-L392)

guitorri:
Thank you. I will use your fix which is sorter. `Verilog_Time()` is not affected because the `&p` char is not used afterwards. I will change it also for consistency.

in3otd:
> Verilog_Time() is not affected because the &p char is not used afterwards

uhm, the pointer is actually used after the [line] (https://github.com/Qucs/qucs/blob/release-0.0.19/qucs/qucs/misc.cpp#L392) where the `QByteArray` formally ends to exist, in all the `strcmp()` that follows. It may actually work, since no further objects are (appears to be) allocated in the meantime, so the memory area which was used by the  `QByteArray` still contains the expected string but there is no guarantee of that, IIUC.
Anyway, I saw you modified that part also, so now it should be ok.

guitorri:
Thank you.

Issue #615, opened 2016-11-24
Export As Image


nvdl:
Image exporting overwrites the image without prompt if the "Browse" dialog is not used.

![export](https://cloud.githubusercontent.com/assets/3097827/20597830/3b2c9978-b246-11e6-801a-57fb3cff3585.png)

ra3xdh:
It is known featurebug of image export. Let's consider it as a bug. See #616 for fix.

PR #616, opened 2016-11-24
Fix for #615

ra3xdh:
This PR is fix for #615. Added a dialog when image file to export already exists. It will preserve image file form automatic overwriting.

### DIFF
--- a/qucs/qucs/imagewriter.cpp
+++ b/qucs/qucs/imagewriter.cpp
@@ -155,6 +155,16 @@ int ImageWriter::print(QWidget *doc)
 
   if (dlg->exec()) {
     QString filename = dlg->FileToSave();
+    if (QFile::exists(filename)) {
+        int r = QMessageBox::question(0, QObject::tr("Overwrite"),
+                                         QObject::tr("File \"%1\" already exists.\nOverwrite ?")
+                                         .arg(filename),
+                                         QMessageBox::Yes|QMessageBox::No);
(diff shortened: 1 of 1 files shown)


qucs/qucs/imagewriter.cpp
```diff
@@ -156,6 +156,16 @@ int ImageWriter::print(QWidget *doc)
   if (dlg->exec()) {
     QString filename = dlg->FileToSave();
     lastExportFilename = filename;
+    if (QFile::exists(filename)) {
+        int r = QMessageBox::question(0,QObject::tr("Export as image"),
+                                      QObject::tr("File already exists."
```
ra3xdh: I replaced messages by already translated from the `qucs_actions`.

qucs/qucs/imagewriter.cpp
```diff
@@ -156,6 +156,16 @@ int ImageWriter::print(QWidget *doc)
   if (dlg->exec()) {
     QString filename = dlg->FileToSave();
     lastExportFilename = filename;
```
in3otd: nit-picking/question: should the `lastExportFilename` be updated only if actually used successfully, i.e. move the assignment after the `if (r == QMessageBox::No)` check?

ra3xdh:
> should the lastExportFilename be updated only if actually used successfully

Yes, `lastExportFilename` should be updated only if file was actually exported. I have changed this code.

PR #610, opened 2016-11-23
Move examples to qucs/examples/

guitorri:
Continuation of #544 

- remove compressed examples, moved to the website
- move top-level `example` directory to `qucs/examples`

PR #619, opened 2016-11-26
Correct handling of QUrl paths

in3otd:
while testing #610 I saw that in Linux the file explorer opened to browse the examples did not show the file system tree correctly, since the path was beginning with `//` instead of `/`.

Tested to work fine on Linux and Windows here.


guitorri:
Tested on Mac here. Thank you.

PR #620, opened 2016-11-29
Tutorial on RF Amplifiers with S-Parameter Techniques

gogobers:
Added a Tutorial to design RF-Amplifiers using S-Parameter techniques, such as
gain- and noisecircles. Complete project is available, but git ignores part of
the files (such as .sch etc.)

PR #624, opened 2016-11-30
Stut

gogobers:
Added also the example files for the S-Parameter Amplifier tutorial, needed to change .gitignore for taht


in3otd:
thanks, no need to open a new PR, you can just keep using the previous one. I saw the here you changed the destination branch; you should be able to do that on an existing PR using the "Edit" button near the PR title or we can do that for you if GitHub does not allow you to do that. In any case that's not too important here, a PR can be (manually) merged to any branch.
And BTW we are now using `master` as the branch containing the latest stable release, so PRs should mainly go in `develop` now.

I will close this PR and keep the other one (#620); I still need to take a look at the latest changes and push my `Makefile`s updates.

Issue #629, opened 2016-12-05
Paintings could not be added into subcircuit symbols


ra3xdh:
I have found a bug that can stop release preparation. Steps to reproduce:

* Open schematic, containing subcircuit symbol
* Switch to subcircuit symbol mode (F9)
* Try to add a painting.  Circuit components will be added instead. 

Attached screenshot shows what happen:
![subcir_bug](https://cloud.githubusercontent.com/assets/4920080/20878278/7db13a70-bae7-11e6-93c1-30659ee1161d.png)

in3otd:
I think I know where this comes from, I'll try to push a fix today

ra3xdh:
Fixed by #630

PR #630, opened 2016-12-05
629 fix - paintings selection in symbol painting mode

in3otd:
Quick fix for #629 ; seems ok here, but tested only briefly .


ra3xdh:
I have found no problems with this fix. Subcircuit symbol editing mode is repaired and searching is working. I think it could be merged.

Issue #627, opened 2016-11-30
 Qucs uses QUCSDIR unconditionally


guitorri:
From #353.

> 
> It is/will be a common situation that a Windows user, which has previously installed Qucs, uses the portable version without uninstalling the old version first. The result will be that some/many things will be taken from the old version, as seen above. IMHO, the portable version should have a qucs.bat file in the base directory that unsets QUCSDIR and then launches Qucs. And/or even better a startup message that inform the user that QUCSDIR is set and this should not normally be.
> 
> I didn't look at the installer but I think it should make sure that QUCSDIR is either not set (best) or set to the newest Qucs (but this may break the old install).

guitorri:
Right, I still have to push the modifications I did for the portable version.
But it does set QUCSDIR in the beginning of the `qucs-portable.bat` (can rename to `qucs.bat` as you say):
```
Rem   Get the location where the batch file resides.
Rem   Note: QUCSDIR ends with a backslash
set QUCSDIR=%~dp0
```

I did not prepare any Windows installers for the `-rc3`, but they still set QUCSDIR system wide... (the very reason why I am refraining from doing them)...

in3otd:
I think we should move away from `QUCSDIR` (didn't we do quite some work in the past for this very same purpose?); IIRC, on Linux is normally not set, except if the user explicitly sets it (also on Mac, I guess?), so the same should be on Windows. Just unset `QUCSDIR` in the `.bat` file, so Qucs will run in the same way as commonly done in the other platforms.

Can the Windows installer (if you plan to do one) be used to *remove* a environment variable? Will it ask to uninstall the old Qucs version first?

OTOH, as mentioned above I think we should really have a startup check that informs the user if `QUCSDIR` is set (or, at least, set to a different value than the actual app bin path) and a message that inform the user that it should not normally be.

felix-salfelder:
need to understand what `QUCSDIR` is trying to implement. (then get rid of it.)

evidently it was reinventing `PATH`, and that's trivial to fix. but what else?

> startup check that informs the user if QUCSDIR is

really? `QUCSDIR` is (supposed to be) a *user variable*. so a user should not set it, unless required.

try `$ PYTHONPATH=/tmp python`. it does not "inform the user", and i think that's the only sensible thing to do. but sure, put whatever you like in a startup script...

in3otd:
> need to understand what QUCSDIR is trying to implement.

[...]

PR #634, opened 2016-12-23
Pkg win portable

guitorri:
First version of scripts for Windows portable package.
Unsed QUCSDIR on batch script to allow older and new versions side by side, see #627 for comments.


ra3xdh:
I have used something similar `qucs.bat` when I was need to prepare a portable version of Qucs for myself. I think this PR could be merged.

PR #635, opened 2016-12-25
"init" is 2nd parameter, not 1st

markablov:
Properties are:
```
  Props.append(new Property("Num", "1", true,
		QObject::tr("number of the port")));
  Props.append(new Property("init", "low", false,
		QObject::tr("initial output value")+" [low, high]"));
  Props.append(new Property("times", "1ns; 1ns", false,
		QObject::tr("list of times for changing output value")));
  Props.append(new Property("V", "1 V", false,
		QObject::tr("voltage of high level")));
```
We want to get value of "init" property, so we should access Props[1], not Props[0]

### DIFF
--- a/qucs/qucs/components/digi_source.cpp
+++ b/qucs/qucs/components/digi_source.cpp
@@ -110,7 +110,7 @@ QString Digi_Source::vhdlCode(int NumPorts)
   int z = 0;
   char State;
   if(NumPorts <= 0) {  // time table simulation ?
-    if(Props.at(0)->Value == "low")
+    if(Props.at(1)->Value == "low")
       State = '0';
     else
       State = '1';


in3otd:
...thanks for spotting this; I see that the property number was changed [a while ago](https://github.com/markablov/qucs/commit/91e7e0902cf94647652ee1d75ea6f9b688b4b389#diff-a980195043514f7964c29831ff05e3a0R114) by @guitorri , likely by mistake; maybe he remembers why that was done.

I have changed the target branch for this PR to `release-0.0.19`, as it makes sense to include this in the upcoming release.

in3otd:
closing and reopening to (try to) force Travis to notice the target branch change...

in3otd:
merged in `release-0.0.19` (190689613be76c67dbcf6c39a5e9a663e3d710f4)

Issue #433, opened 2016-02-01
Remove qucs-edit


guitorri:
Do we really need to provide yet another editor?

in3otd:
well, even if it's the one I normally use with `qucs` (old habits...) can't really see a reason to keep it. BTW, the alternative "text edit" Tab provides the same functionality, I think.

ra3xdh:
Fixed by #652 .

PR #652, opened 2017-01-29
Remove qucs-edit.

guitorri:
Close issue #433


in3otd:
should we remove [this stuff](https://github.com/Qucs/qucs/blob/master/qucs/qucs/qucs_actions.cpp#L681-L699) also?

guitorri:
Yes. I grepped `qucs-edit`and missed this one...

guitorri:
There might be a bit more straightening to do, but I don't want to deviate much from the original issue.
Now text files should either open on the built-in qucs tab (default) or an external editor defined in the settings.

felix-salfelder:
what does `env.insert("PATH", env.value("PATH") );` do in line 711?
(has already been there)

otherwise: looks good.

guitorri:
I think it is an attempt to set the environment for the process. Surely not needed as it is passed as absolute path. I will remove.

in3otd:
that `env.insert("PATH", env.value("PATH") )` is used in a few other places...  better make a not to remove it everywhere :grin:

felix-salfelder:
@in3otd: do you understand what it *does*? it looks like it gets `PATH` from `env`, and puts it back there. how does it have any effect?

thanks

in3otd:
no, I don't understand what it *really* does, I was hoping that guitorri would check :grin: 

One thing that I think will need to be checked is how that behaves in Windows as I have the impression that the `QProcess` stuff works a bit differently there.
`git blame` says that all of them were added by @fransschreuder maybe he remembers why that was needed.

fransschreuder:
Oh did i create these? I think the getenv lines were inserted before i ever started working on qucs. I guess the lines can be removed

guitorri:
I will remove and try to run the build produced by appveyor on Windows just to make sure.

guitorri:
Followup ticked for the environment thing in #663.

in3otd:
testing this I saw that if I select e.g. `emacs` as editor it show an error box saying just *Cannot start text editor!*. Which bothers me, because
- `emacs` is in the path. Maybe we should give a hint that a full path is expected? A *Browse* button? Or just use the path, when a non-absolute application name is specified?
- the message does not which text editor/which full path was tried. Actually [it should](https://github.com/guitorri/qucs/blob/30cc866bc33d83868f25bda2281b958981910281/qucs/qucs/qucs_actions.cpp#L704) but using `canonicalFilePath()` on a non-existent file [returns an empty string](http://doc.qt.io/qt-4.8/qfileinfo.html#canonicalFilePath#canonicalFilePath).  Should we add something "*(tried: /usr/bin/editorname)*" to the message to make it more explicit? We should never have an empty full path there but. I think we already saw something similar somewhere.
Ok, maybe this latter belongs to #378.

Apart from this, works fine here.

guitorri:
Thank you. I will addess the issues.

in3otd:
...while you are at it, you could also update the brief comment about [`slotTextNew()`](https://github.com/guitorri/qucs/blob/dd4b80230d5e2742a12036651d402d437a604660/qucs/qucs/qucs.h#L90)

guitorri:
Should be a bit better now.
A few issues still. The `/usr/bin/emacs` I have will only work on the terminal. TTY I guess. So it will fail.
Further, how text editor take input files: `editor <input>` or `editor -i <input>`.

in3otd:
seems to work fine here, also on Windows.

`emacs` works fine here, not sure I understood the issue you see there.
The command line to use for the external editor *might* be an issue, even if I guess most of the editors work with just `editor <input>`. A similar and probably more common issue is how to tell them to open a new (non existing) file. On Windows, calling `Notepad++` with no parameters just brings up the last file edited. Still I will tend to consider these things as minor, so I'd like to merge this PR unless you would like to do some more work on it.

guitorri:
I was not planing on working further unless really needed. 
I will resolve the conflicts and update the branch.

guitorri:
Rebased.

PR #657, opened 2017-01-31
Tapered line component: rebase #380 to develop

andresmmera:
This PR is a rebase of PR #380 vs develop branch. 

## #380 content
This component is aimed to model the S-param behaviour of a tapered line. It provides 4 impedance profiles: Exponential, Linear, Triangular and Klopfenstein [1]. The linear profile is not defined in [1] but I have added it anyway.

Regarding the implementation details, I didn't find a explicit model which calculates the S-Parameters of a tapered line, so I decided to tackle the problem as follows:

1) The transmission line is divided into differential slots (the size of each slot is set equal to 3e-3*lambda), where lambda is the wavelength. This means that the size of the slot depends on the frequency of the S-param simulation.
2) Every slot is supposed to have uniform characteristic impedance (because its length is much smaller that than wavelength), so the ABCD matrix can be calculated as:

ABCD = [cosh(gamma·dl), Zi·sinh(gamma·dl); sinh(gamma·dl)/Zi, cosh(gamma·dl)]

where gamma = alpha+j·beta is the complex propagation constant and dl is the length of the slots.
https://en.wikipedia.org/wiki/Two-port_network#Table_of_transmission_parameters

3) According to the cascading property of the ABCD parameters, the overall ABCD matrix can be calculated as product of the ABCD matrices of the slots.

4) ABCD is converted into S/Y parameters.

I reckon that this model works, indeed the |S11| perfectly matches the reflection coefficient plots at [1]. Here are the results:

a) Impedance profile
a.1) Qucs taper line model
zprofile
a.2) Pozar [1]
pozarzprofile

b) Exponential taper:

b.1) S11
exponentials11
b.2) S21
exponentials21

c) Triangular taper:

c.1) S11
triangulars11

c.2) S21
triangulars21

d) Klopfenstein taper:

d.1) S11
klopfensteins11

d.2) S21
klopfensteins21

As mentioned above, the |S11| seems to mimic the reflection coefficient plot given in [1]:
pozars11

Please, notice that the length of the line is l=150mm, so freq = 1e9 <=> beta·l = pi

The plots above are taken from (this schematic)[https://gist.github.com/andresmmera/7b1b7a96d68ec0b8741c]. I have also written an [Octave script](https://gist.github.com/andresmmera/c64f23153223bef83fe1) in order to verify the model.

[1] Microwave engineering. David M Pozar. 4th Edition. Pages 261-267


qucs-core/src/components/taperedline.cpp
```diff
@@ -0,0 +1,268 @@
...
+ nr_double_t l = getPropertyDouble ("L");
+  if (l < 0) return;
+  // calculate noise using Bosma's theorem
+  nr_double_t T = getPropertyDouble ("Temp");
+  setMatrixN (4.0 * celsius2kelvin (T) / T0 * real (getMatrixY ()));
+}
+
+void taperedline::initDC (void) {
+  allocMatrixMNA ();
+  // open circuit
```
in3otd: oh, should not really be an open.

Should probably be done like here https://github.com/Qucs/qucs/blob/master/qucs-core/src/components/tline.cpp#L83-L102 . Slightly more difficult since the line is non-uniform.

qucs-core/src/components/taperedline.cpp
```diff
@@ -0,0 +1,268 @@
...
+
+void taperedline::calcImpedanceProfile()
+{
+nr_double_t L = getPropertyDouble ("L");//Length
+  nr_double_t Z1 = getPropertyDouble ("Z1");//Port 1 impedance
+  nr_double_t Z2 = getPropertyDouble ("Z2");//Port 2 impedance
+
+  if (Z1 > Z2)//Handling Z1 > Z2 case
+  {
+     logprint (LOG_ERROR, "WARNING: The impedance at port 1 is bigger than the impedance at port 2 ((Z1 = %g Ohm ) > (Z2 = %g Ohm))\n", Z1, Z2);
```
in3otd: uhm, OK for the warning but if Z1 > Z2 the results are wrong. You are exchanging the ports impedances but not the actual ports connections. Either simply throw an error for this case or handle it by also swapping the two ports (you could use something like S.exchangeRows(0,1); S.exchangeCols(0,1))

andresmmera:
Hell, I didn't notice that stuff... I'll try to fix these issues ASAP

in3otd:
I too didn't notice those during the last review some months ago... I guess a break between reviews is a good thing :grin:

Don't know how easy is to solve exactly the DC case; if too complex I think that using some sort of "constant width" approximation and issuing a waning could be acceptable. But numerical integration may be ok.

The commit message contains some history that is no longer relevant, it may be good to clean it up. I can do that if you are short of time.

qucs/qucs/components/taperedline.cpp
```diff
@@ -0,0 +1,84 @@
+/*
+* taperedline.cpp - Exponentially tapered line implementation
```
in3otd: here and in the `.h`, "Exponentially" should be removed, it's much more general now :grin:

andresmmera:
I was thinking about num. integration too for the DC case. I reckon that it
is more correct.
I'll try to remove that unrelevant commits manually this evening, but
thanks for your offer. In case of panic with the git command I'll ask you
for help on the IRC chat :-)

in3otd:
yes, numerical integration will be fine, but the approximate solution will be faster to implement :grin: . I was initially thinking at a closed-form solution but it's likely not worth the effort, if at all possible.

andresmmera:
Hi there,

I couldn't work on Qucs these days... I've just fixed the Z1 > Z2 as @in3otd suggested.
For the moment, I find more convenient to ignore the DC losses... (a warning is thrown in case of a DC analysis)

May I squash the commits?

guitorri:
yes, you can squash the commits.

andresmmera:
Rebased

in3otd:
pushed some correction to the docs, in several commits just for clarity, feel free to squash everything after review

andresmmera:
Reviewed. I fully agree with the changes.

in3otd:
merged (63c6dec6a999444f3776499f5fc583f779855506)

PR #660, opened 2017-02-01
builddir environment

felix-salfelder:
this fixes running qucs from builddir. this is required for the "check" target to work. (#376 rebased)


qucs/qucs/main.cpp
```diff
@@ -706,7 +706,7 @@ int main(int argc, char *argv[])
   QucsSettings.Editor = "qucs";
 
   /// \todo Make the setting up of all executables below more consistent
-  var = getenv("QUCSATOR");
+  getenv("QUCSATOR");
```
guitorri: Isn't `var = getenv()` needed here?

qucs/qucs/Makefile.am
```diff
@@ -42,7 +52,7 @@ MOCHEADERS = \
 
 MOCFILES = $(MOCHEADERS:.h=.moc.cpp)
 
-qucs_SOURCES = node.cpp element.cpp qucsdoc.cpp wire.cpp mouseactions.cpp   \
+qucs_real_SOURCES = node.cpp element.cpp qucsdoc.cpp wire.cpp mouseactions.cpp   \
```
guitorri: Please enlighten me, why _real_ is needed here?

qucs/qucs/main.cpp
felix-salfelder: !!

thanks

qucs/qucs/Makefile.am
felix-salfelder: "qucs", the executable is a wrapper script managed by autotools.

this PR introduces a custom wrapper script "qucs", generated from "qucs.in", it only lives in the builddir. here it executes "qucs_real".

the "_real" suffix for the autotools wrapper is arbitrary, it is stripped in line 43 (during install).

felix-salfelder:
updated. still needed for testing in upcoming PRs...

felix-salfelder:
can we have this in .20.? seems like a user (leon-vv) has aked for it in the meantime

guitorri:
Testing.

in3otd:
not sure it's related to this, but when I start `qucs` from the builddir I get a message on the console saying
`make: *** No rule to make target 'subdirs'.  Stop.`
but the GUI actually comes up and seems to work.
Any idea where that message comes from?

felix-salfelder:
thanks for spotting this.

yes, there once was a `subdirs` target, which is used to compile `PATH` in the wrapper script.

    subdirs:
    \techo ${SUBDIRS}

(i don't know where it went, probably lost during a rebase)

PR #510, opened 2016-05-01
Transcalc stripline

andresmmera:
Hello all, 

Here I've added a new entry on Transcalc for stripline calculations. The analysis/synthesis equations were taken from [1] and the results were checked with [Mantaro's stripline calculator](http://www.mantaro.com/resources/impedance-calculator.htm)

[1] Analysis Methods for RF, Microwave and Milimeter-Wave Planar Transmission Line Structures,
Cam Nguyen. John Wiley and Sons Inc., 2001, 

selection_057

PR #661, opened 2017-02-01
Rebase transcalc stripline to develop

andresmmera:
This PR is a rebase of #510 to _develop_. I've squashed the commits and kept the _develop_  version of .gitignore as @ra3xdh suggested

## #510 description

Here I've added a new entry on Transcalc for stripline calculations. The analysis/synthesis equations were taken from [1] and the results were checked with [Mantaro's stripline calculator](http://www.mantaro.com/resources/impedance-calculator.htm)

[1] Analysis Methods for RF, Microwave and Milimeter-Wave Planar Transmission Line Structures,
Cam Nguyen. John Wiley and Sons Inc., 2001, 

selection_057


ra3xdh:
Merged. You probably were need to closed the old PR and open new one. You may use rebase and the force push `push -f` to update the old commits.

qucs/qucs-transcalc/transline.h
```diff
@@ -25,6 +25,8 @@
 
 class QucsTranscalc;
 
+const double c0 = 299792458;// Speed of light (m/s)
```
in3otd: you're going too fast, leave some time for reviewing to others also...
Nothing bad here at a first sight, but things like needing to define `c0` *should* smell bad; it's in fact already defined [here](https://github.com/Qucs/qucs/blob/9d924caa3879fbca978987a9fe964eadbc5f0fc4/qucs/qucs-transcalc/units.h#L35) (but at least it has the same value :grin:) and used in many other places.

in3otd:
and it should be mentioned that Kevin also added the stripline code to his branch one and a half years ago, together with the tuner stuff, see https://sourceforge.net/u/tipofthesowrd/qucs/ci/tunerdialog/tree/qucs/qucs-transcalc/stripline.cpp . He ported the QucsStudio implementation, did not check if/how that differs, could have been nice to compare.
For example, it seems that here the synthesis code does not refine its results using the analysis code (which is usually more precise); when you press *Analyze*,  *Synthesize* and then *Analyze* again the results should not change (much) but this may not be the case here.

andresmmera:
Ops, I wasn't aware of Kevin's work... I'll take a look.
@ra3xdh Would it be possible to undo the merge so as to correct the issues @in3otd mentioned?

ra3xdh:
Automatic unmerge will not work here. Unmerge could be done using the hard reset and force push. I will look for a more safe way tomorrow.

in3otd:
well, it's not a big deal, just do another PR and push normally, like a usual bugfix.

The Qucs branches should *never* be force-pushed (IIRC on GitHub force push can be disabled, we discussed this some time ago), once something it's there it stays there.

ra3xdh:
@andresmmera , I have just reverted merge at #665. 

@in3otd Yes, the force push is unsafe, and must not be used. I have found an instruction on proper way for unmerging https://stackoverflow.com/questions/6481575/undo-a-merge-by-pull-request .

andresmmera:
Ok, thank you Vadim. I'll try to work on this PR during the weekend

in3otd:
ok, but IMHO there is no need to revert the merge; this should likely be done just for stuff that we really do not want in the code and was pushed by mistake.
The description at StackOverflow talks about *Someone accepted a pull request which they shouldn't have. Now we have a bunch of broken code*, seems not quit the case here. The stripline code from andresmmera is not broken AFAIU it could just be improved as all the code can be :grin: .

PR #662, opened 2017-02-01
Fix warnings qt3support

### LINKED ISSUE #656: Fix deprecation warnings (Qt3Support)
There is a truckload of deprecation warnings when `-DQT3_SUPPORT_WARNINGS` is enabled.

See log in #653 as an example.

guitorri:
Closes #656.


felix-salfelder:
truly awesome. please merge (regardless of micro-regressions).

guitorri:
For some reason some deprecated constructor warning are not showing on my system (they show up on Windows and Linux builds).
Using `-DQT_DEPRECATED_WARNINGS` seems to do the trick. I will push more fixes shortly.

in3otd:
just saw this here:
- add a diagram (any)
- note that you cannot exit the "place diagram" mode anymore, the diagram outline remains attached to the cursor . I used to press `Esc` for this but now it does not work

Works fine in develop.

guitorri:
`Esc` no longer release any of the items picked from the dock. Looking into it.

guitorri:
9262f75 is to blame. It is deprecated for menus, but not for actions. I will send a fix later.

guitorri:
Rebased, reverted the regression on pressing `Esc`. Silenced a few more warnings. 
There are a few still warnings remaining. I will try to clear them tomorrow.

guitorri:
still a few things to fix

guitorri:
All warnings should be fixed. Shall we enable `-Werror` from now on?
I believe I tested everything. Please take a look.

guitorri:
Bump. 
I think we should merge this before we start merging new features.

in3otd:
lemme do a last check :grin:
About enabling `-Werror` I'm not sure but we could try and revert in case we don't like it...

in3otd:
...it seems `clang++` on Travis (linux) does not like some `Qt4` stuff and emits [some warnings](https://travis-ci.org/Qucs/qucs/jobs/199742953#L3373) (?)

guitorri:
I am afraid there is not much we can do about these warnings. Qt4 is no longer supported on that version of OSX/Clang.  I will try to have a look, if I find a fix I will let you know in another PR. 😄

in3otd:
I think I saw the warnings for the `Qt4` in Linux actually... (but I do not see them here locally)

Instead, I just saw here I have also
```
graphictextdialog.cpp:34:67: warning: unused parameter 'name' [-Wunused-parameter]
savedialog.cpp:34:65: warning: unused parameter 'modal' [-Wunused-parameter]
```
besides a ton of `clang-3.8: warning: -ldl: 'linker' input unused` (as usual here), which is maybe just due to the not up-to-date `clang (?)

I'm starting to think that enabling `-Werror` is maybe too much...

felix-salfelder:
`-Werror` *is* too much as a default build option (with or without debugging) simply because future compilers might warn about stuff that we cannot know yet.

yet, it might be pretty useful so set `-Werror` for CI, at least until we hit a warning that we absolutely do not agree with (or have no clue whatsoever). `make CXXFLAGS=-Wall` ...

guitorri:
Right, we are not ready for  `-Werror`. Even for CI it would create too much noise at this stage. Perhaps before a release we can make some sanitation until we see things are more or less under control.
I thing the solution to `warning: -ldl: 'linker' input unused` is to not pass the flag everywhere.
The `warning: unused parameter` are not Qt3Support issues. We should open another issue.

in3otd:
yeah, as usual noticed a small issue just after merging: when opening a project I get a message on the console 
```
Warning: QLayout: Attempting to add QLayout "" to SaveDialog "", which already has a layout
```
everything seems to work, anyway.

PR #665, opened 2017-02-03
Revert merging of #661

ra3xdh:
This reverts the merge of #661, because there exists some work on stripline done in parallell. @andresmmera , #in3otd , Please look.


in3otd:
as said in #661, I see no need to revert the previous commits, it will just make the history more messy: added stuff ->removed stuff ->re-added similar stuff. Just pushing a patch to the existing code will be fine. There is nothing fundamentally bad in the original PR.
If for some reason, andresmmera ends up with a completely new code for the microstrip this *might* make sense, but still...

andresmmera:
I don't think that the code should suffer deep changes... By the way, do
you have better/more reliable sources for the stripline model?
I use Mantaro's calculator for a number of things and it seems to work
properly... but anyway, feel free to suggest any other tool for checking
the results.

ra3xdh:
@andresmmera, I have understood from the discussion that you are intending to rewrite the code from the scratch. Then I opened this PR. If you decided that only little changes are necessary, you may close this PR and open correcting PR.

in3otd:
> By the way, do you have better/more reliable sources for the stripline model?

I think all the stripline code out there is based on the work of [S. B. Cohn](http://ieeexplore.ieee.org/stamp/stamp.jsp?tp=&arnumber=7379077), done some 50 years ago or so...

andresmmera:
@ra3xdh Well, my intention is to fix the issues described by in3otd, but I think that rewritng the component code from scratch is not needed. Anyway, I'll leave this PR open until I have a clear picture of how to face the problem.

andresmmera:
Closed. See #667

PR #666, opened 2017-02-06
#339 rebased (New matching techniques for the matching tool)

in3otd:
here is #339 rebased to develop. As explained there, I had to squash all the commits to understand how to fix the conflicts. I also removed the whitespace changes done on existing code and undone moving some of the `#include` to `matchdialog.h`; my understanding is that we should try to have only forward declarations in the heders file when possible and including the definitions only in the `.cpp` files.

Did not yet check much, please check if it behaves as expected.

I saw anyway an error: by using the default values in the Matching Tool (just open it and press *Create*) the circuit created by the previous version and this one differs:

Previous version:
image

This PR version (was the same before rebasing):
image

If you define an *Equation Defined RF device* with some S-parameters to match you will see that this PR version of the matching tool draws the Port 2 circuit (on the right side) in the reversed order...

Besides, would be nice to keep the components values with units prefixes instead of the engineering notation and having the component text for vertical components positioned correctly.

Feel free to push corrections to this branch :grin:


andresmmera:
@in3otd @guitorri I created a [number of schematics](https://gist.github.com/andresmmera/28f7f4bf69d0313d841e50ba267e3c41) for testing the new matching tool.
Should I put them at qucs-test?

guitorri:
There are a few LaTex errors like:

```
! LaTeX Error: File `SingleStubOpen.eps' not found.

 ...                                              
                                                  
l.256 ...egraphics[width=50mm]{SingleStubOpen.eps}
```

I think it always nice to have the examples. These will not test the matching tool itself, but will serve as a reference.

in3otd:
BTW, it seems that `qucs-doc` cannot be built correctly out of tree: `dvips` is not able to find the `.eps` figures (but continues anyway)

guitorri:
About `.eps`, I believe that if you do not give the extension LaTex (pdflatex in our case) will take care of the conversion. We have other `.eps` files in use... it should work.

guitorri:
Try removing the extension as in:
```
./tutorial/textmode/content.tex:	\includegraphics[angle=0,width=0.4\linewidth]{rc_ac_circuit}
...
./tutorial/textmode/pics/Makefile.am:  rc_ac_circuit.eps \
```

in3otd:
omitting the extension is the Right Thing but when I build out of tree I get some error messages:

```
$ make 2>&1 > /dev/null | grep "Could not find figure file"

/usr/share/texmf/bin/dvips: Could not find figure file TLpiece.eps; continuing
/usr/share/texmf/bin/dvips: Could not find figure file CYexample.eps; continuing
/usr/share/texmf/bin/dvips: Could not find figure file MNAexample.eps; continuing
/usr/share/texmf/bin/dvips: Could not find figure file NLexample.eps; continuing
/usr/share/texmf/bin/dvips: Could not find figure file MNAnoise1.eps; continuing
/usr/share/texmf/bin/dvips: Could not find figure file MNAnoise2.eps; continuing
/usr/share/texmf/bin/dvips: Could not find figure file Unoise.eps; continuing
/usr/share/texmf/bin/dvips: Could not find figure file Real_r.eps; continuing
/usr/share/texmf/bin/dvips: Could not find figure file Real_c.eps; continuing
/usr/share/texmf/bin/dvips: Could not find figure file Real_ec.eps; continuing
/usr/share/texmf/bin/dvips: Could not find figure file QuarterWavelengthFilter.eps; continuing
/usr/share/texmf/bin/dvips: Could not find figure file SingleStubOpen.eps; continuing
/usr/share/texmf/bin/dvips: Could not find figure file SingleStubShort.eps; continuing
/usr/share/texmf/bin/dvips: Could not find figure file DoubleStubOpen.eps; continuing
/usr/share/texmf/bin/dvips: Could not find figure file DoubleStubShort.eps; continuing
/usr/share/texmf/bin/dvips: Could not find figure file MNAsingular.eps; continuing
```

not sure that building the docs out of tree ever worked, actually.

in3otd:
@andresmmera I see that if matching cannot be done (e.g. when getting the message *It is not possible to match this load using the double stub method*) pasting from the clipboard is activated anyway, so if the clipboard contained parts previously copied there they are proposed to be pasted on the schematic.
This is because `slotEditPaste()` is [always called](https://github.com/in3otd/qucs/blob/b10f74335e212ba38acb4e2e0b574cf92e3e781f/qucs/qucs/dialogs/matchdialog.cpp#L749) even if the matching cannot be done.
 
What's the difference between
```
calcMatchingCircuitLC()
calcMatchingCircuitSingleStub()
calcMatchingCircuitDoubleStub()
calcMatchingCircuitCascadedLambda4()
calcMatchingCircuitCascadedLCSections()
calcMatchingCircuitLambda8Lambda4()
```
I have the impression they all do the same thing, they just pass the parameters to the appropriate ladder code building routine and then they all call `SchematicParser()` and `CreateSchematic()`.
I think that in [`calcMatchingCircuit()`](https://github.com/in3otd/qucs/blob/b10f74335e212ba38acb4e2e0b574cf92e3e781f/qucs/qucs/dialogs/matchdialog.cpp#L931) we could just call the appropriate ladder code building routine and than have a common routine that calls `SchematicParser()` and `CreateSchematic()`, similarly to what is done in [`calcBiMatch()`](https://github.com/in3otd/qucs/blob/b10f74335e212ba38acb4e2e0b574cf92e3e781f/qucs/qucs/dialogs/matchdialog.cpp#L1034).
I'll take a look at this and other minor stuff.

andresmmera:
Indeed they do basically the same... Surely the code ended up like this after moving code back and forth a year ago or so... I'll try to put that code in _calcMatchingCircuit_, where it should be.

Thanks for the code review :+1:

in3otd:
no problem, the code works, that's just to make it easier to read in the future :grin:
If you have the time to reorganize the code that's fine, I was thinking to do that also for some other minor things, like
- `CreateSchematic()` always return `true` and in any case its return value is not used by the various `calcMatchingCircuit*`, so you can simply make them all return `void`. I think that the string returned by the various ladder code building routines (`calc*()`) should be checked and consider an empty string as an error so do not enable pasting from clipboard.
- `calc2PortMatch()` also partially builds the schematic, which is not so nice, this should all be done by `SchematicParser()`; it receives also the `schcode` so it can know what needs to be done, also for two-port matching circuits. `schcode` can use a flag telling if it's a two- or single-port matching, something like `QUCS_TWO_PORT_MATCH | QUCS_S_PARAM_SIM`. One could maybe use [scoped enums](http://www.ibm.com/developerworks/rational/library/scoped-enums/index.html) for this.
II think i this way you can avoid passing the `components`, `wires` and `paintings` string and just receive back a single string containing all the schematic directly from `SchematicParser()`. 
- the widget in the matching dialog sometimes use too little or too much space,  making things difficult to read. The problem here is that every OS/window manager renders widgets in a slightly different way; at least check that when enlarging/shrinking the dialog the widgets resize/move in a nice way, this is usually enough to make it look good on other systems.
Especially when the substrate parameters are shown things get sometimes difficult to read.

andresmmera:
I have just made a few changes so as to tackle the issues pointed out by @in3otd 

* The _calcMatchingCircuit*_ functions were removed.
*   _SchematicParser_ function now treats the S-param simulation block, the ports and the labels as the rest of components. This avoids using the _schcode_ flags and reduces the number of arguments.

Pending stuff:
 - [ ] Check GUI behaviour when resizing
-  [ ] Update docs

andresmmera:
I've just used the 'clang-format' command to give a _well defined_ coding style to matchdialog.cpp/h
I'm not sure if you like that...

andresmmera:
I couldn't work much on Qucs this week but, finally, I've updated the docs... 
I'll do a last overall revision tomorrow...
Anyway, I reckon that it's more or less finished and bug-free now.

in3otd:
yes, thanks, I did some testing last week and didn't see any issue. 
Only that having the dialog window non-resizable might be annoying; the dialog looks mostly fine here, but I'm sure there will be some combination of OS/window manager/screen DPI/etc. which will require adjusting the window size for some users.

Building docs fails here now; AppVeyor complains about some `taperedline` stuff (?)

andresmmera:
Uhm... I'll take a look later

andresmmera:
I've just allowed resizing & push the commit. I've also checked that no weird effects happen after resizing (I mean, fields overlapping or stuff like that)
I'm gonna see what's going on with the commit history...

andresmmera:
I've just removed the unwanted commits and corrected a typo (\Ohm instead of \Omega) in the docs.

in3otd:
I have rebased the PR again, I tried without squashing all the commits but there were too many conflicts.

While testing I saw another issue, when the length of an element (stub, line, etc.) was in the meter range this was written as `1.23 m` but this is interpreted by Qucs as 1 *milli* (meter). I have  then modified again `num2str()` to handle this case.

guitorri:
This looks really good to me. 
Is there anything else you want to add/modify?

andresmmera:
As for me, I'm not planning to do anything else here (at least in the short run :-) )

guitorri:
Very well. Thank you. Merged.

PR #672, opened 2017-02-12
Fix QPushButton with color, used for color picking feedback.

### LINKED ISSUE #74: Stylesheet pushbutton break layout
Related to #69.
To make pushbutton with background color work on Mac OS and Windows (even some Linux), I use stylesheet to fill the color in pushbutton. However, this destroy the size of pushbutton and layout, as shown in the photo below.
https://cloud.githubusercontent.com/assets/1388949/4800913/371ce0b6-5e2b-11e4-9086-0532201287d8.png

---

The code in #135 could be applied to:
- Paintings
  - [x] line
  - [x] arrow #135
  - [x] Text
  - [x] filled / Ellipse
  - [x] filled / Rectangle
  - [x] Arc
- Diagrams Properties
  - [x] Data color
  - [x] Properties, Grid color
- File > Application Settings
  - [x] Document Background Color

guitorri:
Close #74.

Based on PR #662.

### DIFF
--- a/qucs/qucs/diagrams/diagramdialog.cpp
+++ b/qucs/qucs/diagrams/diagramdialog.cpp
@@ -213,7 +213,7 @@ DiagramDialog::DiagramDialog(Diagram *d, QWidget *parent, Graph *currentGraph)
   else if(Diag->Name != "Truth") {
     Label1 = new QLabel(tr("Color:"));
     Box2Layout->addWidget(Label1);
-    ColorButt = new QPushButton("   ");
+    ColorButt = new QPushButton("");
     Box2Layout->addWidget(ColorButt);
     ColorButt->setMinimumWidth(50);
     ColorButt->setEnabled(false);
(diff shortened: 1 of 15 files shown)


guitorri:
Oops something wrong here... need to rebase.

guitorri:
No no, it indeed depends on #662.

guitorri:
Before:
screen shot 2017-07-12 at 20 39 33

After:
screen shot 2017-07-12 at 20 35 51

in3otd:
...well the stylesheet solution looked nicer here... when it worked 

Several "color picking" buttons have a text string consisting of a few spaces (`"   "`), this causes the color pixmap there to be out of center, especially when the dialog is resized here. Removing the spaces the behavior seems ok.

Could a similar solution also work for [all the color-picking buttons](https://github.com/Qucs/qucs/blob/master/qucs/qucs/dialogs/qucssettingsdialog.cpp#L155-L238) in `qucssettingsdialog.cpp` ?

guitorri:
Cleaned up the white spaces and pulled all the color setter/getters into misc::
The issue with stylesheets in my system is that the button only accepts clicks on a portion of the area.

guitorri:
Can someone check on Linux o see if this makes sense?

andresmmera:
I can check that tomorrow...

andresmmera:
I've just tested this PR in Ubuntu and I didn't find issues.

However, I've found the following warning:

> graphictextdialog.cpp:34:67: warning: unused parameter ‘name’ [-Wunused-parameter]
 GraphicTextDialog::GraphicTextDialog(QWidget *parent, const char *name)
                                                                   ^
cc1plus: warning: unrecognized command line option ‘-Wno-deprecated-register’

https://github.com/guitorri/qucs/blob/9b6f495fb46d0fa383841607ba82b1f14cee20d2/qucs/qucs/paintings/graphictextdialog.cpp#L34

It was here before this PR, but, now you've opened this PR, I guess it would be a good idea to remove that...

I think the PR works as expected. I tried to resize the window with different aspect ratios and it always triggers the color picker regardless the area you click inside the button. Moreover, I didn't see any warning in the standard output when clicking the button, so it seems to be ok :+1: 

diagramdialog_pr672

guitorri:
Thank you. I removed the unused variable you pointed and another one i found along the way.
Merging.

PR #675, opened 2017-02-14
Update qucs test

guitorri:
Hopefully qucs-test issues are fixed for OSX.


guitorri:
OSX failing again. qucs and qucsator checks are passing. qucs-equations is failing. It seems that qucsator is crashing.
I tried a few runs on my system. I will try a few more runs...

PR #680, opened 2017-03-18
Added documentation for the rectangular and circular waveguides

andresmmera:
In this PR it is added some documentation for the circular and rectangular waveguide components


ra3xdh:
I have read this documentation addition and found no errors. Merging.

Issue #683, opened 2017-04-27
Ctrl-LMB for selecting multiple objects does not work


in3otd:
in `develop` pressing `Ctrl` to select multiple things with the left mouse button does not work anymore, only the last clicked object remains selected, as if `Ctrl` was not pressed.
Works fine in `'master`; I guess this bug may have been introduced when fixing the Qt3 warnings stuff but did not have time to bisect yet.

in3otd:
fixed in #684

PR #684, opened 2017-05-06
Small gui stuff

in3otd:
a couple of small fixes,
one for #683,
the other for the warning mentioned (with solution, thanks!) by andres [here](https://github.com/Qucs/qucs/commit/2e9f7d1373be75542373764397430bef411e6f9c#commitcomment-21969297)

### DIFF
--- a/qucs/qucs/dialogs/savedialog.cpp
+++ b/qucs/qucs/dialogs/savedialog.cpp
@@ -65,7 +65,7 @@ void SaveDialog::initDialog()
    checkBoxLayout->addWidget(fileView);
    SaveDialogLayout->addWidget(group);
    
-   buttonsLayout = new QHBoxLayout(this);
+   buttonsLayout = new QHBoxLayout();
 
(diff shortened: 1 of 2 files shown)


gildias:
Hello, I think this PR might have introduced a behavior that is possibly not intended.
I see you commented that the logic is now reversed, it may be related to it.

While my historical knowledge of QUCS is limited, as far as I recall, dragging the component with the left mouse button would result in it "snapping" to the grid. It moved a fixed number of units and it made the process of attaching wires and components together really simple.

Now, that doesn't happen. By dragging a component with the left mouse, it enters a state of free movement, it may be useful in some situations but, perhaps, it shouldn't be the default setting.
The way to achieve the "snap" behavior now is by dragging with left mouse + holding control.

Again, if it was not working as intended before and this just fixed it, feel free to dismiss this message. However, as someone who has been working with simulators for a decent time, I find that moving components around is a bit counter-intuitive now.

in3otd:
yes, I saw the mistake just after pushing the related PR and thought to fix it soon... but I realize now that 2 months passed already

PR #688, opened 2017-06-03
Qucsactivefilter: Legendre filter implementation

ra3xdh:
One of my students @gypsysattva implemented Legendre filter for Qucsactivfilter. Filter implementation follows this guideline http://www.crbond.com/papers/lopt.pdf . Select `ApproximationType->Legendre` to get access to this filter. Here is an example of synthesized L-filter.
lfilt


in3otd:
nice, thanks! Tested a little and seems to work ok. The higher order filters may have the cutoff frequency a little off, maybe the poles in the table should have more precision?
Would be good to keep the doc in sync and add a short chapter for this type of filter too.

ra3xdh:
> maybe the poles in the table should have more precision?
In theory, the poles estimation precision may influence on filter cutoff frequency. The same effect should also be observed for Bessel filter, because it also uses table of poles. Further investigation are needed.

> maybe the poles in the table should have more precision?
Yes, I will add a chapter on Legendre filter in Qucs-activefilter documentation later this week.

ra3xdh:
I have just added a note on Legendre filter synthesis at `qucs-doc/report/activefilter` manual.

qucs-doc/report/activefilter/content.tex
```diff
@@ -856,6 +856,92 @@ \subsection{Poles and zeros}
...
+\begin{equation}
+ H(s^2)=\frac{1}{1-s^2-3s^4-3s^6}
+\end{equation}
+\item Calculate roots of denominator. Roots of denominator are the poles of 
+transfer function. 
+\item Synthesize filter circuitry by known transfer function poles. 
+\end{enumerate}
+
+Qucsactivefilter application uses predefined poles table of Legendre filter. 
+The \verb|bessel.h| header file contains poles table. This table is computed in
```
in3otd: here it should be `legendre.h` in stead of `bessel.h` - I'll correct this before merging

in3otd:
yep, also for the Bessel filter the cutoff frequency for the higher orders is not precise. For Butterworth filter is fine also for high orders. BTW, while testing I noticed that for this latter the variable `K` in the created schematic is not in `dB` as for the Bessel and Legendre (did not check the others).

in3otd:
looking at the code I see that actually is the Sallen–Key topology that defines `K=(out.v/in.v)` in the generated schematic equation, while the multiple feedback topology defines `K=dB(out.v/in.v)`. It's ok for you if I add the `dB()` also to the  Sallen–Key topology so the same equation is used for `K` for all the circuits?

qucs-doc/report/activefilter/content.tex
ra3xdh: I have just corrected this typo. Then I squash documentation-related commits before merging.

ra3xdh:
Yes, auto-generated expression for magnitude response could be unified. I have corrected it in the recent commit.

PR #682, opened 2017-04-20
Addition of the Wattmeter and Ohmmeter Modules

gildias:
Hello, my name is Alberto Silva and I'll be continuing the module development started by previous students from my institution (ISEP, Portugal).
This commit is related to the watt probe. Since this is my first interaction with github, I'd appreciate some information on whether I did the pull process correctly, and if not, what needs to be changed.
I'll likely be sending some more modules over the next weeks so I would like to know if I should keep developing over my existing files ( with the wattmeter in them ) or if I should start over for each module.
This is important since some modules will share modified files, with the obvious example of a makefile.
I have also added a simple pdf with 3 examples of the wattmeter working, hope it was attached correctly.
Thank you for your time.
Edit: I did not know you would receive the comment I made when i originally pushed to my git branch, thought this was the commit message, sorry about that.

[Wattmeter examples.pdf](https://github.com/Qucs/qucs/files/945095/Wattmeter.examples.pdf)

Issue #211, opened 2015-04-03
Sweep component with a single value and Type=list gives wrong results


in3otd:
As [discussed on the qucs-help mailing list](https://sourceforge.net/p/qucs/mailman/message/33655321/), using a Sweep component with a single value and `Type=list` gives wrong results; 
A simple test circuit is [here](https://gist.github.com/in3otd/bd7174ca4bd1d99c4098) .
Looking at the code, the problem comes from [here](https://github.com/Qucs/qucs/blob/master/qucs-core/src/sweep.cpp#L143), where, if the list has just one component, `getPropertyDouble()` is called regardless of the list type (`list` or `constant`). A simple fix will be to check the list type, but anyway the code needs to be verified, since, as far as I understood looking at the [commit history](https://github.com/Qucs/qucs/commit/db3d8a54554961c5dfd64d6e1c1749287edfed4b) the special behaviour for sweeps having just one value was done to allow _"parameter sweep variables or other variable in the 'Values' property (when it is marked 'const') of simulation boxes"_  and this does not seems to work either. Or maybe I misunderstood what was the purpose of that modification.

in3otd:
fixed by #691. Still unclear what the original code tried to do, though.

PR #691, opened 2017-07-03
Fix for AC and SP sweep

andresmmera:
It was found a bug in the AC solver while testing #682. The 'freq' variable ('acsolver.cpp', line 100) skips the first value of 'swp'. So if the sweep is performed over a list of length 1, swp->next() always returns 0.

The same logic applies to the SP solver.

P.S.: I've accidentally opened a another PR...

### DIFF
--- a/qucs-core/src/sweep.cpp
+++ b/qucs-core/src/sweep.cpp
@@ -140,8 +140,6 @@ void sweep::reverse (void) {
 nr_double_t sweep::next (void) {
   nr_double_t res = data[counter];
   if (++counter >= size) counter = 0;
-  if (size == 1)
-    return parent->getPropertyDouble ("Values");
   return res;
 }
 


in3otd:
? I've tried to reproduce this bug here but as far as I can see even the current code works well. Do you have an example that is showing the issue? Shouldn't a fix, if needed, go in [`sweep::next()`](https://github.com/Qucs/qucs/blob/develop/qucs-core/src/sweep.cpp#L140)?

gildias:
I can't say I tested it very thoroughly, but it does solve the issue I was having with my PR.
It couldn't properly test a single frequency value over a "list" simulation, it is working correctly now.
This did not happen with multiple frequencies, so it does point to a bug where the first frequency was skipped, if it was the only one.

in3otd:
ok, now it's clearer when this happens, it seems almost the same as #211 , IIRC at that time I was not sure about any side effects of a fix and thought that using *constant* instead of *list* for a single value as Sweep Type was "good enough". As pointed out there, the fix should likely be in `sweep()` instead of the code using it.

andresmmera:
Hello,
Sorry for the delay...

1) Here's the schematic to reproduce the bug: https://gist.github.com/andresmmera/5ef793b59ba1e8198c5dcc0f87ebff95

The 'develop' code gives the following result...
screenshot from 2017-07-04 19-33-11

...whereas this PR gives:
screenshot from 2017-07-04 19-39-05

2) Concerning the solution, you can either use 'swp->get(index)' instead of 'swp->next()' at acsolver.cpp/spsolver.cpp or delete the lines 143/144 [here](https://github.com/Qucs/qucs/blob/master/qucs-core/src/sweep.cpp#L143). The reason why I've chosen the first option is that I don't really know what lines 143/144 are supposed to do. I'll undo the fix soon.

in3otd:
thanks, after realizing that this is the same as #211 I did several test and still it seems to me that [those lines](https://github.com/Qucs/qucs/blob/develop/qucs-core/src/sweep.cpp#L143-L144) in `sweep::next()` don't do anything useful; all the simulations I tried with *constant* and *list* sweeps types, with or without variables in the sweep behave the same - except of course for the case of a list with a single element, which works only if you remove those lines.

andresmmera:
I've just removed lines 143/144 and pushed

PR #693, opened 2017-07-12
N-port S-parameters de-embedding

in3otd:
As mentioned some time ago, I've added an S-parameters de-embedding component. The code seems to work ok, PR is WIP as I may move some code around/rename some files/etc.; the simulator code is based on the S-parameters component and  retains the "variable number of ports" feature even if only two-port de-embedding is supported currently, hoping that we may have 4 or more ports supported in the future.
On the other end the GUI code for this component, while still derived from the S-parameters component was put in a separate file as it seemed easier to expand and/or add new features (ports reordering, maybe?) in the future there.

Simple test schematic is here

[de-embedding_example.zip](https://github.com/Qucs/qucs/files/1143281/de-embedding_example.zip)

with a screenshot

de-embedding

(and yes, maybe also a little documentation should be added)


in3otd:
Now N-port de-embedding is supported, after realizing there was a not-so-complex general formula for the *anti-network* S-parameters; I've added some theory to the docs.
Ports reordering is not included but should not be difficult to add one day, if needed - I do not plan any further work on this PR.

Here is a test schematic for the 4-port de-embedding 
[de-embedding_4_port_example.zip](https://github.com/Qucs/qucs/files/1522650/de-embedding_4_port_example.zip)

andresmmera:
Hello, I've been testing and reviewing this code during the weekend. My comments

First of all, hat off, it works nicely for 2 ports and 4 port spar and IMHO this is a key feature for those who work extensively with spar data :+1: 

- 2 ports

selection_025

- 4 ports

selection_026

So far so good. No problems at all, but...

- The N>4 case is the problem I'm facing now. I've noticed that sometimes qucsator fails calculating the inverse of S although S is full rank => |S| != 0. I mean, for some frequencies[ this assert](https://github.com/Qucs/qucs/blob/develop/qucs-core/src/math/matrix.cpp#L819) is triggered

In order to simplify the problem, I'm focusing on 300 MHz, where the inverse seems to work and I'm observing this:

selection_027

where CoupledInductors3.s6p is the spar file of the inductors in the picture above. It was taken from the following simulation:

selection_028

Here you can find the project I'm using for testing this PR
[de-embedding_TEST_prj.zip](https://github.com/Qucs/qucs/files/1524678/de-embedding_TEST_prj.zip)

in3otd:
um, possibly something wrong with port reordering - e.g. I think the increments [here](https://github.com/Qucs/qucs/pull/693/commits/0e69a9a951629ff9f3906e2ed301fc80c5f523a2#diff-213315595fc59d051d686670489dc111R123) and [here](https://github.com/Qucs/qucs/pull/693/commits/0e69a9a951629ff9f3906e2ed301fc80c5f523a2#diff-213315595fc59d051d686670489dc111R148) should not be there but if I remove them, then even the 4-port case does not work

in3otd:
I've done a correction to the port reordering part, seems to work but didn't have the time to check much...

OTOH , I thought it should not be too difficult to automatically generate netlists to de-embed a random N-Port, defined using the RFEDD, using python, to check for the higher number of ports cases also :grin:

andresmmera:
Uhm, it doesn't work

selection_029

in3otd:
oh, well, I thought port reordering was the *easy* part :grin:
Might be fixed now...

(BTW, while preparing the python script described above I found a bug in `qucsconv`, I'll open an issue later)

andresmmera:
Yes, I confirm it works! :+1: 

selection_030

in3otd:
thanks for testing. I changed again that part of the code as the old one was not always working properly with a higher number of ports 😐 . The code is also more clear now, IMHO.
Now I've tested up to 40 ports, with a python script that generated the test netlist and S-parameters file automatically, and all seem to be ok.
Could you please also verify again on your side? I'll squash the latest commits afterwards.

andresmmera:
Sure! Out of curiosity, are you planning to integrate that script with qucs-test?

andresmmera:
I've just tested your last commit for the N={2,4,6,8} cases and I confirm it works :smiley: 

(I accidentally wrote this message at PR#370, sorry for the noise. I've already removed that...)

in3otd:
I've squashed the corrections, seems to still work fine here :grin:

The test script was actually written for GNU Octave, I've placed it [here](https://gist.github.com/in3otd/b70cd82609e7c3cefed77f130c717247#file-test_de_embedding-m) in case you'd like to play with that. Not sure if it makes sense to have it in `qucs-test`, even if we should have there some test(s) for this new component also.

andresmmera:
I've extended your script up to 100 ports (although, yes, N=2 is the most common case) and I didn't find issues.
selection_032


I consider this PR (and PR #758 ) can be merged soon (maybe next week?)

Issue #633, opened 2016-12-17
Unexpected Duplicate Schematic Behaviour


nouwaarom:
Steps to reproduce:
* create a new project
* create a  schematic
* make a simple circuit
* simulate the circuit
* add some diagrams in the data display
* duplicate the schematic file (using right mouse button -> duplicate in the content menu)

Expected:
the data display file and dataset corresponding to the original schematic to be duplicated too

Actual:
both the original and the duplicated schematic use the original data display file and dataset

in3otd:
personally, I'm used to that behavior but I see it can be confusing at first, also because it's not consistent with what "Save as" does.
It may make sense to do what "Save as" does also when duplicating the schematic, i.e. using the new (base)name also for the default Dataset and Data Display of the duplicated schematic.

I'm not in favor of having "Duplicate" create other files also because it should apply only on the selected file, IMHO; also the Data Displays and Datasets can be duplicated manually if one wants/needs to.

cvgaviao:
It would at least to change the names of files in "document settings" to the new name that was choose.

in3otd:
I thought (a little) about that and I think the current solution is the best compromise: using "Save As..." changes the names of files in "Document Settings", while the operations available in the drop-down menu from the mouse right button in the project content view operate at the file-level only, so without changing anything else. We should of course document the behavior...

ra3xdh:
See also https://github.com/ra3xdh/qucs/issues/26 for the same issue report. Maybe it's better to emulate a `SaveAs` operation in `QucsApp::slotCMenuCopy()` instead of simple copying of file? Current behavior when file is just copied  may mislead user.

in3otd:
to expand a little bit on my previous comment, we currently have two functionalities
- the *Duplicate* in the *Content* tab context menu:
  this just copies the schematic file, it does not modify the copy in any way; the file content is the same as the original file, only the name is different. The schematic still points to the original Dataset and Data Display files
- the "Save as..." menu item in *File*:
  this can be used to save an existing schematic with a different name **and** it will change the schematic Dataset and Data Display names to use the same basename as the new schematic file

I'm reluctant to change any of these because I actually use both: I may have several different version/refinements of a schematic, which I create incrementally using the *Duplicate* menu, so I still have one Data Display will all the graphs and markers/etc. I need and I use *Save as...* to create a separate schematic from some existing one I have around.

[...]

PR #695, opened 2017-07-13
Fix of #633: Part 1

ra3xdh:
Here is a first part of fix for #633. I renamed "Duplicate" to "Copy file". This name is more relevant to actual function of this menu item. The update of documentation will follow soon.


ra3xdh:
See https://github.com/Qucs/qucs-help/pull/15 for documentation.

Issue #697, opened 2017-07-14
Build Error: qrc_qucs.cpp: No such file or directory


aviadb:
Building qucs from latest "develop" branch
fails with error
g++: error: qrc_qucs.cpp: No such file or directory
g++: fatal error: no input files
compilation terminated.
make[4]: *** [Makefile:641: qrc_qucs.o] Error 1

in3otd:
cannot reproduce here - also the automatic builds on Travis and Appveyor were fine.
Which OS and build system (CMake / autotools) are you using?

If you do `rcc -v` do you have some meaningful output? Do you happen to have an `rcc-qt4` or similar on your system also?

If you used `./configure` could you check what was printed exactly on the line saying 
```
checking for rcc...
```

aviadb:
Manjaro linux
Cmake 3.8.2
rcc is not available on my system as executable.
checking for rcc ... no.

Searching for rcc with my package manager gives the following results:
librcc 0.2.12 - Charset conversion library
pyside-tools-common 0.2.15 - PySide lupdate, rcc, and uic development tools
pyside-tools-common 0.2.15 - for python 3
pyside-tools-common 0.2.15 - for python 2

nothing related to qt

in3otd:
so the issue comes from the fact that the `rcc` program, which comes with Qt, is not found, quite strange, it should be available somewhere.
I forgot to ask also for the output of the `configure` lines
```
checking for moc-qt4...
checking for moc...
checking for uic-qt4...
checking for uic...
```
did you check if `rcc-qt4` works? (e.g. `rcc-qt4 -v`) Googling around it seems that Manjaro uses that name for the Qt4 `rcc`, which would explain the issue: for some reason we do not check if `rcc` is available with that name (but do so for other Qt utilities)

aviadb:
checking for moc-qt4... /usr/bin/moc-qt4
no line for checking for moc...
checking for uic-qt4... /usr/bin/uic-qt4
no line for checking for uic...

rcc-qt4 -v
prints: Qt Resource Compiler version 4.8.7

:S

in3otd:
thanks, this confirms that the issue comes from our `configure` script, which [looks only for `rcc`](https://github.com/Qucs/qucs/blob/master/qucs/configure.ac#L710) and not also for `rcc-qt4`, as done for the other Qt utilities. I'll push a fix later.

felix-salfelder:
maybe plain

[...]

PR #699, opened 2017-07-18
RCC and build rules

felix-salfelder:
see #697


in3otd:
Travis fails with an unexpected [error message](https://travis-ci.org/Qucs/qucs/jobs/254811205#L3821); restarted it still fails there. But passes on OS X.
Interestingly on this latter `rcc` is called as
`usr/local/bin/rcc -o qucs_.cpp qucs_.qrc`
while on Linux as
`/usr/bin/rcc -o qucs_.cpp ../../../qucs/qucs/qucs_.qrc`

felix-salfelder:
`/usr/bin/rcc -o qucs_.cpp ../../../qucs/qucs/qucs_.qrc` looks perfectly fine to me.

on my machine, it's not complaining...
rcc bug? i have rcc 5.7.1

(don't know whats wrong on OSx though.)

in3otd:
ah, the difference is that on OS X we run `make` (`qucscheck`) and on Linux `make distcheck` instead. And this latter fails also locally here, with the same error message. But no idea why...

BTW, I thought/assumed that `rcc 5.x` was for Qt5.

What about naming the generated file `qucs.qrc.cpp` similarly to what is done with the `.moc` files?

felix-salfelder:
> rcc 5.x was for Qt5.

perhaps it's backwards compatible? the buggy file lookup does not look related to the qt version to me.

> qucs.qrc.cpp

sounds good.

in3otd:
...it's seems we opened another small to Pandora's box: the `make distcheck` are failing because the components bitmaps are not distributed in the tarball (there was no need to, as we distributed the output of `rcc`) so the `.qrc` processing fails since the files to process are not there. 
Then we have other `.qrc` files for the Tools in the sub-directories which would need to be handled in the same way as the main app resource file.

I'm ok to just add the `rcc-qt4` to `configure.ac` for the moment but feel free to handle also the stuff above if you feel so inclined.

felix-salfelder:
> there was no need to

hmm, i think the bitmaps qualify as source, and should be included.

thanks for figuring that out, i was definitely on the wrong track

qucs/qucs/Makefile.am
```diff
@@ -100,15 +104,16 @@ EXTRA_DIST = \
 	bitmaps/qucsrescodes.icns \
 	bitmaps/qucstrans.icns
 
-SUFFIXES = .h .moc.cpp .qrc.cpp
+SUFFIXES = .h .moc.cpp .qrc
 
-.qrc.cpp.cpp:
-qucs.qrc.cpp: qucs.qrc
+.qrc.cpp:
```
in3otd: what is this for? Doesn't the *Suffix rules with no recipe are also meaningless* (unlike pattern rules) from the docs apply also here?

felix-salfelder: my understanding is, that `.qrc.cpp:` instructs automake to use the path `.qrc -> .cpp -> .o`, where applicable.

cf https://www.gnu.org/software/automake/manual/html_node/Suffixes.html

in3otd:
thanks,
what shall we do with the other `.qrc` files
```
./qucs/qucs-transcalc/qucstrans.qrc
./qucs/qucs-attenuator/qucsattenuator.qrc
./qucs/qucs-activefilter/qucsactivefilter.qrc
./qucs/qucs-filter-v2/qucsfilter.qrc
./qucs/qucs-lib/qucslib.qrc
./qucs/qucs-rescodes/qucsrescodes.qrc
./qucs/qucs-filter/qucsfilter.qrc
```
at a quick glance also there we distribute the compiled file - but also the source `.qrc` file, without the source bitmaps... seems a bit weird to me.

in3otd:
thanks, just a couple of details: in `qucs/stuff.am` you have the GPL blurb set to version 3. Could that file be included also by `qucs/qucs/Makefile.am`, as done for the other makefiles?

in3otd:
one more thing, we would need an updated rule(s) in `.gitignore` for the compiled resource files, we currently have a line `qrc_*.cpp` which of course does not work anymore

in3otd:
looks good, do you plan to squash/rebase/something before merging?

in3otd:
I realized we forgot to update the `CMakefiles` with the new `.qrc` files names, so I've pushed a commit to fix that. It builds fine here with `cmake` now - I'll try later, in another PR, to add a build using `cmake` to Travis.

felix-salfelder:
thanks. did'nt think about that.

i once thought that appveyor runs the cmake system. if that's not the case, why do we need cmake then?

in3otd:
don't remember if we used `cmake` for AppVeyor in the past but currently it also uses the autotools.
I think the `cmake` build was added at some point in the past as a possible alternative and, looking at the old commits comments, because the cmakefiles can be used in Qt Creator but I don't know much more.

felix-salfelder:
in general, id prefer to reduce the clutter.

if someone (here: you) maintains the cmakefiles, i do not object `:)`

in3otd:
yep, don't know how much `cmake` is actually used to build Qucs - adding a `cmake` build to Travis is an attempt to shift the maintenance burden to every developer :grin:

BTW, I'll merge this PR shortly, unless there are any objections...

felix-salfelder:
> shift the maintenance burden to every developer

let's discuss how to get rid of this in another place. it's likely not too urgent for 0.20.

PR #701, opened 2017-07-22
Admittance/impedance marker data management at Smith chart diagrams

andresmmera:
So far, when placing a marker at a Smith chart diagram (both impedance and admittance types), the markers always show S-par and impedance data. In this sense, this PR fixes this and includes a new option in marker dialog so as to let the user decide which of these parameters (impedance, admittance, both or none) wants to see.

selection_001

The default behavior is to show impedance data at conventional Smith charts and admittance data at  admittance Smith chart diagrams.

It was added two extra fields ('display admittance' and 'display impedance') to the format employed to save the marker settings on the .dpl file.

I'll try to implement this feature in #407 soon. After that, if #407 gets merged, this PR can be closed.

### DIFF
--- a/qucs/qucs/diagrams/diagram.cpp
+++ b/qucs/qucs/diagrams/diagram.cpp
@@ -383,9 +383,11 @@ int Diagram::regionCode(float x, float y) const
 // Is virtual. This one is for round diagrams only.
 bool Diagram::insideDiagram(float x, float y) const
 {
-  float R = float(x2)/2.0 + 1.0; // +1 seems better (graph sometimes little outside)
+  float R = x2/2.0;
   x -= R;
   y -= R;
+  R += 1.0; // +1 seems better ? (allow graph to go a little outside)
+  //qDebug() << "insideDiagram" << x << y << R << (x*x + y*y- R*R);
   return ((x*x + y*y) <= R*R);
 }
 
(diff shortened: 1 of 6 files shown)


andresmmera:
I've tested 428b687 and it's fine.
admittancestuff

in3otd:
I've added the options to show also the series and parallel equivalent components values: 

export

andresmmera:
Reviewed and tested locally. It works fine. 
eqcktfeature

However, I think I've found a bug... When the marker is about to cross the inductive/capacitive border it jumps to the center of the Smith chart, it stays there for a while and then it continues following the curve trace (it's quite weird...). Please, see the following screencast:

https://drive.google.com/open?id=0BwDjrwGfd_z7SFYxUHYtNFBrdVU

Notice that the marker reading is correct though... 
This issue is not related to this PR, but it also happens at the master branch...
I'll try to see what's going on here...

in3otd:
thanks for the fast review, while testing I saw the same issue and fixed it after pushing to this PR. I was not sure if it should go to a separate PR, it's just a small thing [here](https://github.com/Qucs/qucs/blob/develop/qucs/qucs/diagrams/diagram.cpp#L384); the `+1` should have been added later to `R`, if at all. I'll push the change so you can take a look.

PR #704, opened 2017-07-30
Fix polar() function

andresmmera:
It was found a bug in the polar() function while trying to help with this issue https://sourceforge.net/p/qucs/discussion/311050/thread/005906e4/?limit=25#e2b9

The line which converts the argument from polar to Cartesian coordinates is wrong:
 r·exp(j*theta) =  r·exp(nr_complex_t(0, Re{theta}))                r, theta \in R

### DIFF
--- a/qucs-core/src/math/complex.cpp
+++ b/qucs-core/src/math/complex.cpp
@@ -550,7 +550,7 @@ nr_complex_t limexp (const nr_complex_t z)
 */
 nr_complex_t polar (const nr_double_t mag, const nr_double_t ang )
 {
-#ifdef HAVE_CXX_COMPLEX_POLAR_COMPLEX
+#ifdef HAVE_CXX_COMPLEX_POLAR
     return std::polar (mag, ang);
 #else
     return nr_complex_t (mag * cos (ang), mag * sin (ang));
--- a/qucs-core/src/math/complex.cpp
+++ b/qucs-core/src/math/complex.cpp
@@ -568,7 +568,7 @@ nr_complex_t polar (const nr_complex_t a, const nr_complex_t p)
 #ifdef HAVE_CXX_COMPLEX_POLAR_COMPLEX
     return std::polar (a, p);
 #else
-    return a * exp (nr_complex_t (imag (p),-real (p)));
+    return a * exp(nr_complex_t(0.0, 1.0) * p);
 #endif
 }
 


in3otd:
interesting... 
`polar(double, double)` is correctly defined and should have been called, IMHO, for the schematic in the discussion quoted above since I would expect `mag()` and `rad2deg()` to return a `double` and not a `complex`. But instead the buggy `polar(complex, complex)` function is used.

This is different from what I see with a slightly simpler example:
export
note that forcing one of the variables to be of `complex` type triggers the bug.

Probably related to how the `polar()` function is [defined](https://github.com/Qucs/qucs/blob/develop/qucs-core/src/evaluate.cpp#L3519-L3584) for different arguments.

Anyway, the correct formula should be a little more complex (pun intended), something like
```
a * exp(nr_complex_t(0.0, 1.0) * p)
```
since your correction works for real `p` only and the [Qucs docs](http://qucs.sourceforge.net/docs/tutorial/functions.pdf) say that `polar()` works for complex arguments also.
Other fixes are needed also in [evaluate.cpp](https://github.com/Qucs/qucs/blob/develop/qucs-core/src/evaluate.cpp), will take a look later.


Unrelated to this bug, note that [this line](https://github.com/Qucs/qucs/blob/develop/qucs-core/src/math/complex.cpp#L553) should be `#ifdef HAVE_CXX_COMPLEX_POLAR` since it's about `complex polar(double, double)`, which is tested by [this macro](https://github.com/Qucs/qucs/blob/develop/qucs-core/m4/complexcxx.m4#L161-L183)

andresmmera:
> Anyway, the correct formula should be a little more complex (pun intended), something like

> a * exp(nr_complex_t(0.0, 1.0) * p)

> since your correction works for real p only and the Qucs docs say that polar() works for complex arguments also.

Does it really make sense to have a *complex module* and a *complex angle*? In my opinion, both should be real...

In the formula you suggested
> a * exp(nr_complex_t(0.0, 1.0) * p)

the real part is:

image

whereas the imaginary part is:

image

To be honest, I cannot grasp the physical meaning of these results... **My guess** is that polar(nr_complex_t, nr_complex_t) was implemented to do the polar2cartesian conversion for *real valued* nr_complex_t variables... I'll try to do some research in this sense

in3otd:
well, I also do not have an intuitive understanding of what exactly taking a complex number to a complex power means but as said the Qucs docs say that `polar()` works with complex arguments so either we modify the code or the docs but both need to agree. And since all the code is already there and the fix seems not so complex (...) I propose to fix the code. BTW, part of the fix is just moving the minus sign, from `return a * exp (nr_complex_t (imag (p),-real (p)));` to `return a * exp (nr_complex_t (-imag (p),real (p)));`, which is equivalent to the formula above.

andresmmera:
Don't take me wrong, I completely see your point. 
Let's apply the fix you propose.
Anyway, (someday...) I'll try to see where and why polar(nr_complex_t, nr_complex_t) is used...

andresmmera:
I've just pushed the correction you suggested. I've also included an additional commit with the fix for the #ifdef at polar(nr_double_t, nr_double_t)

felix-salfelder:
line 571 looks sensible to me now.

i don't understand what happens if or when HAVE_CXX_COMPLEX_POLAR_COMPLEX is defined. it would be useful to know that the behaviour will be similar in that case...

> Anyway, (someday...)

see how in gnucap "untested();" is placed in parts of the code that nobody has managed to test or otherwise verify. put "unreachable();" if you think the control should never reach that point (which i do).

try to figure the qucs code segmented in unreachable and untested parts, and how that would facilitate refactoring...

in3otd:
> Anyway, (someday...) I'll try to see where and why polar(nr_complex_t, nr_complex_t) is used...

I too was curious to know so I took a look around: IIUC all the data vectors use complex elements (see [here](https://github.com/Qucs/qucs/blob/develop/qucs-core/src/vector.h#L239), so every function must be able to operate on complex arguments, even if the result may actually make sense only for real numbers. This is likely why `polar(complex, complex)` was defined in the first place and is then called by the schematic above.

An even simpler test schematic showing the bug:
export

> i don't understand what happens if or when HAVE_CXX_COMPLEX_POLAR_COMPLEX is defined.

I was wondering which compiler/library has `polar(complex, complex) `, I didn't find it anywhere

guitorri:
I also thing the functions with complex arguments is a form of overload to allow all argument types to be processed by the equation system.
Merging.

PR #706, opened 2017-08-05
Ubuntu Trusty environment on Travis

in3otd:
Just because Travis warns that it will switch to Ubuntu Trusty (14.04 LTS) soon, wanted to check that everything still build fine there

Edit: added caching and a `cmake` build while I was at it

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,39 +15,58 @@
 # - doxygen build (master only)
 
 language: cpp
-
 dist: trusty
-
 sudo: required
-
-os:
-  - linux
-  - osx
-
-compiler:
-  - gcc
-  - clang
 
 python:
   - "2.7"
+
 virtualenv:
   system_site_packages: true
+
+# packages required by all Linux builds
+#   additional packages are defined in the matrix, depending on needs
+addons:
+  apt:
+    packages: &native_deps
+      - gperf
+      - latex2html
+      - octave
 
 # Qt 4.8.7 official installer works up to OSX 10.10 or xcode6.4
 # Use rebuilt (with Qt3Support) homebrew bottle for OSX 10.12 or xcode8.2
 osx_image: xcode8.2
 
 matrix:
-  exclude:
-    - os: osx
+  include:
+    - os: linux
       compiler: gcc
+      addons:
+        apt:
+          packages:
+            - *native_deps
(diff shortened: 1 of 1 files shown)


in3otd:
I was hoping to improve the build time, or at least not to increase it too much by using the container-based environment on Travis but actually thing got quite worse, except maybe on OS X:

with the sudo-enabled environment used so far:
image

and now using the container-based environment:
image

It's true that now the builds in the container-based environment start in a couple of seconds (for the Linux builds) instead of 20 seconds or so but still this is not so helpful.

in3otd:
with `ccache` enabled, but cache empty, and `export MAKEFLAGS="-j2"` (since it seems we have 2 cores available)

image

and then restarting the build - so this time the cache for `ccache` is filled
image

in3otd:
added a build with `cmake also`, here are the build times on a rebuild with warm cache:

image

I don't plan to add anything else for the moment :grin:

guitorri:
Nice. 
I still don't get why the Xcode is even mentioned on the Linux instances.

in3otd:
yep, I was also puzzled by that but didn't bother to find out why, up to now :grin:

I have now moved the `osx_image` variable assignment in the OS X section in the matrix and the description in Travis is fine now.

image

The `linux` builds above took longer since the cache was invalidated by the "change" of the environment.

BTW, it seems now that with a container-based build we should be able to run and debug an [image locally](https://docs.travis-ci.com/user/common-build-problems/#Running-a-Container-Based-Docker-Image-Locally) - nice option to have but did not try that yet.

BTW2, I don't remember, what was the reason for having the secure variables definition in the `.travis.yml` file, having the docs/help built and uploaded somewhere automatically?

guitorri:
Really nice. I didn't know one could run the container localy. There is PR related to Docker which is a similar idea.
 
As for your BTW2, the secure variables were once used to push to <https://github.com/Qucs/qucs-doxygen>.
It was removed in bf2168e7a6.

in3otd:
was that removed by mistake? does it make sense to restore uploading the `doxygen` docs and the Coveralls update?

guitorri:
Aff, Travis seems to be having issues with Trusty: https://github.com/travis-ci/travis-ci/issues/8449

PR #558, opened 2016-09-08
Shortcut manager

andresmmera:
Hello, 

This PR is a rebase of the [shortcut manager](https://github.com/yodalee/qucs/tree/feature54-custom-shortcut) written by yodalee. Personally, I find this feature quite useful so I'd like to see it merged someday :)

Cheers

---
rebased from yodalee feature54-custom-shortcut

PR #708, opened 2017-08-13
Add context menu to documents tabs

in3otd:
since I sometimes open a lot of tabs, e.g. while looking for a particular schematic/feature, I needed a convenient way of closing all the tabs currently open *except* the one of interest.
I've then added a context menu to the tabs, like  this

image

I've also added a *Close all* action and, why not, a simple *Close* for the current tab.

Works fine here, I do not actually plan big modifications to the code, just maybe a little cleanup to `qucs.cpp`.

### DIFF
--- a/qucs/qucs/qucs.cpp
+++ b/qucs/qucs/qucs.cpp
@@ -2979,6 +2979,9 @@ void ContextMenuTabWidget::showContextMenu(const QPoint& point)
   APPEND_MENU(ActionCxMenuCloseOthers, slotCxMenuCloseLeft, "Close all to the left")
   APPEND_MENU(ActionCxMenuCloseOthers, slotCxMenuCloseRight, "Close all to the right")
   APPEND_MENU(ActionCxMenuCloseAll, slotCxMenuCloseAll, "Close all")
+  menu.addSeparator();
+  APPEND_MENU(ActionCxMenuCopyPath, slotCxMenuCopyPath, "Copy full path")
+  APPEND_MENU(ActionCxMenuOpenFolder, slotCxMenuOpenFolder, "Open containing folder")
 #undef APPEND_MENU
 
     menu.exec(tabBar()->mapToGlobal(point));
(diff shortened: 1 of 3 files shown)


guitorri:
What about an option to "Open file location in File browser" or something like that.
Often I want to jump to project directory to copy figures or take a look at the files and I never know which of my multiple "Qucs Home" locations is currently set. 
Similar action could also go to the Project and Contents dock.
Let me know, if it makes more sense I will open an issue/feature request.

in3otd:
yes, an "Open file location in File browser" will make sense (also in the Project View), I also do need to copy stuff from a project directory from time to time. And BTW, I also have multiple "Qucs home" locations and find frustrating that I cannot navigate these within the Project View, but this latter will be for (much) later.
Would be nice also to be able to select multiple tabs and have a "Keep only selected tabs" menu but I don't think that `QTabWidget()` allows selecting more than one tab.

Coming back to this PR, the cleanup I mentioned above was about some code duplication I think there is in `qucs.cpp`, nothing urgent.

andresmmera:
Just a suggestion for improvement: I really like to put the diagrams in the schematic sheet, so more often than not I deactivate the option "Open data display after simulation" under the "File->Document Settings" menu. Would it make sense to add this option to the tab context menu?

in3otd:
um, I also do this relatively often but I don't think that enabling/disabling "Open data display after simulation" should go in a context menu about opening/closing Tabs/Documents.
Using `Ctrl-.` takes a little less than opening the menus with the mouse :grin:

Since this choice is usually done at file creation it should be selectable there; I think we discussed in the past having Schematics templates, so under the menu *File* we could have *New*, *New w/o Data Display* and *New from template* or something like that (keeping the current keyboard shortcuts).

in3otd:
I've added "Open containing folder", "Copy full path" and other items to the tab context menu.
Also updated the application menu to have the context menu options there also (makes easier to find them, IMHO) and to be able to create a new Document with "Open Data Display after simulation" unchecked.
Maybe I overdid it :grin:

image

image

image

andresmmera:
> and to be able to create a new Document with "Open Data Display after simulation" unchecked.

Thank you :+1: 

I haven't tested the latest commit yet... but what about having a shortcut for creating a new stand-alone schematic (maybe shift+N)?

in3otd:
yep, I was not sure if I should add shortcuts for all the new stuff but I too thought of adding the a shortcut for the stand-alone schematic. I think Ctrl-Shift-N might be more in line with the other menu shortcuts, all start with Ctrl and Shift is sometimes a modifier to the Ctrl- action.

andresmmera:
I agree. This reminds me that #558 is still there. I'll be back with it after 0.0.20 :grin:

in3otd:
I've added a shortcut for creating a new stand-alone schematic but Ctrl+Shift+N was already used by the "New Project" action, so I used Ctrl+Shift+H since it seemed one of the few not-already-in-use shortcuts.
I've also tried to remove some code which was duplicated around, please check for any added bug.

andresmmera:
I've done a quick check and I didn't find issues.

Things I've checked:
- [x] No additional warnings at compile.
- [x] Both autotools and CMake builds work.
- [x] All the context menu options work as expected.
- [x] No weird runtime warnings when using the new features.

I've also taken a look at the last commit. It looks good :+1:

Issue #710, opened 2017-09-09
Qucsator crashes on trivial netlists using glibc-2.26


pienjo:
Hello,

Qucsator (from unmodified qucs-0.0.19 sources) suddenly started crashing on my system (void linux, Linux 4.12.10, x86_64, glibc 2.26), with a segmentation fault in qucs::net::containsAnalysis(), line 224. It doesn't really matter what circuit I feed it; my crash dump below is using the supplied netlist, consisting of a voltage source and a resistor.

[netlist.txt](https://github.com/Qucs/qucs/files/1289880/netlist.txt)

Output:

[martijnb@tinkerbell .libs]$ ./qucsator -i ~/.qucs/netlist.txt 
project location: 
modules to load: 0
factorycreate.size() is 0
factorycreate has registered:
parsing netlist...
checking netlist...
netlist content
      1 DC instances
      1 R instances
      1 Vdc instances
creating netlist...
Segmentation fault
[martijnb@tinkerbell .libs]$ 

gdb stack trace:

#0  0x00007ffff78b855a in std::list<qucs::analysis*, std::allocator<qucs::analysis*> >::begin (this=0x21)
    at /usr/include/c++/6.3/bits/stl_list.h:841
#1  0x00007ffff78b6fb4 in qucs::net::containsAnalysis (this=0x61ff10, child=0x6214e0, type=1) at net.cpp:224
#2  0x00007ffff78b7870 in qucs::net::sortChildAnalyses (this=0x61ff10, parent=0x6226e0) at net.cpp:355
#3  0x00007ffff78b7768 in qucs::net::orderAnalysis (this=0x61ff10) at net.cpp:344
#4  0x00007ffff78b7144 in qucs::net::runAnalysis (this=0x61ff10, err=@0x7fffffffe61c: 0) at net.cpp:249
#5  0x00000000004049a5 in main (argc=3, argv=0x7fffffffe808) at ucs.cpp:251

Relevant source:

220     /* Looks recursively for a type of analysis. */                                                  
221     int net::containsAnalysis (analysis * child, int type) {                                      
222       ptrlist<analysis> * alist = child->getAnalysis ();                                            
223       if(alist != nullptr) {                                                                                          
224         for (auto *a : *alist) {                                                                                    
225           if (a->getType () == type)                                                                          
226             return 1;                                                                                                  
227           else if (a->getType () == ANALYSIS_SWEEP)                                                                         

228             return containsAnalysis (a, type);                                                           
[...]

PR #712, opened 2017-09-09
Avoid crash in net::sortChildAnalyses

pienjo:
Removing an element from a list invalidates any iterators pointing at
these elements. As a result, (indirectly) calling remove() on a list
while it is traversed using a range-based for loop triggers undefined
behaviour.

Rewrite the loop, so no invalidated iterators are referenced anymore.
This is arguably not the neatest solution.

Fixes issue #710

### DIFF
--- a/qucs-core/src/net.cpp
+++ b/qucs-core/src/net.cpp
@@ -350,7 +350,13 @@ void net::orderAnalysis (void) {
 void net::sortChildAnalyses (analysis * parent) {
   ptrlist<analysis> * alist = parent->getAnalysis ();
   if (alist != nullptr) {
-    for (auto *a: *alist) {
+
+    for (auto it = alist->begin(); it != alist->end(); /* empty */) {
+      // Copy the value of the element (a pointer), and advance the
+      // iterator prior to manipulating the list.
+      analysis *a = *it;
+      ++it;
+
       if (a->getType () == ANALYSIS_DC
 	  || containsAnalysis (a, ANALYSIS_DC)) {
 	parent->delAnalysis (a);


felix-salfelder:
> This is arguably not the neatest solution.

still it looks correct. thanks for the spot. lets clean it up a bit:

    auto *a = *it++;

this is hard to read for just

    auto b = it;
    ++it;

and then put

    parent->delAnalysis (*b).

which will also indicate the possible bug in delAnalysis (what if two subsequent items in parent compare equal to *b?).

i would suggest to remove the additional comments from the code, as they do not explain anything.
* 'Move' in used in the comment block which is not followed by an instruction. i can't see anything being moved here.
* "avoid range based loop" + reason. correct, but commit message, not code.

pienjo:
```
auto b = it;
++it;
[...]
parent->delAnalysis(*b);
parent->addAnalysis(*b);

```
wouldn't solve the problem - 'b' would still be an iterator, so delAnalysis(*b) would still leave b in an undefined state - the dereference of 'b' for the call to addAnalysis would still be undefined. 

```
auto b = *it;
++it;
[...]
parent->delAnalysis(b);
parent->addAnalysis(b);
```
*would* fix it (because 'b' is no longer an iterator, but just a naked pointer). This is one of the reasons why I personally dislike using auto - it hides too much. In regards to two subsequent items comparing to *b: this is exactly why I don't think this is the best solution - delAnalysis() would delete them both, but only a single copy would be inserted at the head of the list. If this were my code, I'd rewrite it so it wouldn't use std::list::remove(analysis *) (via delAnalysis) but std::list::erase() on the iterator, so only a single element gets moved at the time, but I wanted to keep the patch as constrained as possible:)

That said, I don't think this is really an issue. The items in the list are pointers to objects, and from what I could gather the list actually *owns* these objects. If there are multiple list entries pointing to the same object, you're in much, much deeper problems than just this...

I'll rewrite my patch.

felix-salfelder:
thanks for the explanation.

> If there are multiple list entries pointing to the same object,

in order to remove something unique from a known list position, why would one not pass an iterator to a delete function? what is the point of calling *parent functions over just modifying alist?

anyway, your refreshed patch looks good enough. i'm wondering if adding a comment between the lines of "TODO: rewrite" or "FIXME: need to do $this", or even "WARNING: fragile method, do not use in future code" would be adequate. or just merge...?

Issue #711, opened 2017-09-09
qucs-0.0.19 fails to build - configure failed while checking for Qt library


1greyghost:
(This is on a new install of Linux Mint 18.2)

I did the following: 

ADMS
- downloaded and unpacked adms-2.3.6
- sudo apt-get install build-essential
- sudo apt-get install automake libtool gperf flex bison
- sudo apt-get install libxml2 libxml2-dev
- sudo apt-get install libgd-perl
- sudo sh bootstrap.sh
- sudo ./configure --enable-maintainer-mode
- sudo make install

Great adms made!!

QUCS
- downloaded and unpacked qucs
- sudo apt-get install libqt4-dev libqt4-qt3support
- sudo apt-get install texlive-font-utils octave-epstk
- installed qt4-dev-tools via synaptic 
- sudo ./configure

Bummer, the configure failed.  below are the last few lines of the configure process:

    checking for X... no
    checking for Qt headers... found in /usr/include/qt4 -I/usr/include/qt4/Qt -I/usr/include/qt4/QtGui -I/usr/include/qt4/QtCore -I/usr/include/qt4/QtSvg -I/usr/include/qt4/QtXml -I/usr/include/qt4/QtScript -I/usr/include/qt4/Qt3Support
    checking for Qt... 4 (multi-threaded)
    checking for Qt library... configure: error: not found
    configure: error: ./configure failed for qucs

Any thoughts?

felix-salfelder:
checking for X... no

looks strange to me. no X on Mint?

    checking for Qt library... configure: error: not found

so you did instal some Qt libraries? which/how?
configure should have written more details to config.log,  can you show us the relevant chunk?

1greyghost:
felix, below is the config.log:

This file contains any messages produced by compilers while
running configure, to aid debugging if configure makes a mistake.

It was created by qucs configure 0.0.19, which was
generated by GNU Autoconf 2.69.  Invocation command line was

  $ ./configure 

## --------- ##
## Platform. ##
## --------- ##

hostname = Elfworks
uname -m = x86_64
uname -r = 4.10.0-33-generic
uname -s = Linux
uname -v = #37~16.04.1-Ubuntu SMP Fri Aug 11 14:07:24 UTC 2017

/usr/bin/uname -p = unknown
/bin/uname -X     = unknown

/bin/arch              = unknown
/usr/bin/arch -k       = unknown
/usr/convex/getsysinfo = unknown
[...]

PR #713, opened 2017-09-18
configure, qt

felix-salfelder:
cleaning up... removes a bogus library check, which might (or might not) be the reason for #711.

generally, configure.ac is a bit historic. should be redone, such as
* implement header/library checks using `AC_*`,
* proper (optional) `--with-<feature>` flags,
* remove undocumented hacks, such as `QTDIR`.

or (easier, more reliable) setting system specific variables (here `LDFLAGS`) should be left to the user/maintainer.


guitorri:
I keep wondering if the `X11` things are still needed. 

Anyway, it passes CI, merging.

PR #715, opened 2017-10-05
Fix "export as image" with non integer scale factor

in3otd:
When setting a non-integer scaling factor for the exported image, Qucs silently fails to create the image.
This is because the scaled image dimensions are written as floating points numbers in the Width/Height line edits and then wrongly converted to zero by `QString.toInt()`.
The `QIntValidator()` on those line edits is not used when setting the contents using `setText()` as pointed out in the [docs](http://doc.qt.io/qt-4.8/qlineedit.html).

Steps to reproduce:
- open a schematic
- *File*->*Export as image...*
- uncheck *Original size*
- enter a floating point *Scale factor*, check that *Width in pixels* and/or *Height in pixels* contains a floating point number instead of an integer
- click *Export*
- verify that no image is created

### DIFF
--- a/qucs/qucs/dialogs/exportdialog.cpp
+++ b/qucs/qucs/dialogs/exportdialog.cpp
@@ -200,8 +200,8 @@ void ExportDialog::calcWidth()
 {
     if (cbRatio->isChecked()) {
         float h = editResolutionY->text().toFloat();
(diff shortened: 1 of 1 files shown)


guitorri:
I will review this one.

PR #720, opened 2017-10-06
Netlist fix

felix-salfelder:
cleanup. might have side effects, let's see


in3otd:
uh? The Travis build with `clang` complains *ERROR: files left in build directory after distclean:
./qucs/qucs/qucs_.cpp* but not the other builds ?? This seems about another PR we already merged...
I've restarted the build but it still stops with the same error.

felix-salfelder:
.. not one of the side effects i had expected `:|`

guitorri:
Maybe because your first PR was against `master`?

felix-salfelder:
strange. let's see

guitorri:
Included in #749.

PR #721, opened 2017-10-07
Add to CMake files used for dynamic load

guitorri:
- add files to the project
- add DL libraries at link time


guitorri:
This is a no-op as CMake is not running on CI. Merging.
I will merge #706 just after.

PR #723, opened 2017-10-07
Replaced QLineedit by QComboBox for frequency input (for phasor and AC temporal diagram dialogs)

andresmmera:
The aim of this PR is to replace the QLineEdit by a QComboBox as the frequency input widget at the dialog window in the phasor and AC temporal diagrams. This way, the QCombo shows only the available frequencies.

In order to do that, a function named 'LoadAvailableFreqs()' has been added to diagramdialog constructor. The purpose of this function is to explore the content of the 'acfrequency' field of the current dataset and, finally, it returns that data as a QStringList which is used to feed the QComboBox.

'LoadAvailableFreqs()' is called at the constructor but also when the click() event of the 'ChooseData' combo is triggered.

diagramdialogcombo

[Screencast here](https://drive.google.com/open?id=0BwDjrwGfd_z7ck03TzZXMHdiR2s)

### DIFF
--- a/qucs/qucs/diagrams/diagramdialog.cpp
+++ b/qucs/qucs/diagrams/diagramdialog.cpp
@@ -37,7 +37,6 @@
 #include <QLineEdit>
 #include <QCheckBox>
 #include <QSlider>
-#include <QComboBox>
 #include <QListWidget>
 #include <QTableWidget>
 #include <QPainter>
@@ -46,6 +45,7 @@
 #include <QHeaderView>
 #include <QDir>
 #include <QDebug>
+#include <QComboBox>
 
 
 #define CROSS3D_SIZE   30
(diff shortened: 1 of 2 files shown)


guitorri:
Looking good. 
Do you have a youtube channel? You could put your screencasts online to showcase new features 😄

andresmmera:
lol, no. I'm just doing that to prove myself I checked things.
Feel free to use these videos if you want :-)

in3otd:
@andresmmera , yep, I also thought you should put those screencasts to Youtube :grin:

andresmmera:
Seriously?
hahaha, this could be the beginning of the official Qucs Youtube channel :joy:

guitorri:
Actually, these diagrams are supposed to show results for multiple frequencies at the same time.
So instead of a QComboBox, we need perhaps a QListWidget from which we can select one or more available frequencies.
There are a few more things that need fixing. I will add this one to the list.

PR #725, opened 2017-10-08
travis: use pip install --user

guitorri:
Trusty is not using sudo. Install modules into user home.


guitorri:
Hum, LINUX+GCC+Autotools is failing:
```ImportError: No module named _tkinter, please install the python-tk package```

PR #727, opened 2017-10-08
Refactor main, add unit test

guitorri:
- Adds basic support for unit test for the Qt based code.
 - Use QtTest in conjunction with CMake for the moment.
 - Include a few initial test units.
 - Create new shared library libqucsschematic
 - Rename libqucs to libqucsator

Hopefully this can be used to help refactoring while keeping the codebase in check.


felix-salfelder:
looks useful. thank you!

guitorri:
For some reason travis cannot find/run the tests. Looking into it.

guitorri:
Hum. Linux+GCC are hitting undefined references.
https://travis-ci.org/Qucs/qucs/jobs/285206888#L4073-L4089

guitorri:
Nope, order is still wrong: https://travis-ci.org/Qucs/qucs/jobs/285234918#L3068

guitorri:
Rebased to fix issues around redefinition of Pi (from powercombiningtool)

guitorri:
It runs. The unit test for Component and Schematic needs work. It seems we have some cyclic dependencies going on...
Please don't merge (yet).

guitorri:
Go figure. Linking works with Linux-clang but fail with Linux-gcc...

felix-salfelder:
i have started to cleanup the build a bit and switch to libtool libraries in #589. would it make sense to backport that?

guitorri:
Yes, I thing the way to go is to create a shared library. 
@felix-salfelder can you cherry pick what is needed for libtool from the PR you mentioned (there is quite a lot going on there).

felix-salfelder:
yes, will have a look. that PR needs this kind of cleanup anyway.

regarding the cmake build: can you make sure the "library" remains private? it might be counterproductive if somebody really starts using that...

guitorri:
What do you mean by private?

guitorri:
I see, yes we can make it private:
https://cmake.org/cmake/help/v3.0/command/target_link_libraries.html

felix-salfelder:
i think it will be sufficient to not install them with other libraries. e.g. in $prefix/lib/$package, rather than $prefix/lib, so the linker will not find them. (also don't install any headers, but i guess that was never your intent...)

guitorri:
My intent right now is to be able to run unit tests. I am working on finishing the QGraphics stuff, but so far all the testing is manual, not good.
In the future, if ever get to the point of having a simple API, we can thing about making the library available to other people.

On OSX/CMake it works works, but the Linux/CMake does not.
I need to setup again a Linux box/virtual machine. Something like `-fPIC` or `-Wl,--export-all-symbols` needs to be passed on Linux... I will try again tomorrow.

felix-salfelder:
you need `-fPIC` on linux, certainly. with extra `-DPIC` you are on the safe side. `--export-all-syms` must be something windows specific, maybe a globally enforced `__declspec(dllexport)` (but never seen nor used that).

felix-salfelder:
the libtool switch is in #731. but it only switches to libtool, doesn't move more stuff into libraries. i think we want/need both.

guitorri:
Yes, IMHO it makes sense to put everything into shared libraries. Maybe one big shared lib for now. 
I think you called it `libqucs`. Note that `qucs-core/scr/Makefile.am` also creates a `libqucs` ... (There is too many qucs things around). 
I called it `libschematic` with with the CMake build, no big deal at this moment, but we need to settle to something clearer.

guitorri:
@felix-salfelder  I'll cherry-pick from your changes soon.

felix-salfelder:
> Maybe one big shared lib for now.

my idea was to only let library code into `libqucs`, removing stuff from a library is much harder than adding... i did not touch the other "libraries", but some (finally: all) of them should be plugins.

qucsator also ships a libqucs (thanks for pointing that out), i don't have a solution to this, perhaps it should be libqucsator instead.

guitorri:
Shall we settle for now with `libqucsschematic` and `libqucsator` ?

As we do the modularity and plugin refactoring passes we dissect `libqucsschematic` as needed.  

I am looking at your path for making the first private.

felix-salfelder:
> Shall we settle for now with libqucsschematic and libqucsator ?

i'm fine with this, i will introduce libqucs as soon i get to introducing the dispatcher (maybe in verilog_nl). the dispatcher must be core.

thanks

felix-salfelder:
ready to go, isntit?

guitorri:
For me it is ready to go.

It seems I misunderstood what you meant by private library.

In #731, you want to install the library into a private location, right?
Something like `/usr/local/lib/qucs/libqucsschematic.so` ?

Based on this [cmake feature](https://cmake.org/pipermail/cmake/2016-May/063400.html), I thought it was about hiding stuff from the public api.

felix-salfelder:
maybe there are different shades of "private". i would guess that "private location" does the trick for us. you won't be able to accidently link to it, and that's all we need. (frankly, i never heard of PRIVATE...)

will have a look and probably merge

PR #728, opened 2017-10-08
am tests (WIP)

felix-salfelder:
try to run guilhermes tests from autotools


felix-salfelder:
will rebase, after #731.

felix-salfelder:
this is solved in #898

PR #729, opened 2017-10-08
cleanup build

felix-salfelder:
- remove stray rules
- don't override default compiler
- move interface tests to more sensible place


in3otd:
no problem, inline comments can be seen also as normal comment in the issue/PR.

Comment above was mainly a reminder to whoever first open a PR to `qucs-core` to include that "fix" also.

I guess `clang` was first because the original code author liked it better - may be the same reason why the autotools defaults to `gcc` when both compilers are installed :grin:

guitorri:
Back on the day, putting clang first was my way of working around a crippled gcc on OSX.
Feel fee to remove, today I know better how to override the compiler.
Open a issue, if nobody fixes it now, it will be forgotten again.

PR #730, opened 2017-10-10
fix PATH use

### LINKED ISSUE #663: Check if is needed to set PATH for processes
The following snipped is used in 6 places:
```
QProcessEnvironment env = QProcessEnvironment::systemEnvironment();
env.insert("PATH", env.value("PATH") );
Program->setProcessEnvironment(env);
```
It tries to set the environment of the calling process for the process to be executed.

Most (?) processes are started with an absolute path to the executables, this should not be needed.
Needs to be checked.

felix-salfelder:
PATH is en environment variable, lets not mess with it. fixes #663

### DIFF
--- a/qucs/qucs/dialogs/simmessage.cpp
+++ b/qucs/qucs/dialogs/simmessage.cpp
@@ -594,9 +594,6 @@ void SimMessage::startSimulator()
   QString sep(":");
 #endif
 
-  // append process PATH
-  // insert Qucs bin dir, so ASCO can find qucsator
-  env.insert("PATH", env.value("PATH") + sep + QucsSettings.BinDir );
   SimProcess.setProcessEnvironment(env);
 
   qDebug() << "Command :" << Program << Arguments.join(" ");
--- a/qucs/qucs/components/spicefile.cpp
+++ b/qucs/qucs/components/spicefile.cpp
@@ -390,7 +390,6 @@ bool SpiceFile::recreateSubNetlist(QString *SpiceFile, QString *FileName)
     }
 
     QProcessEnvironment env = QProcessEnvironment::systemEnvironment();
-    env.insert("PATH", env.value("PATH") );
     SpicePrep->setProcessEnvironment(env);
     SpicePrep->start(script.join(" "));
     //QucsHelp->setCommunication(0);
@@ -440,7 +439,6 @@ bool SpiceFile::recreateSubNetlist(QString *SpiceFile, QString *FileName)
   // startup SPICE conversion process
   QucsConv = new QProcess(this);
   QProcessEnvironment env = QProcessEnvironment::systemEnvironment();
-  env.insert("PATH", env.value("PATH") );
   QucsConv->setProcessEnvironment(env);
 
   qDebug() << "SpiceFile::recreateSubNetlist :Command:" << prog << com.join(" ");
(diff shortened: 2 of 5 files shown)


qucs/qucs/dialogs/simmessage.cpp
```diff
@@ -594,9 +594,6 @@ void SimMessage::startSimulator()
   QString sep(":");
 #endif
 
-  // append process PATH
-  // insert Qucs bin dir, so ASCO can find qucsator
-  env.insert("PATH", env.value("PATH") + sep + QucsSettings.BinDir );
   SimProcess.setProcessEnvironment(env);
```
in3otd: now `asco` cannot find `qucsator` anymore. What should be the Right Thing? I have several `qucs` versions installed, none is in the path and of course I need `asco` to run the `qucsator` associated with the Qucs version from which is started.

qucs/qucs/dialogs/simmessage.cpp
```diff
@@ -594,9 +594,6 @@ void SimMessage::startSimulator()
   QString sep(":");
 #endif
 
-  // append process PATH
-  // insert Qucs bin dir, so ASCO can find qucsator
-  env.insert("PATH", env.value("PATH") + sep + QucsSettings.BinDir );
```
in3otd: oh, I thought I added a comment but cannot see it anymore.

Now `asco` cannot find `qucsator` anymore. I have several versions of Qucs that I use in parallel, none is in the path. How can we make sure `asco` finds and run the right `qucsator`?

felix-salfelder:
i am assuming that the fix introduced here is the reason for the problem becoming visible.

here are four possible solutions
- install asco, so you can run it like
    $ asco
- set you path variable, so the above works. type `export PATH=/path/to/asco:$PATH`
- just run qucs with `PATH` set the way you need it
    $ PATH=/see/above:$PATH qucs
- set `QUCSATOR_ASCO` to `/full/path/to/asco` (check if that's implemented properly)

maybe there's something in qucssettings? (if so, better remove it for now)

felix-salfelder:
revisiting...
- what is telling asco to run `qucsator` rather than `/path/to/your/current/qucsator`?
- where is `env` initialized, and why is it not what you need/want?

in3otd:
- `asco` just calls `qucsator` using `system()` so it has to be in the `PATH` of its environment; as far as I know, there is no (other) way to tell `asco` to look for `qucsator` in a particular place.
- `env` comes from [here](https://github.com/felix-salfelder/qucs/blob/fe412f6136504ccaf4ff3c08b0e6190b97378fab/qucs/qucs/dialogs/simmessage.cpp#L365) and it's the environment of the Qucs GUI. IMHO, using `env` to tell `asco` where to find `qucsator` is what we need to do.

BTW, I think that the remaining lines like `setProcessEnvironment(env)` where `env` is an unmodified `QProcessEnvironment::systemEnvironment()` are redundant, as the called processes automatically get the same environment of the parent process.

felix-salfelder:
> so it has to be in the PATH of its environment

no, you could wrap asco in a script that hacks the environment accordingly. or just install qucsator, as a workaround.

> lines [..] are redundant.

true. thanks for spotting. i will take care of them in the next refactoring iteration, but anyone is invited `:)`

in3otd:
I don't understand the difference between wrapping `asco` in a script that hacks its environment and hacking its environment via the `QProcess()` used to call it, in both cases we just modify its local environment, or?

felix-salfelder:
the wrapper fixes asco invocation, the Qprocess hack messes with simulator invocation, no matter if asco is involved or not. and no matter whether the modification is required at all. in the end it's a random limitation outside of qucs, and there are several workarounds.

we will introduce an asco driver class (just for asco). maybe in the meantime, we could do

     // if(this_is_about_asco){ // BUG: not possible right now, move later
        env.insert("PATH", env.value("PATH") + sep + QucsSettings.BinDir ); // BUG: must point to qucsator (who cares?)
     // }

but that is equally wrong (see #663).

PR #731, opened 2017-10-10
rearrange libraries

felix-salfelder:
- backport from component cleanup (untested)
- one more library
- switch to shared private libraries
- use libtool

### DIFF
--- a/qucs/qucs/Makefile.am
+++ b/qucs/qucs/Makefile.am
@@ -75,16 +75,22 @@ nodist_libqucsschematic_la_SOURCES = $(MOCFILES)
 
 AM_CPPFLAGS = $(X11_INCLUDES) $(QT_INCLUDES) -I$(top_srcdir)/qucs-lib
 libqucsschematic_la_LDFLAGS = $(X11_LDFLAGS) $(QT_LDFLAGS)
-libqucsschematic_la_LIBADD = components/libcomponents.la diagrams/libdiagrams.la \
-  paintings/libpaintings.la dialogs/libdialogs.la \
+
+# TODO:
+# - parts must go to libqucs
+# - other parts need to be loaded at run time
+libqucsschematic_la_LIBADD = \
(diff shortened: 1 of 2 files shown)


felix-salfelder:
rebases. some parts are already in, such as libtool

felix-salfelder:
now i'm seeing these en/decode string missing symbols, but on appveyor. @guitorri do you remember how to get around these? is there an extra library i need to link to?

guitorri:
Try this:
https://github.com/Qucs/qucs/blob/develop/qucs-core/src/Makefile.am#L45-L46

felix-salfelder:
thanks. right direction.

i guess it's better to only clutter the global namespace with what we need. trying to figure that out.

what is the preferred way to check if this is built on win? i could only find `#ifdef _WIN32`, but that looks like 32 bit, no? background: platform.hpp uses `#ifdef MS_DLL`, but that's not in config.h.

in3otd:
FYI, https://msdn.microsoft.com/en-us/library/b0084kay.aspx :
*<b>_WIN32</b> Defined as 1 when the compilation target is 32-bit ARM, 64-bit ARM, x86, or x64. Otherwise, undefined.*

felix-salfelder:
@in3otd thanks. currently it checks whether `__MINGW32__` is defined, will consider to add `_WIN32`. good to know that `32` does not have any (relevant) meaning.

anyway, are there any differences to expect between mingw (what appveyor does) vs. m$visual compiler (the one you linked to)?

in3otd:
I think that `_WIN32` is always defined in Windows for the compilers we typically use.
[Here](http://nadeausoftware.com/articles/2012/01/c_c_tip_how_use_compiler_predefined_macros_detect_operating_system#WindowswithCygwinPOSIX) is a page with some details on which macros are defined when, not sure if it's up to date.

felix-salfelder:
doesn't do much. but it's ready. will do more in another PR.

PR #737, opened 2017-10-13
rename qucsator library to libqucsator

felix-salfelder:


guitorri:
Cherry-picked into #727.

PR #740, opened 2017-10-14
Toolbar icons not showing.

guitorri:
Link the .qrc resources statically with main.

### DIFF
--- a/qucs/qucs/Makefile.am
+++ b/qucs/qucs/Makefile.am
@@ -67,7 +67,7 @@ libqucsschematic_la_SOURCES= \
   node.cpp element.cpp qucsdoc.cpp wire.cpp mouseactions.cpp   \
   qucs.cpp wirelabel.cpp qucs_init.cpp qucs_actions.cpp            \
   viewpainter.cpp mnemo.cpp schematic.cpp schematic_element.cpp textdoc.cpp \
-  schematic_file.cpp syntax.cpp module.cpp octave_window.cpp qucs_.qrc   \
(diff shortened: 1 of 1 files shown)


felix-salfelder:
i see a similar problem in #731. i don't yet understand the fix (assuming that you do get icons now)

guitorri:
I do get the icons. I was getting them with Cmake, I just tried using a similar incantation in the Makefile.am.
From [here](http://doc.qt.io/qt-5/resources.html), it states _Currently, Qt always stores the data directly in the executable_. Ok the doc is for qt5, but i guess they did not change that behavior.

in3otd:
for Qt4 is the [same](http://doc.qt.io/qt-4.8/resources.html#compiled-in-resources). Anyway I don't understand the issue and its fix :grin:

felix-salfelder:
the fix does not seem to change anything here. need to double check.

in3otd:
here instead this commit brings back the missing icons (tested on Linux only).

felix-salfelder:
short story: it must also work with shared libraries, and it doesnt.

longer story: if the resources are in qucs.real and some other stuff in libqucsschematic, then it does not show symbols. i was expecting that these would just have to be in the same archive. due to some qt limitation, resources *must* be in the main executable... (does somebody understand this?)

so the only way (right now) is give up on libqucsschematic. as it seems, it's only used for testing anyway. i don't mind keep it for that purpose.

felix-salfelder:
here it doesn't. but i have rebased #731 on top of this.

PR #742, opened 2017-10-16
circular library dependencies and plugins

felix-salfelder:
figure out intermediate pre-plugin build


felix-salfelder:
the reason why this does not work on appveyor is, that mingw does not allow unresolved symbols in a library.
that means that (prospective) extension modules must be linked against the library resolving these symbols. right now that would require a weird and possibly platform dependent SUBDIR ordering.

i do now understand better, why gnucap has independent lib, apps (extensions) and main directories `:)`. failing to think of anything better, i suggest to replicate that, more or less. as a first step i would move what now is `libqucsschematic` to a directory called `lib` within the gui source tree. later migrate some stuff to libqucs (in the same directory) to form a/the public api.

i guess that if i do clean moves (100% match), then git will be able to track files in a useful way... any objections?

guitorri:
Can you elaborate a bit on what you are trying to achieve? 
You want each subdir to become a plugin?
Shouldn't we also be looking into how Qt handles/support plugins instead of going the gnucap way?

If git fail to track the files we have a big problem. All the work we have hanging around will be broken. So, for now I object.

felix-salfelder:
> Can you elaborate a bit on what you are trying to achieve?

i would like to build the library (what is now libschematic), then the extensions (what is now lib{diagrams,components,...}) and then build/link the main executable (in that order). not so sure if this is possible at the curent stage, but maybe worth a try.

it might be sufficient to move out the main executable code, and leave the rest where it is, would that be less objectionable?

felix-salfelder:
> You want each subdir to become a plugin?

no. plugins should be things with clear boundaries. such as languages (#316) and components (wip). the subdirs are just something i need to work with.

> Shouldn't we also be looking into how Qt handles/support plugins instead of going the gnucap way?

maybe the Qt way makes sense for plugging in UI stuff, if that exists. it's not on my list right now. everything i want to pull in from gnucap is the dispatcher (see `l_dispatcher.h`). i doubt there is anything similar to that in Qt.

felix-salfelder:
okay, this is working, in particular the dynamic loading on mingw. requirement for #316 (never worked with mingw) and other plugin stuff.

don't know how to fix cmake, please help

andresmmera:
On my side (Ubuntu), it compiles fine but when I execute the qucs binary it crashes leaving the following message:

> qucs: main.cpp:525: int main(int, char**): Assertion `qfl.exists()' failed.
Aborted (core dumped)

felix-salfelder:
thanks.

i am not entirely sure how these qt resources work. i'd suspect that it should all be in qucs_.o, which ends up in libqucsgui.so.

ideas
* what happens if you comment out the assertion? it should be a warning... do you get everything without icons?
* did you rerun bootstrap? some executables have been moved, maybe you are using the old ones (not linked to libqucsgui.so)

NB: i do not see any icons at all (for a while) regardless of the assertion.

andresmmera:
>     did you rerun bootstrap? some executables have been moved, maybe you are using the old ones (not linked to libqucsgui.so)

Ops... I didn't. Now it works.

> NB:  i do not see any icons at all (for a while) regardless of the assertion.

Me neither, but now Qucs doesn't crash, I can start playing :-)
Thanks a lot!

felix-salfelder:
thanks for playing with this, and sorry for the late response.

> [copy] "/qucs/qucs/bitmaps" folder in the "qucs/main/"

this is surprising to me. i thought the resource file (and the compilation thereof) is mainly to make qucs independent of any locations of these images.

regardless:

- these icons do not show up for me anyway (not even in the develop branch, not even after passing the assertions i have added)
- i do not understand the point of this (read: why are icons not just put into `$pkgdatadir/bitmaps`)

it seems, this needs more work, although i did not intend to touch bitmaps in this pr at all. (yikes).

felix-salfelder:
@andresmmera with `QT_SELECT=4` (see #768) i now get icons. but even without the proposed copying.

felix-salfelder:
seems to work now. (still) need help with cmake.

felix-salfelder:
rebased. still working, except cmake

felix-salfelder:
part of refactor branch

Issue #719, opened 2017-10-06
Review side effects of PR #682


guitorri:
TODO:
- [x] in `qucs-core/scr/circuit.h` remove unneeded virtual functions. The `dc/trsolver` can set the state for ohmmeter and the special cases can be handled inside the `ohmmeter::initDC`
  - [x] fix virtual void initDC2 (void) { }//only used in ohmmeter.cpp
  - [x] fix virtual void initDC2 (void) { }//only used in ohmmeter.cpp

- [x] Remove netlist hacking to implement internal resistance, #720.
  Revert addition of internal resistance for vdc, vac, idc, iac, iprobe, vprobe
  The right way to do it would be to pass the `Ri` parameter and the component sets up the series or parallel resitor. The issue is that for a series resistor the vdc needs to allocate an internal node to add the resistor, even if `Ri==0`.  So, instead of 2 nodes for each vdc, it will now need 3. I don't thing we should be doing that. If the user wants, just add the Ri manually.

- [x] Remove netlist hacking for internal resistances of wattmeter, also #720.
  Same thing, pass Riv, Rvv to the wprobe and let it takes care of setting the resistors. Here again, instead of 4 nodes, 5 nodes will be required. If the user is looking for more information it might make sense to pay the price of an extra node. 

- [x]  What about iprobe, vprobe? Shall we "upgrade" them to have internal resistance? No, decided to keep them ideal.

~~- [ ] Concerning AC Phasorial diagram, <Phasor ... > _Unknown diagram_. Implement now a similar fallback mechanis we have for unknown components, allow to skip unknown diagrams on load.~~ Similarly an unknown Paintings need such thing. A refactor on the parser is needed.

- [x] update docs to reflect that phasordiagram modify the <Diagram ... > signature, append the selected frequency 
  Eg. `<Phasor 150 617 217 217 3 #c0c0c0 1 00 1 -687.584 500 687.584 1 -687.584 500 687.584 1 -1 0.5 1 315 0 225 "" "" "" "30 Hz;60 Hz;120 Hz;">` https://github.com/Qucs/qucs-help/pull/16

~~- [ ] Remove `Hz` from the Phasor list of frequencies, see above.~~

- [x] phasor diagram display a list of frequency points, see #723 for initial attempt.

[...]

PR #744, opened 2017-10-18
Revert "Replaced QLineedit by QComboBox for frequency input (for phasor and AC temporal diagram dialogs)"

guitorri:
Reverts Qucs/qucs#723

The original intention was to input and plot several frequencies. The combobox does not allow that.

### DIFF
--- a/qucs/qucs/diagrams/diagramdialog.cpp
+++ b/qucs/qucs/diagrams/diagramdialog.cpp
@@ -37,6 +37,7 @@
 #include <QLineEdit>
 #include <QCheckBox>
 #include <QSlider>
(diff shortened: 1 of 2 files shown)


felix-salfelder:
> if(Diag->Name == "Phasor" || Diag->Name == "Waveac") 

while we are at it. could we get rid of these lines? phasor code needs to go to phasor.cpp, waveac needs to go to waveac.cpp

it reminds me of the wattmeter situation. i'd opt for removing this feature, unless somebody cares to implement it.

guitorri:
Yes, we can get rid of those lines. 
Noted in #719. I will add to my other fixes.

PR #747, opened 2017-10-31
Add smooth() function

in3otd:
I've added a `smooth()` function, similar to what is often used in VNAs to remove some noise from group delay measurements.

Here is an example of the group delay computed from some measured S-parameters, with and without smoothing:

export


guitorri:
Looks good to me.

PR #751, opened 2017-11-18
groupdelay() function

andresmmera:
In order to complement the recently added smooth() function, here it is a function for calculating the group delay from the S parameters.

_groupdelay(S)_ returns the delay group between ports 1 and 2, and _groupdelay(S, P2, P1)_
does the same but for ports P2 and P1.

[UPDATED Nov 26 2017]
Tests:
[Consistency with theoretical results](https://gist.github.com/andresmmera/2d1eaceaa3899e15835bfcf2098a7692) 
[P1 and P2 out of range](https://gist.github.com/andresmmera/5d6322e711430f55c04768e971cfcda8) 
[Handle P1=P2 case](https://gist.github.com/andresmmera/0626a8eb6c12f74ff57dc74d542c7848) 

Examples:
[Branch line coupler](https://gist.github.com/andresmmera/fba7ab9953ffed6c03b294ef0e9eb7ca)
[Wilkinson combiner](https://gist.github.com/andresmmera/d3c6b371a87c796cd877e6e02a2b7c66) 

selection_024

### DIFF
--- a/qucs-core/src/applications.h
+++ b/qucs-core/src/applications.h
@@ -745,6 +745,13 @@ struct application_t qucs::eqn::applications[] = {
   { "smooth", TAG_VECTOR, evaluate::smooth_v_d, 2,
     { TAG_VECTOR,  TAG_DOUBLE } },
 
+  { "groupdelay", TAG_VECTOR, evaluate::groupdelay_mv, 1,
+    { TAG_MATVEC} },
+  { "groupdelay", TAG_VECTOR, evaluate::groupdelay_v, 1,
+    { TAG_VECTOR} },
+  { "groupdelay", TAG_VECTOR, evaluate::groupdelay_mv_i_i, 3,
+    { TAG_MATVEC, TAG_DOUBLE, TAG_DOUBLE } },
+
   { "rms", TAG_DOUBLE, evaluate::rms_d, 1, { TAG_DOUBLE  } },
   { "rms", TAG_DOUBLE, evaluate::rms_c, 1, { TAG_COMPLEX } },
   { "rms", TAG_DOUBLE, evaluate::rms_v, 1, { TAG_VECTOR  } },
--- a/qucs-core/src/evaluate.cpp
+++ b/qucs-core/src/evaluate.cpp
@@ -3441,6 +3441,77 @@ constant * evaluate::smooth_v_d (constant * args) {
   _RETV (smooth (*y, a));
 }
 
+// Calculates the delay group between two ports specified by the user
+constant * evaluate::groupdelay_mv_i_i (constant * args) {
+  _ARMV0 (S);
+  int index1 = D (_ARES(1))-1;//Input port
+  int index2 = D (_ARES(2))-1;//Output port
+  _DEFV ();
+  
+  strlist * deps = _ARG(0)->collectDataDependencies ();  
+  if (!deps || deps->length () != 1) {//Check there's an unique dependency
+    THROW_MATH_EXCEPTION ("Not an appropriate dependent data vector");
+    _RETC (0.0);
+  }
+  qucs::vector * freq = SOLVEE(0)->getDataVector (deps->get (0));
+
+  //Check if the length of S matvec and the freq vector are the same.
+  if (freq->getSize() != S->getSize()) {
+    THROW_MATH_EXCEPTION ("The S matrix and the frequency vector must have the same length");
+    __RETV ();
+  }
+  //Check if the input matrix is square
+  if (S->getRows() != S->getCols())
+  {
+    THROW_MATH_EXCEPTION ("The S matrix must be square");
+    __RETV ();
+  }
+
+  //Check if the port numbers are valid
+  if ((index1 < 0) || (index1 > S->getRows()-1))
+  {
+    THROW_MATH_EXCEPTION ("The 3rd argument must be a valid port number");
+    __RETV ();
+  }
+  if ((index2 < 0) || (index2 > S->getRows()-1))
+  {
+    THROW_MATH_EXCEPTION ("The 4th argument must be a valid port number");
+    __RETV ();
+  }
+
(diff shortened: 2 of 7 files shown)


in3otd:
thanks; I have a few questions:
- why force the group delay to be positive ? It can indeed be negative and even using `abs()` will not correct the negative peaks that can happen around frequencies where the magnitude has a dip
- in your  examples you compute *d(phase)/d(frequency)* but the group delay is *d(phase)/d(angular frequency)*. Maybe the 2&#960; should be included in the function.
- why exclude the P1=P2 case? One may want to know the group delay the reflection...

It would be nice to avoid having to explicitly specify the (radian) frequency in the function call, since Qucs knows the vector dependency but I think this is not passed to the equation subsystem.

andresmmera:
> why force the group delay to be positive ? It can indeed be negative and even using abs() will not correct the negative peaks that can happen around frequencies where the magnitude has a dip
in your examples you compute d(phase)/d(frequency) but the group delay is d(phase)/d(angular frequency). Maybe the 2π should be included in the function.

Ok, I agree :-)

> why exclude the P1=P2 case? One may want to know the group delay the reflection...

Honestly, I've never done such measurement, but if you think it might be useful, I'll include that case

> It would be nice to avoid having to explicitly specify the (radian) frequency in the function call, since Qucs knows the vector dependency but I think this is not passed to the equation subsystem.

Good idea, I'll investigate if that's possible.

I guess I'll amend the commit soon. 
Thanks for the quick review :+1:

andresmmera:
Hello,

I've just corrected the issues you pointed out.

On the other hand, [I figured out how to get the frequency data without passing that as an argument.](https://github.com/Qucs/qucs/pull/751/files#diff-a00a113d45f536de2592768556b7c8d9R3452)
However, this piece of code is a bit obscure (at least for me). According to the [macro definition](https://github.com/andresmmera/qucs/blob/388b961d25767fae20db27c2f10192ec6071faa9/qucs-core/src/evaluate.cpp#L140), I think SOLVEE(i) is getting the solver object from the i-th argument of _evaluate::groupdelay_mv_i_i_, right?

in3otd:
nice - I'm also not too familiar with the Equations code and didn't realize the independent variable was readily accessible. I agree with your explanation of what the `SOLVEE(i)` macro is doing, but it's the first time I look at that :grin:.

I think there is no need to look for the `frequency` independent variable, you can just take the independent variable, regardless of the name, as done [here](https://github.com/Qucs/qucs/blob/develop/qucs-core/src/evaluate.cpp#L3296) maybe checking that there is just a single dependency, as done just above in that code. This will allow to easily use the `groupdelay()` function also with AC simulations (which define `acfrequency`). I don't think there is an (immediate) need to support multiple dependencies (e.g. when doing a parameter sweep), as other functions also does not support it, IIRC.

Also adding a function to compute the group delay for a simple vector would be nice, so it will work for  generic vectors with a single dependency (e.g. on a variable `gain=S[2,1]`)

The docs are missing a minus sign in the formula.

It would be nice to add an example schematic and a test schematic to `qucs-test` (I realize that I forgot to do myself this when adding the `smooth()` function :grin:)

andresmmera:
> I think there is no need to look for the frequency independent variable, you can just take the independent variable, **regardless of the name**

Ok, but I guess that in this case a user could use the groupdelay() function along with (e.g.) the transient simulation without having errors. Consequently, the final result will be wrong.
Wouldn't it be safer to look for "frequency" or "acfrequency"? I mean, something like:

qucs::vector * freq = SOLVEE(0)->getDataVector ((char *) "frequency");
if (isempty(freq)) freq = SOLVEE(0)->getDataVector ((char *) "acfrequency")

> Also adding a function to compute the group delay for a simple vector would be nice, so it will work for generic vectors with a single dependency (e.g. on a variable gain=S[2,1])
It would be nice to add an example schematic and a test schematic to qucs-test

I fully agree :+1:

in3otd:
well, I think there is no Right choice, I tend to leave the users the freedom to make whatever they want, assuming the usual [GIGO law](https://en.wikipedia.org/wiki/Garbage_in%2C_garbage_out) is known - or should be learned the hard way :grin:.
Also the current functions seem to do only the minimal checking needed to assure no crashes ever occurs but they do not (try to) check the meaning of the data carried by the input variables. E.g., you can pass any matrix to `Rollet()` or `GaCircle()` but if that is not an S-parameters matrix the results will not make sense.
Having say that, as long as `groupdelay()` works with AC and SP simulations I'm fine, I'll leave to you to decide how much checking should be done :grin:.

andresmmera:
Ok, you convinced me  :smiley:

andresmmera:
- [x] [Get independent variable without taking care of its name ](https://github.com/Qucs/qucs/pull/751/files#diff-a00a113d45f536de2592768556b7c8d9R3456)+ [check there's an unique data dependency](https://github.com/Qucs/qucs/pull/751/files#diff-a00a113d45f536de2592768556b7c8d9R3452)
- [x] [Add function for calculating the group delay from a single vector](https://github.com/Qucs/qucs/pull/751/files#diff-a00a113d45f536de2592768556b7c8d9R3502)
- [x] [Update docs](https://github.com/Qucs/qucs/pull/751/files#diff-4cf4d72cb280742b62bed85d2b518f84R6340) and [correct the missing minus sign](https://github.com/Qucs/qucs/pull/751/files#diff-4cf4d72cb280742b62bed85d2b518f84R6378)
- [ ] Add example to qucs-test. Perphaps, the appropriate moment for that will be when this PR gets merged. I reckon that [this](https://gist.github.com/andresmmera/2d1eaceaa3899e15835bfcf2098a7692) would be a good one.

I added the function vector groupdelay(vector svec, vector freq) at vector.cpp since I think that having the following line:

> -diff(unwrap(arg(svec)),(2*pi)*(freq));

replicated in groupdelay_v(), groupdelay_mv() and groupdelay_mv_i_i() is redundant.

PR #755, opened 2017-12-01
Correct marker misbehaviour when pointing to data outside the diagram limits

andresmmera:
When moving a marker back and forth, it points to (0,0) if the data the marker is pointing at lies outside the diagram limits. See [this](https://drive.google.com/open?id=1smNdsFWsCOArMRTy7Y7Rhlb55d376ks9)

This PR solves that issue by checking if that point is above the upper limit or below the lower limit of the y-axis and setting fCX and fCY according to such limits. For this purpose, [I've set Diagram::regioncode() to be public](https://github.com/Qucs/qucs/compare/develop...andresmmera:FixMarker_Bug_when_pointing_to_hidden_data?expand=1#diff-793f9ef8aa12236983a5432bbd58b798R125). I think this shouldn't have any side-effect...
In addition to this, [I defined a private boolean variable](https://github.com/Qucs/qucs/compare/develop...andresmmera:FixMarker_Bug_when_pointing_to_hidden_data?expand=1#diff-11755930ce8a7c8de63278c3bdb3f58fR78) in the Marker class which is set to true when the graph data is outside the limits and is used [here](https://github.com/Qucs/qucs/compare/develop...andresmmera:FixMarker_Bug_when_pointing_to_hidden_data?expand=1#diff-d6888ff24dbf42ac6c04ad6e983c02bcR533) to display a big red point to emphasize that the data is missing.

This is the marker behavior using this patch https://drive.google.com/open?id=1mhPCOiz5HuMX_rbZdYvBWq9irZh4aHve

### DIFF
--- a/qucs/qucs/diagrams/curvediagram.cpp
+++ b/qucs/qucs/diagrams/curvediagram.cpp
@@ -74,10 +74,10 @@ void CurveDiagram::calcCoordinate(const double*, const double* yD, const double*
 }
 
 // --------------------------------------------------------------
-void CurveDiagram::finishMarkerCoordinates(float& fCX, float& fCY) const
+void CurveDiagram::finishMarkerCoordinates(Marker *m) const
 {
-  if(!insideDiagram(fCX, fCY)) {
-	  fCX = fCY = 0.0;
+  if(!insideDiagram(m->fCX, m->fCY)) {
+	  m->fCX = m->fCY = 0.0;
   }
 }
 
--- a/qucs/qucs/diagrams/curvediagram.cpp
+++ b/qucs/qucs/diagrams/curvediagram.cpp
@@ -76,7 +76,7 @@ void CurveDiagram::calcCoordinate(const double*, const double* yD, const double*
 // --------------------------------------------------------------
 bool CurveDiagram::clipCoordinates(float &fCX, float& fCY) const
 {
-  if(!insideDiagram(fCX, fCY)) {
+  if(insideDiagram(fCX, fCY)) {
     return false;
   }
 
--- a/qucs/qucs/diagrams/curvediagram.cpp
+++ b/qucs/qucs/diagrams/curvediagram.cpp
@@ -240,7 +240,7 @@ Frame:
 // ------------------------------------------------------------
 bool CurveDiagram::insideDiagram(float x, float y) const
 {
-  return (regionCode(x, y) == 0);
+  return (regionCode(x, y) == CS_INSIDE);
 }
 
 // ------------------------------------------------------------
--- a/qucs/qucs/diagrams/curvediagram.cpp
+++ b/qucs/qucs/diagrams/curvediagram.cpp
@@ -74,14 +74,11 @@ void CurveDiagram::calcCoordinate(const double*, const double* yD, const double*
 }
 
 // --------------------------------------------------------------
-bool CurveDiagram::clipCoordinates(float &fCX, float& fCY) const
+void CurveDiagram::finishMarkerCoordinates(float &fCX, float& fCY) const
 {
-  if(insideDiagram(fCX, fCY)) {
-    return false;
+  if(!insideDiagram(fCX, fCY)) {
+    fCX = fCY = 0.0;
   }
-
-  fCX = fCY = 0.0;
-  return true;
 }
 
 // ------------------------------------------------------------
--- a/qucs/qucs/diagrams/curvediagram.cpp
+++ b/qucs/qucs/diagrams/curvediagram.cpp
@@ -74,11 +74,14 @@ void CurveDiagram::calcCoordinate(const double*, const double* yD, const double*
 }
 
 // --------------------------------------------------------------
-void CurveDiagram::finishMarkerCoordinates(Marker *m) const
+bool CurveDiagram::clipCoordinates(float &fCX, float& fCY) const
 {
-  if(!insideDiagram(m->fCX, m->fCY)) {
-	  m->fCX = m->fCY = 0.0;
+  if(!insideDiagram(fCX, fCY)) {
+    return false;
   }
+
+  fCX = fCY = 0.0;
+  return true;
 }
 
 // ------------------------------------------------------------
--- a/qucs/qucs/diagrams/curvediagram.h
+++ b/qucs/qucs/diagrams/curvediagram.h
@@ -32,7 +32,7 @@ public:
   int  calcDiagram();
   void calcLimits();
   void calcCoordinate(const double*, const double*, const double*, float*, float*, Axis const*) const;
-  bool clipCoordinates(float &fCX, float& fCY) const;
+  void finishMarkerCoordinates(float &fCX, float& fCY) const;
   bool insideDiagram(float, float) const;
 
 protected:
--- a/qucs/qucs/diagrams/curvediagram.h
+++ b/qucs/qucs/diagrams/curvediagram.h
@@ -32,7 +32,7 @@ public:
   int  calcDiagram();
   void calcLimits();
   void calcCoordinate(const double*, const double*, const double*, float*, float*, Axis const*) const;
-  void finishMarkerCoordinates(float&, float&) const;
+  void finishMarkerCoordinates(Marker*) const;
   bool insideDiagram(float, float) const;
 
 protected:
--- a/qucs/qucs/diagrams/curvediagram.h
+++ b/qucs/qucs/diagrams/curvediagram.h
@@ -32,7 +32,7 @@ public:
   int  calcDiagram();
   void calcLimits();
   void calcCoordinate(const double*, const double*, const double*, float*, float*, Axis const*) const;
-  void finishMarkerCoordinates(Marker*) const;
+  bool clipCoordinates(float &fCX, float& fCY) const;
   bool insideDiagram(float, float) const;
 
 protected:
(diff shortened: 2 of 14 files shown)


in3otd:
yep, this was long needed :grin:

Works for the rectangular diagram, but now the Smith Chart (and probably others) always display a red point on the marker, likely because `point_outside_graph` is uninitialized.
But I'll suggest to do all this in a more general way, so the case of a marker pointing outside the visible diagram region can be easily handled also for the other diagrams (code for other diagrams can be added in future PRs).
We already have `finishMarkerCoordinates()` which handles the case of "marker pointing outside", only it does not do it properly. Since is a virtual function every diagram can decide how to handle this case, so no need for things like `if (diag()->Name=="Rect")`, we can just continue to call `diag()->finishMarkerCoordinates()` but passing the whole Marker `diag()->finishMarkerCoordinates(this)` (ok, maybe this function needs a different name now).
Now every diagram can check if the marker is outside, for their own definition of "outside", set the Marker `point_outside_graph` and decide what to do with the Marker coordinates.

BTW, I'm starting to think we should maybe have some derived Marker classes for the various diagrams, I do not like much all the stuff like `if (diag()->Name !=`*something*`)` in `marker.cpp` but this is of course outside the scope of this PR.

felix-salfelder:
> `if(diag()->Name=="Rect")`

do *NOT* do that. this kind of code is the root of most evil in qucs.

(sorry for repeating myself)

andresmmera:
@in3otd 

> Works for the rectangular diagram, but now the Smith Chart (and probably others) always display a red point on the marker, likely because point_outside_graph is uninitialized.

panic

@felix-salfelder It smells like we have something to do with this:

https://github.com/Qucs/qucs/blob/develop/qucs/qucs/diagrams/marker.cpp#L66
https://github.com/Qucs/qucs/blob/develop/qucs/qucs/diagrams/marker.cpp#L137
https://github.com/Qucs/qucs/blob/develop/qucs/qucs/diagrams/marker.cpp#L137
https://github.com/Qucs/qucs/blob/develop/qucs/qucs/diagrams/marker.cpp#L160
https://github.com/Qucs/qucs/blob/develop/qucs/qucs/diagrams/marker.cpp#L357
etc.

> But I'll suggest to do all this in a more general way, so the case of a marker pointing outside the visible diagram region can be easily handled also for the other diagrams (code for other diagrams can be added in future PRs).

@in3otd and @felix-salfelder please, excuse my ignorance, but is there some work ongoing to restructure that code? According to your comments, a more extensive work than this PR is needed.

andresmmera:
I'm going to close this (at least for the moment)

in3otd:
@andresmmera , no need to close the PR :grin: , its purpose was good but the Diagram and Marker code is a bit messy already and it's not easy to modify it "cleanly" (for some definition of "clean"). IMHO, just expanding `finishMarkerCoordinates()` is clean enough here.

There was/is some work from @felix-salfelder ongoing on the markers since quite a long time, honestly I've to admit I'm not too familiar with that.

felix-salfelder:
we need (something like) #370 asap. the main point is representation of plottable data ("graphs") and how markers move on top of that. my last rebase got stuck (did it break something?), there are conflicts with more hacks that have been added in the meantime.

(`if(object->name()=="whatnot")` and the like...)

felix-salfelder:
> if(diag()->Name=="Phasor") m = phasormk(pz,px,nnn);

whatever "Phasor" is. this code must go to diagram_phasor.cpp. same as with the meters: if nobody does it, i do hope we will just get rid of it.

in3otd:
@andresmmera I've modified your code to show where the changes should IMHO be.
At first, I've added a revert commit to start again from develop, of course this and the one before can simply be removed if we decide to merge this one day.

The code that check if a marker points outside the diagram is now in `Diagram::finishMarkerCoordinates()`; this has the advantage that the right function is automatically called for every diagram. For this to work I had to make public a couple of thing in the `Marker()` class - maybe I could just have made the `Diagram()` class or only the method above a `friend`.
I've also added the code for the Smith chart diagram, actually for all the diagrams which do not override `Diagram::finishMarkerCoordinates()`; note that there is some "asymmetry" in the code, in that `Diagram()` defines some methods that are meant to be used with round diagrams (like `insideDiagram()` and `finishMarkerCoordinates()`) so the rectangular diagrams need to override them (see e.g. `Waveac()` or `RectDiagram()`) in the same way. IMHO it would be cleaner to have two classes `RoundDiagram()` and `RectangularDiagram()` and have the actual diagrams derive from these, but that will be for the future.

I did only some minimal checking as, as said, this was mainly to show the code I had in mind.

felix-salfelder:
i would like to advise against making internals public even more. we need the opposite.

what is `outside_graph` supposed to mean? (can you please remove this?)

in3otd:
yes, I understand, but I saw no other way to set the coordinates of a `Marker` from a `Diagram` (ok, besides making `Diagram::finishMarkerCoordinates()` a friend of `Marker`). Could have added a setter and getter but for simple types I'm not sure it's worth it.
`outside_graph` should actually have been named `outside_diagram_boundaries`, is to tell the marker "please draw yourself with a different color". As above, I didn't see any practical alternative solution but I'm open to suggestions :grin:.

felix-salfelder:
> `outside_diagram_boundaries`

turn it into a function. it could become a virtual member of a marker base class in the future (i'm not saying that i see the need for that yet).

i've seen some "region code" stuff. cleaning that up might help. what is the common denominator of region codes and `outside_diagram_boundaries`?

> friend

i don't think that is a good idea. selectively exposing internals is no better than exposing internals. friends will break when introducing inheritance.

qucs/qucs/diagrams/curvediagram.cpp
```diff
@@ -74,11 +74,14 @@ void CurveDiagram::calcCoordinate(const double*, const double* yD, const double*
 }
 
 // --------------------------------------------------------------
-void CurveDiagram::finishMarkerCoordinates(float& fCX, float& fCY) const
+bool CurveDiagram::clipCoordinates(float &fCX, float& fCY) const
 {
   if(!insideDiagram(fCX, fCY)) {
```
in3otd: oh, #!@&!, the `!` should have been removed here

in3otd:
I've rearranged the code so that nothing is moved to `public`, seems to work still.

> what is the common denominator of region codes and outside_diagram_boundaries?

the *region codes* are the *outcodes* for the Cohen-Sutherland algorithm, Wikipedia told me
```
             left central right
        top  1001   1000   1010
    central  0001   0000   0010
     bottom  0101   0100   0110
```
and a marker is then outside the diagram if its position region code is not 0 .

felix-salfelder:
something like

    enum cs_REGION {
      CS_TOP_LEFT=9,
      CS_TOP_CENTRAL=8,
      [..]
    };

(in a sensible namespace or class) might make things more readable...

andresmmera:
I couldn't do an detailed check, but I saw [this](https://drive.google.com/open?id=1f7zf90cHGBACqqzBvI0yuhRLklw8xJvM). The marker is pointing outside the diagram, but this behavior seems to depend on the graph. Notice that the marker on the green trace is working as duly.

The schematic is used for the test is [this](https://gist.github.com/andresmmera/7a446e6db30ca14b8bc05ba1a1909d56)

in3otd:
thanks, I had a feeling that the code [here](https://github.com/Qucs/qucs/pull/755/files#diff-1fd4a37e1a744ba635da060f91316309) was not quite right, now I understand why :grin:.
Likely I should check for all the regions, not only top/bottom/left/right or even better, check every top/bottom/left/right flag and set the marker (clipped) coordinates accordingly.

andresmmera:
Thanks! Everything works as expected.

felix-salfelder:
(why) do we need to change the name of the function in the diagram interface?

`finishMarkerCoordinates` sounds about perfectly right to me (for that purpose).

in3otd:
I changed that mainly because now the function returns a `bool` to tell whether the coordinates were modified (clipped) or not, so a `true` or `false` from a function named `clipCoordinates()` should help to remember/understand that when reading the code. A `true` or `false` from a function named `finishMarkerCoordinates()` sounds less clear to me.
But I can change that back if "needed".

felix-salfelder:
the problem is, that, onceever some other coordinate finish is required, one will have to modify the interface again. this contradicts the idea of the interface, which is supposed to allow just what is needed to implement the functionality -- which is *any* kind of coordinate finish, and which should not go into the clip function (maybe just because it does not clip).

i didn't notice the return type. perhaps it's much better to have
`virtual void isVisible() const;` or `withinBoundaries()` instead.

and.. `assert(diag());` is gone. that might turn out useful for catching segfaults.

thanks

in3otd:
>   one will have to modify the interface again.

yep, when things evolve the interface may need to change, I do not see a problem with this.

Anyway, I've modified the code according to what I understood of your suggestions; to me it seems uglier now, since now `insideDiagram()` has to be called twice, once to check if the marker is inside the diagram and another by `finishMarkerCoordinates()`. And now one has to always call the two functions "together", like
```
 outside_diagram_boundaries = !diag()->insideDiagram(fCX, fCY);
 diag()->finishMarkerCoordinates(fCX, fCY);
```

> and.. assert(diag()); is gone.

yep, should have been moved instead - I removed since I thought that `paint()` will never be called/reached if `diag()` is null but, well, an `assert()` helps checking if the reasoning is right :grin:

PR #757, opened 2017-12-04
fix visibility of 3 component attributes.

felix-salfelder:
handle legacy abuse cases, not changing behaviour


felix-salfelder:
all but cmake is working now.

guitorri:
Sorry I messed up. It should be only one commit to fix the CMake build.

guitorri:
Rebased your branch to clean up my mess. My apologies.

in3otd:
after 4b39404098369f41bd14e952d764aa22920ac1e9 the way Grounds symbols are shown on the schematic changed slightly:
image
there is now a sort of reference printed near the symbol; I tried to figure out from where this comes from but was not really able to follow the code

felix-salfelder:
could have to do with component.cpp+802 ff.

will have another look
thanks

felix-salfelder:
yes. 4b39 did something "wrong" on GND. i have untangled this a bit more.

PR #758, opened 2017-12-09
Fix qucsconv matlab file generation

in3otd:
Fixes a bug introduced [here](https://github.com/Qucs/qucs/commit/85e658588bed1ff88fda328d4cb07cd1bb40b57f#diff-10505be9f9b84a5b48cae5a915a76b88) in the variable name sanitization which caused `qucsconv` to crash when writing a matlab file.

Found while testing #693 ; in practice I think `qucsconv` currently segfaults on any conversion to a matlab file.


andresmmera:
I can reproduce the bug on my side and, yes, it crashes for every qucs data file I tried to convert to matlab file format.

Besides this, I tested this PR locally and confirm it solves the problem. :+1:

PR #762, opened 2017-12-28
fix "make subdirs" call cwd dependency

felix-salfelder:


in3otd:
(for context, see latest comments in #660)

PR #763, opened 2017-12-31
Copy examples to home dir

andresmmera:
Hello,

In this PR it was added a new option ("Copy examples in the user workspace") under the file menu. The reason for this new feature is that the examples are located at /usr/share, which is a read-only directory (so the user cannot run simulations nor save changes).

In addition to this, it was added an extra field to the schematic file format in order to provide a short description while hovering the mouser pointer over the file name in the project tree
screenshot from 2017-12-31 08-12-40

Finally, it was added another file menu option (right-click on the file name) so as to let the user to edit such description
screenshot from 2017-12-31 08-46-30

### DIFF
--- a/qucs/qucs/qucs.cpp
+++ b/qucs/qucs/qucs.cpp
@@ -254,7 +254,7 @@ void QucsApp::initView()
   TabView->addTab(Content, tr("Content"));
   TabView->setTabToolTip(TabView->indexOf(Content), tr("content of current project"));
 
-  connect(Content, SIGNAL(clicked(const QModelIndex &)), 
+  connect(Content, SIGNAL(clicked(const QModelIndex &)),
           SLOT(slotSelectSubcircuit(const QModelIndex &)));
 
   connect(Content, SIGNAL(doubleClicked(const QModelIndex &)),
@@ -887,7 +887,7 @@ void QucsApp::initCursorMenu()
   connect(action, SIGNAL(triggered()), SLOT(slot())); \
   ContentMenu->addAction(action); \
   }
-  
+
   APPEND_MENU(ActionCMenuOpen, slotCMenuOpen, "Open")
   APPEND_MENU(ActionCMenuCopy, slotCMenuCopy, "Copy file")
   APPEND_MENU(ActionCMenuRename, slotCMenuRename, "Rename")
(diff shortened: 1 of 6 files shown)


felix-salfelder:
using qucs to copy files does not seem to be the right approach to me.

files in /usr/share being readonly seems quite common. i would suggest to just open one of them, then 'save as' in a writeable location, or just copy before open.

maybe it would be useful if qucs knew that the currently open file is readonly, for better guidance.

(how do other qt applications handle this?)

in3otd:
Could the schematic description be a field in the schematic Frame (maybe with optional visibility)? The Frame info and the description have a similar purpose.

When the destination directory already exists, you inform the user that its content will be *overwritten* but you first remove completely the old content, so I'll say the the content will be *replaced* instead, *Overwriting* should leave any file which is not present in the source directory untouched, IMHO.

For the `QMessageBox` you should specify the parent (`this`), because it's the Right Thing :grin: and because doing so we will get the right Qucs icon on the message window (for some OS/Window manager)

> readonly

yep, this will make a lot of sense in general - we should tell the user that the file/project is read-only and disallow any modification. It would be useful also when casually browsing thru projects/schematics, to avoid modifying things (been there, done that...). Not sure how easy it will be to implement a "view-only" schematic in Qucs.

> using qucs to copy files does not seem to be the right approach to me.

yes and no - surely we can copy files manually but having Qucs doing that makes it straightforward, the user does not need to know where the examples are located and where they need to be copied.
I think it will also be useful e.g. if one day we implement lock files for the Projects (I think we briefly discussed that in the past), so that Qucs will know not to copy the lock.

IMHO the examples should be reorganized to have one Project for every example, so you could also have several Schematics and Data Display for the same example. For this to be useful we should be able to browse thru directories with the Qucs Projects Browser so all the examples Projects could go in a separate subdirectory of the "Qucs Home".

andresmmera:
Well, it seems that the examples stuff deserves a PR of its own. So I think it'd be a good idea if I put the schematic description stuff in a separate PR and remove that commits from this branch.

So, if I understood correctly, we should take the following steps:
- [] When the user opens the schematic, detect if it is read-only and prevent modification
- [] Make the project viewer capable of navigate through directories
- [] Put examples in separate project dirs

Right?

@in3otd meanwhile I addressed the stuff you mentioned concerning the QMessageBox objects.
@in3otd @felix-salfelder you guys decide if copying files using Qucs could be a problem or not. :-)

andresmmera:
I ignore why it shows 13 commits here... Maybe I did something wrong during the rebase...
Does anyone know how to solve that?

in3otd:
regarding the *Make the project viewer capable of navigate through directories*, I have some working code already, as mentioned on IRC, just need to polish it a bit

andresmmera:
Cool. Then, I'll start trying to implement the lock feature (in the least intrusive way possible)

PR #764, opened 2018-01-01
Show schematic description while hovering the mouser pointer over the file name

andresmmera:
The aim of this PR is to show a schematic description when the user hovers the mouse pointer over the project tree. In this sense, it was written a function for reading the schematic title from its frame.

Please refer to #763 for context
screenshot from 2018-01-01 09-41-57

### DIFF
--- a/qucs/qucs/projectView.cpp
+++ b/qucs/qucs/projectView.cpp
@@ -184,5 +184,6 @@ QString ProjectView::ReadDescription(QString file)
     } while (!line.isNull());
 
     QucsDocument.close();
+    if (description == "Title") description = "";//Prevent schematics without description from showing "Title" in the tooltip
     return description;
 }
--- a/qucs/qucs/projectView.cpp
+++ b/qucs/qucs/projectView.cpp
@@ -107,7 +107,10 @@ ProjectView::refresh()
     extName = QFileInfo(workPath.filePath(fileName)).suffix();
 
     columnData.clear();
-    columnData.append(new QStandardItem(fileName));
+    QStandardItem * d = new QStandardItem(fileName);
+    QString description = ReadDescription(workPath.absoluteFilePath(fileName));
+    d->setToolTip(description);
+    columnData.append(d);
 
     if(extName == "dat") {
       APPEND_CHILD(0, columnData);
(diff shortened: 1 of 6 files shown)


guitorri:
Looks good to me. But IMHO there should be an option on the settings to enable/disable the tooltip.

andresmmera:
Ok, I've put that option under the "Application Settings" menu since I thought it was the most appropriate place. Additionally, I left the checkbox unchecked by default

showdescriptionmenu

in3otd:
why there is a [commit](https://github.com/Qucs/qucs/pull/764/commits/3277589d8986874ec44bf9fca11b8f41e581e50a) for the lock function here? Also, it seems that a schematic is locked by default (?).

If a multiline description is used the tooltip show all the text on a single line, with `\n` separators; maybe this [depends](http://doc.qt.io/archives/qt-4.8/qtooltip.html#details) on the text format used.

The `QucsSettings.ShowDescriptionProjectTree` setting value is not saved on exit.

andresmmera:
> why there is a commit for the lock function here?

That shouldn't be there. I've just removed that from the commit history

> The QucsSettings.ShowDescriptionProjectTree setting value is not saved on exit.

Fixed here https://github.com/Qucs/qucs/pull/764/commits/5e7f9601f803be956cbd6129493cf9e9a8439f92

In addition to this, I've also added a [commit](https://github.com/Qucs/qucs/pull/764/commits/ccb1ef2a1024c90d91e806f0d276236993fee12b) to prevent those files where the description field is empty from showing the default text ("Title") in the tooltip

> If a multiline description is used the tooltip show all the text on a single line,

I've been doing some tests here... If I replace \n by the html breakline tag, < br > it works as duly

screenshot from 2018-04-07 10-09-14

It may sound stupid, but I'm having problems to replace '\n' by "< br > at ProjectView::ReadDescription(QString). I think the proper way to do that should be:

`description.replace(QRegExp("[\\n]"),"<` br >"); [1]
or even `description.replace('\n',"<` br >");

but it doesn't work... Any guess?

P.S.: I'm adding a blank space after and before br because otherwhise the github wiki markdown replaces that by a breakline...

[1] http://www.qtcentre.org/threads/22060-removing-carriage-return-from-QString

qucs/qucs/projectView.cpp
```diff
@@ -157,3 +163,28 @@ ProjectView::exportSchematic()
...
+            index2 = line.indexOf(">", index);
+            index+=11;//Skip FrameText0=
+            description = line.mid(index, index2-index);
+            break;
+        }
+    } while (!line.isNull());
+
+    description.replace("\\n", "<br>");
+    QucsDocument.close();
+    if (description == "Title") description = "";//Prevent schematics without description from showing "Title" in the tooltip
```
in3otd: the description string is internationalized: in [schematic.cpp](https://github.com/Qucs/qucs/blob/develop/qucs/qucs/schematic.cpp#L98) the default value of `Frame_Text0` is `tr("Title")` so it would have been better to compare to `tr("Title")`.
Even better will be to show the description only if `showFrame` is set; the rationale is that the description is shown in the schematic and so has probably set to something sensible only if that is set..
This will also handle correctly the case when you open a schematic created with another Qucs language setting. E.g. when Qucs is set to use Italian the `Frame_Text0` is correctly set to *Titolo* but if you now open this schematic from a Qucs with language setting set to English it will check if the description is set to *Title* so the comparison will fail and it will show the *Titolo* string as description in the tooltip.

andresmmera: I agree. 
Since this PR is already merged, I've opened a new one (#868) to fix this.

PR #767, opened 2018-01-07
Allow navigating directories in the Projects tab

in3otd:
As mentioned in #763, I did some small change to allow navigating thru directories in the Projects tab.

image

Code seems to be working but is not yet finished:
- currently the QucsHome setting is changed to the current working directory, it will be better to keep it fixed (sort of "Qucs start-up home") and use another variable to keep the current working directory
- in the Projects Tab maybe the normal directories and the projects should be grouped together and not simply mixed in alphabetical order
- etc.

### DIFF
--- a/qucs/qucs/dialogs/packagedialog.cpp
+++ b/qucs/qucs/dialogs/packagedialog.cpp
@@ -99,8 +99,8 @@ PackageDialog::PackageDialog(QWidget *parent_, bool create_)
     connect(ButtCancel, SIGNAL(clicked()), SLOT(reject()));
 
     // ...........................................................
-    // insert all projects
-    QStringList PrDirs = QucsSettings.QucsHomeDir.entryList(QStringList("*"), QDir::Dirs, QDir::Name);
+    // insert all projects in the current directory
+    QStringList PrDirs = QucsSettings.projsDir.entryList(QStringList("*"), QDir::Dirs, QDir::Name);
     QStringList::iterator it;
     for(it = PrDirs.begin(); it != PrDirs.end(); it++)
        if((*it).right(4) == "_prj"){   // project directories end with "_prj"
@@ -292,7 +292,7 @@ void PackageDialog::slotCreate()
     if(p->isChecked()) {
       s = p->text() + "_prj";
       Stream << Q_UINT32(CODE_DIR) << s.toLatin1();
-      s = QucsSettings.QucsHomeDir.absolutePath() + QDir::separator() + s;
+      s = QucsSettings.projsDir.absolutePath() + QDir::separator() + s;
       if(insertDirectory(s, Stream) < 0) {
         PkgFile.close();
         PkgFile.remove();
@@ -358,7 +358,7 @@ void PackageDialog::extractPackage()
   }
   QDataStream Stream(&PkgFile);
 
-  QDir currDir = QucsSettings.QucsHomeDir;
+  QDir currDir = QucsSettings.projsDir;
   QString Version;
   VersionTriplet PackageVersion;
   Q_UINT16 Checksum;
--- a/qucs/qucs/qucs.cpp
+++ b/qucs/qucs/qucs.cpp
@@ -353,6 +353,19 @@ void QucsApp::initView()
   m_homeDirModel->setFilter(QDir::NoDot | QDir::AllDirs);
 
   // ............................................
+  QString path = QucsSettings.QucsHomeDir.absolutePath();
+  QDir ProjDir(path);
+  // initial projects directory is the Qucs home directory
+  QucsSettings.projsDir = path;
+
+  // create home dir if not exist
+  if(!ProjDir.exists()) {
+    if(!ProjDir.mkdir(path)) {
+      QMessageBox::warning(this, tr("Warning"),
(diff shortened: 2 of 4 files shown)


in3otd:
updated to keep the QucsHome setting, used as "Qucs start-up home", and using a separate variable to track the current project directory. Also in the Projects Tab the normal directories and the project directories are grouped together for easier navigation.
Seem to work ok here, please check :grin:

andresmmera:
In Windows, the Qucs icons are not shown in the project list.

projnav

As suggestion for improvement, I think it would be nice if the project navigator could go a folder level up by pressing the backspace key instead of clicking ".." with the mouse.

in3otd:
> In Windows, the Qucs icons are not shown in the project list.

Thanks, I tested the code on Linux only. But now I have downloaded the Windows artifact produced by AppVeyor for this PR and using that I do see the icons. Did you build the executable yourself or use the AppVeyor zip package?

> explore the folder content and look for Qucs schematics instead of inspecting the folder name

not sure I understood what advantage this will bring, I mean, it will not change the number of folders shown on the Project navigator :grin:. I keep my `_prj` directories in different subdirectories, roughly ordered by subject, for an easier navigation.

oh, by the way, thanks to your remark I noticed that I just check for `prj` instead of `_prj`, I need to fix this.

andresmmera:
Yes, I used the AppVeyor 64 bits artifact to do the test... Windows 10
here, not sure if it matters. I'll try to repeat the test in other
machine...

Regarding the _prj stuff, I think Qucs should determine that a folder is a
project or not because of its content, not because of the _prj suffix. I
know it won't reduce the number of folders, but I don't see why all my Qucs
project folders should end with _prj. In practice, I order my Qucs
schematics in folders, not always using the _prj ending. Consequently,
these folders are not interpreted as Qucs projects and are not included in
the project list. As said, this is just a comment, maybe irrelevant.

andresmmera:
More on the _prj topic... Probably, everybody arrange general projects (not necessarily Qucs project) using folders, but these folders may contain a bunch of random documents: datasheets, reports, s2p measurements and sometimes Qucs schematics. My point is that it would be nice if Qucs could interpret such kind of folders as Qucs projects and filter out the non-Qucs files.

in3otd:
I've also used the AppVeyor 64 bits artifact on Windows 10, strange it behaved differently. Just to be sure, did you run the `qucs.exe` found in `qucs-win64/bin` and not the one in `qucs-win64/c/qucs-win64/bin` ? This latter should not be there (this is corrected in #801) and does actually not even start properly here.

I got your point on the `_prj` extension - not sure I like that every directory can be a project but this is likely jut a personal preference. I prefer that Qucs is able to launch the right application for every file in the project directory(#750). Also we would need to scan the directories content to determine if it's a project directory or not, not sure if it will cause some noticeable delay when browsing directories.

andresmmera:
> I've also used the AppVeyor 64 bits artifact on Windows 10, strange it behaved differently. Just to be sure, did you run the qucs.exe found in qucs-win64/bin and not the one in qucs-win64/c/qucs-win64/bin ? This latter should not be there (this is corrected in #801) and does actually not even start properly here.

I'm sure I run that from qucs-win64/bin, but I cannot repeat the test right now. However, I downloaded again the artifact to my laptop (Ubuntu+Wine) and the icons are missing here too:

selection_046

> I got your point on the _prj extension - not sure I like that every directory can be a project but this is likely jut a personal preference. I prefer that Qucs is able to launch the right application for every file in the project directory(#750). Also we would need to scan the directories content to determine if it's a project directory or not, not sure if it will cause some noticeable delay when browsing directories.

Yes, this is a personal preference and I agree this feature could come with some delay...
Forget it :-)

in3otd:
oh, well, I just tried the windows 32-bit executable on Wine and the Qucs icons are missing :confused:

Googling around it seems that Qt issues with `svg` files in Windows are not uncommon, so I've pushed a commit to use a `png` image instead, let's see if that works fine on Windows.

in3otd:
yep, works also in Wine here now :grin:
I'll squash the latest commits.

andresmmera:
As expected, it works on Windows 10. I didn't see any other issue.

in3otd:
some mess going on here: the Travis CI was green for this PR but [failed](https://travis-ci.org/Qucs/qucs) for the merge and the [build badge](https://github.com/Qucs/qucs#readme) on the Qucs GitHub README page is green anyway ??

andresmmera:
`6079672 ` builds fine locally. I don't understand why the merge failed.

in3otd:
yep, same here, taking a closer look...

in3otd:
correction, I forgot to pull to update `develop` - now it fails also locally.

Something went bad when merging this PR, as both this and #770 touched `slotUpdateTreeview()`; this PR should probably have been rebased before merging.

I'll push clean up later: besides not compiling, the current code also contains unneeded calls to `slotUpdateTreeview()`

in3otd:
> the build badge on the Qucs GitHub README page is green anyway ??

I realized that this is because the badges are about the `master` branch only. Either we need to have a separate `README.md` for every branch with the appropriate badges or we put badges for both `master` and `develop` in there.

PR #769, opened 2018-01-13
Read-only feature

andresmmera:
See #763 for more context...

This is a new feature for locking schematics. [Screencast here](https://drive.google.com/open?id=1Lf0BKoLL91MLX2KDdQE2KAxb1aqazTlq)

In the qucsdoc class, it was set two boolean flags: readonly and locked.
The first indicates if the file has read-only permissions whereas the latter is a parameter set by the user from the Qucs GUI. In both cases, the user input should be disabled and I think that the proper way to do that is by locking the viewport() and blocking the editDelete signal.

The lock feature only supports schematic files. The reason for that is that we need to modify (somehow) the document to make the lock property permanent. In this sense, I added a <locked> tag in the <Properties> field schematic file format, but this is **only** included when the schematic is locked. Since the most common case is to work with unlocked schematics, the <locked> tag is not added for them in favor of (some) backwards compatibility.
Obviously, we cannot proceed this way with other type of documents...

selection_035
(I posted a screenshot of my text editor since github seems to have trouble with the tag format...)

Finally, a new button was added to the panel so as to let the user to lock/unlock the schematic.

Hope this helps

### DIFF
--- a/qucs/qucs/qucs.cpp
+++ b/qucs/qucs/qucs.cpp
@@ -1436,7 +1436,7 @@ bool QucsApp::saveFile(QucsDoc *Doc)
     return saveAs();
 
   int Result = Doc->save();
-  if(Result < 0)  return false;
+  if(Result < 0)  return saveAs();
 
   updatePortNumber(Doc, Result);
   slotUpdateTreeview();
@@ -3016,8 +3016,14 @@ Schematic *ContextMenuTabWidget::createEmptySchematic(const QString &name)
 {
   // create a schematic
   QFileInfo Info(name);
+  QString filename = Info.fileName();
   Schematic *d = new Schematic(App, name);
-  int i = addTab(d, QPixmap(":/bitmaps/empty.xpm"), name.isEmpty() ? QObject::tr("untitled") : Info.fileName());
+  if (d->isreadonly())
+   {// If the file is not writable, then add [READ-ONLY] to the tab title.
+    // The document can be edited but not saved.
+       filename += " [READ-ONLY]";
+   }
+  int i = addTab(d, QPixmap(":/bitmaps/empty.xpm"), name.isEmpty() ? QObject::tr("untitled") : filename);
   setCurrentIndex(i);
   return d;
 }
@@ -3026,8 +3032,14 @@ TextDoc *ContextMenuTabWidget::createEmptyTextDoc(const QString &name)
 {
   // create a text document
   QFileInfo Info(name);
+  QString filename = Info.fileName();
   TextDoc *d = new TextDoc(App, name);
-  int i = addTab(d, QPixmap(":/bitmaps/empty.xpm"), name.isEmpty() ? QObject::tr("untitled") : Info.fileName());
+  if (d->isreadonly())
+   {// If the file is not writable, then add [READ-ONLY] to the tab title.
+    // The document can be edited but not saved.
+       filename += " [READ-ONLY]";
+   }
+  int i = addTab(d, QPixmap(":/bitmaps/empty.xpm"), name.isEmpty() ? QObject::tr("untitled") : filename);
   setCurrentIndex(i);
   return d;
 }
--- a/qucs/qucs/qucs.cpp
+++ b/qucs/qucs/qucs.cpp
@@ -1810,8 +1810,10 @@ void QucsApp::slotChangeView(QWidget *w)
   }
   // for schematic documents
   else {
+    assert(w);
     Schematic *d = (Schematic*)w;
     Doc = (QucsDoc*)d;
+    assert(Doc);
     magAll->setDisabled(false);
     // already in schematic?
     if(cursorLeft->isEnabled()) {
@@ -1910,7 +1912,9 @@ void QucsApp::updatePortNumber(QucsDoc *currDoc, int No)
     if(isTextDocument (w))  continue;
 
     // start from the last to omit re-appended components
+    assert(w);
     Schematic *Doc = (Schematic*)w;
+    assert(Doc);
     for(Component *pc=Doc->Components->last(); pc!=0; ) {
       if(pc->Model == Model) {
         File = pc->Props.getFirst()->Value;
@@ -2005,7 +2009,11 @@ void QucsApp::slotIntoHierarchy()
 {
   slotHideEdit(); // disable text edit of component property
 
-  Schematic *Doc = (Schematic*)DocumentTab->currentWidget();
+  assert(DocumentTab);
+  QWidget *w = DocumentTab->currentWidget();
+  Schematic *Doc = dynamic_cast<Schematic*>(w);
+  assert(Doc);
+
   Component *pc = Doc->searchSelSubcircuit();
   if(pc == 0) { return; }
 
@@ -2070,7 +2078,9 @@ void QucsApp::slotSimulate()
   QucsDoc *Doc;
   QWidget *w = DocumentTab->currentWidget();
   if(isTextDocument (w)) {
+    assert(w);
     Doc = (QucsDoc*)((TextDoc*)w);
+    assert(Doc);
     if(Doc->SimTime.isEmpty() && ((TextDoc*)Doc)->simulation) {
       DigiSettingsDialog *d = new DigiSettingsDialog((TextDoc*)Doc);
       if(d->exec() == QDialog::Rejected)
@@ -2409,7 +2419,10 @@ void QucsApp::slotSelectSubcircuit(const QModelIndex &idx)
 void QucsApp::slotSelectLibComponent(QTreeWidgetItem *item)
 {
     // get the current document
-    Schematic *Doc = (Schematic*)DocumentTab->currentWidget();
+    assert(DocumentTab);
+    QWidget *w = DocumentTab->currentWidget();
+    Schematic *Doc = dynamic_cast<Schematic*>(w);
+    assert(Doc);
 
     // if the current document is a schematic activate the paste
     if(!isTextDocument(Doc))
@@ -2541,6 +2554,7 @@ bool QucsApp::isTextDocument(QWidget *w) {
 // symbol.
 void QucsApp::slotSymbolEdit()
 {
+  assert(DocumentTab);
   QWidget *w = DocumentTab->currentWidget();
 
   // in a text document (e.g. VHDL)
@@ -2560,7 +2574,8 @@ void QucsApp::slotSymbolEdit()
     slotChangePage(TDoc->DocName,TDoc->DataDisplay);
 
     // set 'DataDisplay' document of symbol file to original text file
-    Schematic *SDoc = (Schematic*)DocumentTab->currentWidget();
+    Schematic *SDoc = dynamic_cast<Schematic*>(w);
+    assert(SDoc);
     SDoc->DataDisplay = Info.fileName();
 
     // change into symbol mode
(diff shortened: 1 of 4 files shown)


felix-salfelder:
thanks, this makes sense.

    if (isTextDocument(DocumentTab->currentWidget()))

i suggest to do an `assert(pointer);` before every dereference operation. at least in new code.

    Schematic * d = (Schematic*)DocumentTab->currentWidget();

not good. do

     assert(DocumentTab);
     widget* w = DocumentTab->currentWidget();
     Schematic * d = dynamic_cast<Schematic*>(w);
     assert(d);

instead. (always like this, at least in new code).

andresmmera:
@felix-salfelder thanks.
Once there, I've also corrected existing code concerning the Schematic dereferences.

in3otd:
I assume this is about *When the user opens the schematic, detect if it is read-only and disable changes* (see [here](https://github.com/Qucs/qucs/pull/763#issuecomment-354645727)).
I'm not sure that we need to save the `locked` state in the document itself, I think it should be something that just exists for the current Qucs instance, something similar to the *Open read only* feature found in many editors. If a document is not writeable it will be opened in read-only mode by default.
(OTOH, one day we may implement a file/project locking using [lock files](https://en.wikipedia.org/wiki/File_locking#Lock_files) where we can put additional info about which Qucs process is using the file/project, which user since when, etc. but that would be another PR)

Besides, the schematics can still be modified when
- doing *Select All* and *Cut* or *Paste*
- using *Undo*/*Redo*

and probably also in other ways

BTW, being able to disable schematic modifications may be helpful for #471 also.

Having the *Lock* command under the *Insert* menu seems a bit strange - one locks or allows/disallows modifications of a Document/File so it should better go under the* File* menu. As said above, we should likely have an *Open read only* and a *Toggle read only* there.

andresmmera:
> I assume this is about When the user opens the schematic, detect if it is read-only and disable changes (see here).

Yes, that's right. This PR detects if a file is writable or not at the time the user opens it and appends the "[READ-ONLY]" string to the file name https://github.com/Qucs/qucs/pull/769/files#diff-07a46d9c7b31a19eccbb820ca0cd3aa8R1387

The following screenshot shows the result of opening an example file located at /usr/local/share/qucs/examples/ where I have permission 644.
lockfeature1

In addition to this, I thought it would be convenient to have a feature to lock schematics. The idea is to somehow _protect_ a schematic against accidental modifications (I reckon that Microsoft Word does something similar). 

> I'm not sure that we need to save the locked state in the document itself, I think it should be something that just exists for the current Qucs instance

Well, I don't think so. Suppose that you want to share/distribute certain schematic file, but you want it to be protected against accidental edition. I guess there's no way to do that unless you add some extra information in the header (e.g. a locked field).

> Besides, the schematics can still be modified when

>     doing Select All and Cut or Paste
>     using Undo/Redo

> and probably also in other ways

> BTW, being able to disable schematic modifications may be helpful for #471 also.

> Having the Lock command under the Insert menu seems a bit strange - one locks or allows/disallows  modifications of a Document/File so it should better go under the* File* menu. As said above, we should likely have an Open read only and a Toggle read only there.

I fully agree. I ignore why I put the lock command under the _Insert_ menu... I would swore it was once under the _File_ menu :grin: 

**Question**: Should I put the lock stuff in a separate branch and leave the read-only stuff here?

felix-salfelder:
On Wed, May 02, 2018 at 10:25:38AM -0700, Andrés Martínez-Mera wrote:
> Well, I don't think so. Suppose that you want to share/distribute
> certain schematic file, but you want it to be protected against
> accidental edition. I guess there's no way to do that unless you add
> some extra information in the header (e.g. a locked field).

what happens if the file is simply read-only? wouldn't that protect from
modifications?

i think this protection should kick in, once you try to save. attemps to
edit should only raise a warning "you are modifying a read-only file".

> **Question**: Should I put the lock stuff in a separate branch and leave the read-only stuff here?

i think read-only (for a file) and lock (for an object) should use the
same code.

one question i have is how and when readonly-flags should be propagated
and queried up and down the hierarchy... (and this should be only
implemented once).

in3otd:
So a Document can be non-modifiable because of the file permissions or because we would like to protect it from accidental editing.

In summary the possibilities should be something like
- a normal (writeable) schematic file
  - can be open read/write (the default *Open* action)
    - once loaded one can toggle the read-only status but this will only disallow editing in the current Qucs session and not have any effect on the actual file permissions
  - can be open read-only, via a *Open read only* menu)
    - once loaded one can toggle the read-only status to allow editing
  - under Document Properties there will be a *Open read-only* (the "lock" feature) field that can be checked/unchecked
  - if the Document has the *Open read-only* property set it will be open as read-only by default
- a read-only (non writeable) schematic file
  - will be open as read-only
    - read-only status can still be changed in Qucs to allow editing
  - can not be saved with the same name
  - can be saved as a non-read only file with a different name

If you'd like to implement both things now you can put everything in this branch.

Not sure I understood the readonly-flags hierarchical propagation, is this about subcircuits? I think every file (schematic, subcircuit, data display) should have its own read-only flags.

felix-salfelder:
On Wed, May 02, 2018 at 12:28:38PM -0700, in3otd wrote:
> Not sure I understood the readonly-flags hierarchical propagation, is
> this about subcircuits? I think every file (schematic, subcircuit,
> data display) should have its own read-only flags.

(i don't see why these three should be different. in gnucap, there is
CARD_LIST, and it looks appropriate for the three purposes.)

i argue that the readonly flag should be stored in the container, not
within every single component. also, every object (down to ELEMENT?)
must know its owner. using the same data type for the container in the
cases above will make this easier.

anyway, maybe the hierarchical stuff is relevant in cases where you
right click on an instance and go "down to schematic". that's where we
should propagate the flag from the owner to the container. i don't know
how this is currently implemented in qucs, but it should be
straightforward.

andresmmera:
Hum, it smells like we're opening a can of worms :-)

I guess I'd be better if we stick to the task _"When the user opens the schematic, detect if it is read-only and disable changes"_ [(here)](https://github.com/Qucs/qucs/pull/763#issuecomment-354645727) and leave apart the _lock feature_ for the moment.

This means:

1. Qucs checks the file permissions at the time the user opens it.
2. In case the file is not writable, Qucs must disable edition: The user cannot modify the schematic with the mouse and the undo/redo/select all commands must be disabled.
3. The user may save the schematic in some other folder where he/she has write permission.

Do you agree with this?

in3otd:
> Do you agree with this?

YES!

qucs/qucs/qucs.cpp
```diff
@@ -1378,12 +1378,28 @@ bool QucsApp::gotoPage(const QString& Name)
...
+  else {
+    d = new TextDoc(this, Name);
+    if (d->isreadonly())
+    {// If the file is not writable, then add [READ-ONLY] to the tab title.
+     // The document can be edited but not saved.
+        filename += " [READ-ONLY]";
+    }
+    i = DocumentTab->addTab(dynamic_cast<TextDoc*>(d), QPixmap(icon), filename);
   }
+  DocumentTab->setCurrentIndex(i);
```
in3otd: Why all the changes above are not in `createEmptySchematic()` and `createEmptyTextDoc()`?

qucs/qucs/qucs.cpp
```diff
@@ -1589,6 +1605,7 @@ void QucsApp::slotFileSaveAll()
     // Hack! TODO: Maybe it's better to let slotFileChanged() know about Tab number?
     //   if saving was successful, turn off "saving needed" icon
     DocumentTab->setSaveIcon(!saveFile(Doc), No-1);
+    DocumentTab->setTabIcon(No-1,QPixmap(":/bitmaps/empty.xpm")); // know about Tab number?
```
in3otd: this does not seem correct, now a *Save all* clears the save icon also on the readonly file...

in3otd:
would also have been nice to have the pointer dynamic cast stuff changes and the actual functional changes in separate commits (`git add -p` can help with that)

andresmmera:
Done.
I have added an extra commit, [8bd6f0b](https://github.com/Qucs/qucs/pull/769/commits/8bd6f0b0562cf88901172822fdd980a0d510bff8), since I thought it would be convenient to ask the user to save the document into a different location when he/she tries to save a read-only schematic.

qucs/qucs/qucs.cpp
```diff
@@ -1436,7 +1436,17 @@ bool QucsApp::saveFile(QucsDoc *Doc)
     return saveAs();
 
   int Result = Doc->save();
-  if(Result < 0)  return false;
+  if(Result < 0)
+  {//The file was not saved.
+      if (Result == -2)
+      {//Save the document in another location
+          saveAs();
```
in3otd: this should be `return SaveAs()` since that function already takes care of `updatePortNumber()` and `slotUpdateTreeview()` which are done below here. Besides, currently if you cancel the "Save As" the functions continues and returns `true` and clears the diskette icon, as for a successful save.

qucs/qucs/schematic_file.cpp
```diff
@@ -343,13 +343,18 @@ int Schematic::saveSymbolJSON()
...
+// Returns the number of subcircuit ports. In case Qucs cannot save the file it returns a generic error code (-1) or a flag indicating
+// the user wants to save the document into another location
 int Schematic::saveDocument()
 {
   QFile file(DocName);
   if(!file.open(QIODevice::WriteOnly)) {
-    QMessageBox::critical(0, QObject::tr("Error"),
-				QObject::tr("Cannot save document!"));
+    int ret = QMessageBox::critical(this, QObject::tr("Error"),
+                QObject::tr("Cannot save document!\nDo you want to save it into a different folder?"), (QMessageBox::No|QMessageBox::Yes), QMessageBox::Yes);
```
in3otd: *< picky mode on>* I'll say *with a different name*, since it may still be possible to save it in the same folder, but the (full)name must be changed *< /picky mode off>*

Besides, we may avoid asking the user and simply show the MessageBox and calling `SaveAs()` afterwards anyway; `Schematic::saveDocument()` will continue to return `-1` for a failed save and `QucsApp::saveFile()` will take care of calling `SaveAs()`

qucs/qucs/qucsdoc.cpp
```diff
@@ -71,3 +71,16 @@ QString QucsDoc::fileBase (void) {
...
+bool QucsDoc::isreadonly()
+{
+    if (DocName.isEmpty())
+    {//The file wasn't created yet
+     return false;
+    }
+    else
+    {
+      QFileInfo Info(DocName);
+      return !Info.isWritable();
```
in3otd: for non-existing files `isWritable()` seems to return `false` so things do not work as expected in this case:

- open a read-only schematic
- save it with a new name
- run a simulation
- a new Data Display will open but it ill be incorrectly tagged as `[READ-ONLY]`

qucs/qucs/qucs_actions.cpp
```diff
@@ -478,8 +505,12 @@ void QucsApp::slotInsertPort(bool on)
...
 void QucsApp::slotEditUndo()
 {
-  Schematic *Doc = (Schematic*)DocumentTab->currentWidget();
-  if(isTextDocument(Doc)) {
+    assert(DocumentTab);
+    QWidget* w = DocumentTab->currentWidget();
+    Schematic *Doc = dynamic_cast<Schematic*>(w);
+    assert(Doc);
+
+    if(isTextDocument(Doc)) {
```
in3otd: wrong indent here :grin:

qucs/qucs/qucsdoc.cpp
andresmmera: Reproduced. Thanks!

andresmmera:
I've just corrected the issue you spotted: https://github.com/Qucs/qucs/pull/769/commits/923e81e8947aae0a73de3746c03000b827e485be#diff-2db52a5f64aeeea81fd7b1231486253cR84
It simply checks if the file exist, if not it returns "false".

Indentation corrected. I think it's ok now: https://github.com/Qucs/qucs/pull/769/commits/5277f318093049dea80e2ab9233bc114d9594390#diff-d85a0f8c94a4eee7d0284775e8e30f25R508

PR #770, opened 2018-01-20
Update the Project Content view automatically

in3otd:
The Project Content Tab should always show the actual directory content but this was not always the case since its update was manually triggered.
There were some actions that changed the directory content but did not update the view, e.g.:
- run a simulation
- change the destination Dataset name
- rerun the simulation
- note that the new Dataset does not show up in the Project content

also the view was not updated when copying something into the Project directory outside of Qucs (like a schematic or other files).
Moreover the manual update reinitialized the view completely, forcing to show only the Schematics content even if some other rows were manually expanded previously.

This PR solves the above issues.

### DIFF
--- a/qucs/qucs/dialogs/qucssettingsdialog.cpp
+++ b/qucs/qucs/dialogs/qucssettingsdialog.cpp
@@ -666,7 +666,6 @@ void QucsSettingsDialog::slotApply()
     if(changed)
     {
         App->readProjects();
-        App->slotUpdateTreeview();
         App->repaint();
     }
 
--- a/qucs/qucs/projectView.cpp
+++ b/qucs/qucs/projectView.cpp
@@ -28,17 +28,20 @@
 #include <QStringList>
 #include <QDir>
 #include <QStandardItemModel>
+#include <QFileSystemWatcher>
 #include <QDebug>
 
 ProjectView::ProjectView(QWidget *parent)
   : QTreeView(parent)
 {
   m_projPath = QString();
-  m_projPath = QString();
+  m_projName = QString();
   m_valid = false;
   m_model = new QStandardItemModel(8, 2, this);
+  watcher = new QFileSystemWatcher(this);
+  connect(watcher, SIGNAL(directoryChanged(const QString&)), SLOT(dirChanged(const QString&)));
 
-  refresh();
+  init();
 
   this->setModel(m_model);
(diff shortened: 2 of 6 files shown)


andresmmera:
I've been testing this PR for a few days (in Ubuntu and Windows 10) and I didn't find any significant issue. However, I've noticed the following "minor issue" in Windows (in Ubuntu works fine):

**Sometimes** after copying a schematic (or any other file) into the project with the Windows file explorer, Qucs takes a few seconds until it updates the project tree. I repeated the test in two different computers and the result was the same...

After reviewing the code, I concluded that the problem might lie in the Qt libs side rather than in your code so I would absolutelly ignore this issue...

If you're not planning to work more on this PR I think it could be merged (if nobody objects)

in3otd:
thanks; is Qucs still responsive while waiting for the file to appear?
I tested it only with Linux and there, as you also saw, it worked well.

andresmmera:
Yes, Qucs is responsive. I reproduced the issue right now and this time I had to reload the project to see the new file on the project tree.

in3otd:
ok, googling around it seems that similar issues with `QFileSystemWatcher()` have been seen before on Windows. A suggested solution/workaround is to add a timer for refreshing, just to be sure; I guess a 1 s refresh timer could be reasonable.

in3otd:
the Qt docs [say](https://doc.qt.io/archives/qt-4.8/qfilesystemmodel.html#caching-and-performance) that `QFileSystemModel()`, used by the Project Viewer, also uses `QFileSystemWatcher()` so there should be a similar issue for the list of projects shown there. Can you check if this is the case, by adding/removing a directory ending in `_prj`? I'll also try to check this later.

andresmmera:
I've just checked that. However, in this case the project tree is updated
at the same moment I do the cut/paste/delete operations from the Windows
file explorer.

in3otd:
I've tested this on Windows 10 and didn't see any issue with the project tree update; when copying or deleting files using the File Explorer or the command line the update in Qucs was instantaneous. Is there any typical action that seems to result in a delayed update on your machines or does this happen at random?

andresmmera:
Well, I would say this stuff happens randomly under normal condition. It doesn't happen every time I copy/paste files from the Windows file explorer, but I saw that quite often, roughly, one out of four times I do file operations.

in3otd:
rebased

PR #775, opened 2018-02-10
Initial condition label updates

mbharat23:
This patch appends "init" to the labels of the initial conditions for the capacitor
and inductor.  Also fixed a typo in capacitor class implementation.

### DIFF
--- a/qucs-core/src/components/capacitor.cpp
+++ b/qucs-core/src/components/capacitor.cpp
@@ -25,7 +25,7 @@
 /*!\file capacitor.cpp
    \brief capacitor class implementation
 
-   A capacitor is a passive device that strore electric energy
+   A capacitor is a passive device that stores electric energy
 */
 
 
--- a/qucs/qucs/components/capacitor.cpp
+++ b/qucs/qucs/components/capacitor.cpp
@@ -24,7 +24,7 @@ Capacitor::Capacitor()
 
   Props.append(new Property("C", "1 pF", true,
 		QObject::tr("capacitance in Farad")));
-  Props.append(new Property("V", "", false,
+  Props.append(new Property("Vinit", "", false,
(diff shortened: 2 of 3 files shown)


felix-salfelder:
`Vinit` seems wrong. see e.g. `qucs-core/src/components/capacitor.cpp` line 142.

there's a real problem to it: `Props` must be private. (with that fixed, you could easily introduce aliases.)

in3otd:
um, I think this will break backward compatibility, as it changes the `qucsator` netlist format. If one uses (library) subcircuits with capacitors or inductors they will not work anymore.
Sure, we could handle this as a special case but honestly I do not see a good reason to do this change at all; what `V` and `I` stands for is explained in the Propery description, shown in the Edit Property dialog box. If further explanations are needed, those should not go into the Property name but into a proper Help document, the [Qucs Reference Manual](https://qucs.github.io/qucs-manual/0.0.19/html-en/index.html) (and, yes, maybe one day I'll finish #448, so that help can easily be accessed from the component Edit Property dialog).

mbharat23:
Thanks for the clarifications all. 

I suggest to withdraw this pull request.  I think the best way to contribute to QUCS development is to look through the various issues and find one that interests me. 

How do I go about "abandoning" this pull request?

Thx.

Bharath

in3otd:
No problem Bharath, it's a good thing you previously sent your proposals to the qucs-devel mailing list, unfortunately we were slow to follow up so we couldn't start a discussion there. I'll try to comment further there later.

You should see a button *Close and comment* under here, this will mark it as closed and make it disappear from the list of things to review/merge.

mbharat23:
Closing (abandoning) this change as the approach is incorrect and overall patch is unnecessary.

PR #785, opened 2018-03-12
Develop

eroen:
While updating the packaging of Qucs for Gentoo Linux I needed to make some
changes to the code. Please review and consider applying the attached patches.

I don't know all the reasons behind the existing design, so please excercise
due caution. :-)


felix-salfelder:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=726021

Shared library is no longer provided due to upstream recommendation.
From README.Debian:

         Use of precompiled libgtest Not Recommended
         -------------------------------------------

 The Google C++ Testing Framework uses conditional compilation for some
 things.  Because of the C++ "One Definition Rule", gtest must be
 compiled with exactly the same flags as your C++ code under test.
 Because this is hard to manage, upstream no longer recommends using
 precompiled libraries [1].

eroen:
That's an ...interesting statement, but sure, I'll do it differently.

felix-salfelder:
well.. perhaps the reason why it's done that way. otherwise, i don't know what problem(s) gtest is introducing/solving here.

is there no source package for gentoo?

the rest: looks good. thanks for cleaning up.

eroen:
Nevermind the gtest stuff, the `--with-gtest=` option makes things easy if not exactly elegant. There is a source package for qucs 0.0.18, the plan was to update it and I figured I'd send upstream our changes.
Thanks for your feedback btw!

in3otd:
I was just browsing the AppVeyor build logs and apparently after this PR there is a warning/error
```
libtool:   error: Could not determine the host path corresponding to
libtool:   error:   'C:/msys64/mingw64/lib:C:/msys64/mingw64/bin'
libtool:   error: Continuing, but uninstalled executables may not work
```
This was not there before, compare e.g. [this](https://ci.appveyor.com/project/qucs/qucs/build/1.0.1582/job/sm4q4hd4fy4uhypq?fullLog=true#L2578) and [this](https://ci.appveyor.com/project/qucs/qucs/build/1.0.1585/job/8wrii6d5yg85k10s?fullLog=true#L2579).
I've tried to run the 64 bit executable on Windows and it appears to run fine, though.

felix-salfelder:
thanks for spotting this. i had a closer look.

    -AC_SUBST(QT_LDFLAGS)
    +QT_LIBS="${QT_LIBS} ${QT_LDFLAGS}"

    -qucs_real_LDFLAGS = $(QT_LDFLAGS)
    +qucs_real_LDFLAGS = $(QT_LIBS)

seems totally wrong. *_LDFLAGS ends with "LDFLAGS" because that's the place where the flags for the linker (should) go. will try and sort it, or revert the change.

PR #787, opened 2018-03-18
trying to fix LDFLAGS issue #785

felix-salfelder:
see #785

### DIFF
--- a/qucs/configure.ac
+++ b/qucs/configure.ac
@@ -25,6 +25,9 @@ AC_PROG_CC
 AC_CHECK_TOOL(AR, ar, :)
 LT_INIT()
 
+# llvm spams the log, breaks CI
+CXXFLAGS="$CXXFLAGS -Wno-unused-command-line-argument"
+
 dnl Check for debugging option.
 AC_ARG_ENABLE([debug],
   AC_HELP_STRING([--enable-debug],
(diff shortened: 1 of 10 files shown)


guitorri:
I think the original defaults were left as a fall-back? Very long ago, the Qt installers for OSX did not ship/install pkg-config files. I don't know if that is still the case.

guitorri:
You also seem to have updated the qucs-test submodule on the first commit.

felix-salfelder:
On Sun, Mar 18, 2018 at 09:47:42PM +0000, Guilherme Brondani Torri wrote:
> You also seem to have updated the qucs-test submodule on the first commit.

eek, that was in error. thanks for the spot!

felix

felix-salfelder:
travis is unhappy.

    no output has been received in the last 10m0s

relation to LDFLAGS and QT_LIBS unclear to me. will have another look, later

in3otd:
I think this is due to the `qucs-test` submodule "update", it seems it's now pointing at a version with a `skip.txt` file which does not including schematic with removed components (like the phasor diagram) so qucs is complaining when opening the schematic and shows a Message Box about that but on Travis nobody can press the *OK* button...
Not sure why the build appears to continue after trying to convert that schematic, the outputs seems a bit out of sync (?)

felix-salfelder:
thanks for the reminder. that makes sense. i thought i had fixed the qucs-test mistake already `:|`

PR #801, opened 2018-05-20
[WIP] Windows portable package update

in3otd:
currently just playing with AppVeyor, mainly to see how to build a more complete portable package (including ASCO, adms, etc.)


in3otd:
that 's all for now. Included is a patch for the ASCO optimizer which is needed on Windows; I sent it to the ASCO maintainer some time ago but didn't get a response so far.

in3otd:
yes, in fact I was unsure where to stop adding stuff. On one side it would be nice to have AppVeyor build a complete package with "everything", including the compiler and even maybe GNU Octave but this should probably be done conditionally only on `develop` or even just on release tags. Maybe in the future...

PR #812, opened 2018-07-13
Inhibit Simulation Messages Window

devel0:
- per document settings
- a warning message still appears in statusbar if some warnings detected

### DIFF
--- a/qucs/qucs/dialogs/settingsdialog.cpp
+++ b/qucs/qucs/dialogs/settingsdialog.cpp
@@ -88,6 +88,9 @@ SettingsDialog::SettingsDialog(Schematic *Doc_)
                                     Tab1);
     gp->addWidget(Check_RunScript,4,0,1,2);
 
+    Check_SimInhibitWindow = new QCheckBox(tr("inhibit sim window"), Tab1);
+    gp->addWidget(Check_SimInhibitWindow,5,0,1,2);
+
     t->addTab(Tab1, tr("Simulation"));
 
     // ...........................................................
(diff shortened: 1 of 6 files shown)


devel0:
I found useful to allow inhibit of simulation window that appears when press F2 ; specifically this patch allow to override per-project and disable automatic window from appearing ; if there are warnings a statusbar already shows a blink red message about warnings in the simulation log.

guitorri:
I don't think this should be stored in the document settings.
For me an option on the app settings is better suited.

devel0:
I agree but I followed previous concept that already exists for **"open data display after simulation"** that's per document setting. Btw I could change to application settings if this patch will be integrated.

in3otd:
IIRC, there were some discussions in the past about what to do with the Simulation Messages Window (SMW), surely its current behavior can be annoying sometimes.

One reason to always show the SMW is that it contains the Progress Bar, besides the various info messages, so the user knows if and how fast/slow the simulation is progressing. But when the simulation finishes without **errors** I think the SMW should automatically disappear (close/minimize) in any case. If there are warnings the statusbar message will inform the user.
IIRC, this is what happens in the default case when  *open data display after simulation* is checked; when this is unchecked, usually because the Graphs were placed on the Schematic page, the SMW stay in the foreground and needs to be closed manually.

So I'll propose to modify the SMW behavior as above:
- SWM is always shown in the foreground when a simulation starts
- when simulation ends
  - if no errors, SMW is closed
  - if errors it stays in the foreground

the only disadvantage I see is that for fast simulations you will just see something flickering in foreground, since the SMW will open and close quickly.
The inhibit options could still be added - but I see it more document dependent :grin:

devel0:
I was added this feature because I had a very simple transistor circuit with a current probe at the base and I wanted to see the graph changing respect the previous when changing base resistor and press F2; appearing of the dialog just disallowed me to see graph changes with ease for a fast inspection of variations.

PR #813, opened 2018-07-14
one more regression test

felix-salfelder:


guitorri:
Isn't it included in #814 ?

felix-salfelder:
On Sun, Jul 15, 2018 at 02:36:36PM -0700, Guilherme Brondani Torri wrote:
> Isn't it included in #814 ?

#813 is older. it was meant to be a single PR, but i wasn't so sure
about #814.

rebased both of them.

felix-salfelder:
fixed in refactor branch

PR #814, opened 2018-07-15
fold in tests from cmake

felix-salfelder:
trying to make qucs/tests work with autotools

### DIFF
--- a/qucs/qucs/tests/SchematicTests.cpp
+++ b/qucs/qucs/tests/SchematicTests.cpp
@@ -4,6 +4,7 @@
 #include "SchematicTests.h"
 
 #include <QTest>
+#include <QApplication>
 
 void SchematicTests::testConstructor()
 {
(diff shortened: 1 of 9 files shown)


qucs/qucs/tests/SchematicTests.cpp
```diff
@@ -1,13 +1,12 @@
 
 #include "qucs.h"
 #include "schematic.h"
+#include "SchematicTests.h"
```
guitorri: Why do you have to split the header?

felix-salfelder:
On Sun, Jul 15, 2018 at 02:36:17PM -0700, Guilherme Brondani Torri wrote:
> guitorri commented on this pull request.
> 
> 
> 
> > @@ -1,13 +1,12 @@
>  
>  #include "qucs.h"
>  #include "schematic.h"
> +#include "SchematicTests.h"
> 
> Why do you have to split the header?

moc only needs to see the class declaration. this is how it is done in
the other places (more or less). common.mk.

it looks right to me, and didn't want to reinvent the wheel.

PR #816, opened 2018-07-16
Q3ptrlist

felix-salfelder:
intermediate Q3ptrlist substitute. possibly only relevant during refactoring.

### DIFF
--- a/qucs/common.mk
+++ b/qucs/common.mk
@@ -32,3 +32,5 @@ SUFFIXES += .moc.cpp
 
 .h.moc.cpp:
 	$(MOC) -o $@ $<
+
+$(MOCFILES): $(top_builddir)/qt_version
--- a/qucs/configure.ac
+++ b/qucs/configure.ac
@@ -52,6 +52,7 @@ if test "$enable_qucs_gui_debug" = yes; then
   AC_DEFINE(DEBUG, 1, [Enable debug build.])
   QT_DEF="-DQT_DEBUG"
   CXXFLAGS="$CXXFLAGS -g -O0"
+  CPPFLAGS="$CPPFLAGS -DTRACE_UNTESTED"
 else
   AC_DEFINE(NDEBUG, 1, [Disable debug build.])
   QT_DEF="-DQT_NO_DEBUG -DQT_NO_DEBUG_OUTPUT"
@@ -433,6 +434,7 @@ case $host_os in
     QT_LDFLAGS="$QT_LDFLAGS -headerpad_max_install_names"
 
     if test "$enable_qucs_gui_debug" = yes; then
+      # BUG. -W and -Wall are not CPPFLAGS
       CPPFLAGS="$CPPFLAGS -W -Wall"
     fi
 
--- a/qucs/qucs/Makefile.am
+++ b/qucs/qucs/Makefile.am
@@ -95,7 +95,7 @@ qucs_real_LDADD = \
 	libqucsschematic.la
 
 noinst_HEADERS = $(MOCHEADERS) wire.h qucsdoc.h element.h node.h \
-	qt_compat.h \
+	qt_compat.h trace.h \
   wirelabel.h viewpainter.h mnemo.h mouseactions.h syntax.h module.h misc.h \
   projectView.h printerwriter.h imagewriter.h
 
--- a/qucs/qucs/Makefile.am
+++ b/qucs/qucs/Makefile.am
@@ -95,6 +95,7 @@ qucs_real_LDADD = \
 	libqucsschematic.la
 
 noinst_HEADERS = $(MOCHEADERS) wire.h qucsdoc.h element.h node.h \
+	container.h \
   wirelabel.h viewpainter.h mnemo.h mouseactions.h syntax.h module.h misc.h \
   projectView.h printerwriter.h imagewriter.h
 
--- a/qucs/qucs/Makefile.am
+++ b/qucs/qucs/Makefile.am
@@ -95,7 +95,7 @@ qucs_real_LDADD = \
 	libqucsschematic.la
 
 noinst_HEADERS = $(MOCHEADERS) wire.h qucsdoc.h element.h node.h \
-	container.h \
+	qt_compat.h \
   wirelabel.h viewpainter.h mnemo.h mouseactions.h syntax.h module.h misc.h \
   projectView.h printerwriter.h imagewriter.h
 
--- a/qucs/qucs/components/Makefile.am
+++ b/qucs/qucs/components/Makefile.am
@@ -25,6 +25,8 @@
 
 noinst_LTLIBRARIES = libcomponents.la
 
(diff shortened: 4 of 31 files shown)


felix-salfelder:
okay, things conflict a bit more than i was hoping. i am trying to get the iteration loops in a shape that we do not have to edit again and again... 

my hope is to agree on the changes here (some more work), merge when green (and #814 etc), and then rebase qt5-1 once more.

felix-salfelder:
... the crashes seem to come from Props, which is a Q3Ptrlist. so this is a kind of blocker anyway `:|`

felix-salfelder:
`\o/`

guitorri:
IMHO we should branch `release-0.0.20` before we start breaking stuff on develop. From now on we could make the `qt5-*` branches more like `unstable develop` kind of thing. Until we get qt5 in useable shape to merge back into `develop`. 

Rebasing all the time can be a pain. On `qt5-1` you seem to have merged qt5 onto develop, squashed some commits and rebased. In this way all my branches which are older and have a longer history are conflicting at every commit. Maybe merging is more practical and we can do always with PRs.

felix-salfelder:
> On qt5-1 you seem to have merged qt5 onto develop, squashed some commits and rebased

i did not quash anything. maybe a few commit ended up empty?

> Maybe merging is more practical and we can do always with PRs.

that's what i am doing, no? i wouldn't mind doing 0.0.20 first... it would be nice if 0.0.20 worked for qt4 without qt3support though..

guitorri:
I am pretty sure I did not put the `(also introduce "elemType()" to finally get rid of it.)` into  e469989f791845308a376eb258e8c095cdf05945. Maybe because of the rebase...

Did you try to disable qt3support  on after getting rid of q3ptrlist?

felix-salfelder:
On Wed, Jul 18, 2018 at 12:54:57AM -0700, Guilherme Brondani Torri wrote:
> I am pretty sure I did not put the `(also introduce "elemType()" to finally get rid of it.)` into  e469989f791845308a376eb258e8c095cdf05945. Maybe because of the rebase...

possibly -- yes. the problem with elemType type() and Type is that we
need to settle on one. ideally we should just get rid of it. if we do
that, theres no point arguing about comments in the code ...

(NB: a proper solution is on the way)

> Did you try to disable qt3support  on after getting rid of q3ptrlist?

one thing at a time.. my strategy is as follows:

- look at qt5-$revision can do (e.g. find bugs in container.h,
  QrectF Boundingbox e.g.)
- fix the issue, file a PR.
- see if the tests pass (which is only possible on develop).
- try and find a reviewer ...

i understand how you feel uncomfortable with adding changes to develop.
i like the 0.0.20 branch idea. maybe we should agree on a plan, such as

- leave 0.0.20 behind, status-quo
- target the PRs that do not need qt5 to develop, but leave the choice.
- keep CI green in the process
- keep rebasing qt5-$revision onto develop
- create a 0.0.21 before merging stuff that requires qt5

i do feel your pain about the rebase. but there are so many bugs in
qt5-$r, and not even CI. i cant work with it. (my only PR targeted at
qt5-$r is for exploration only).

for example, finding the bugs in container.h was only possible after
including more tests and after rebasing to develop.

we will have to fix those bugs in any case. my hope is that
qt5-$revision will only shrink and finally be reviewable.

felix-salfelder:
On Wed, Jul 18, 2018 at 12:54:57AM -0700, Guilherme Brondani Torri wrote:
> Did you try to disable qt3support  on after getting rid of q3ptrlist?

looking at it, with some more of Eduards commits, this should basically work.

what doesnt seem to work is getting rid of Q3Ptrlist entirely, hence i
did with container.h. note that some of the segfaults in qt5-1 are bugs
in container.h..

felix-salfelder:
i have tested this in qt5-* (see #820) and it works (i.e. fixes all the q3Ptrlist bugs)

also, i have created a 0.0.20 branch, so you can safely merge this into develop.

in3otd:
um, it's normal that this crashes when trying to place a wire?

felix-salfelder:
thanks. indeed another bug in q3ptrlist.

in3otd:
thanks, it works better but by drawing some wires and moving them around I can get it to crash still.
`gdb` says
```
#0  0x000000000048639b in Schematic::copySelectedElements(Q3PtrList<Element>*) (this=0x2ccb740, p=0x29526f0)
    at schematic_element.cpp:1759
#1  0x0000000000672a49 in MouseActions::MMoveMoving(Schematic*, QMouseEvent*) (this=0x29526b0, Doc=0x2ccb740, Event=0x7fff683e7070) at mouseactions.cpp:443
#2  0x000000000047329b in Schematic::contentsMouseMoveEvent(QMouseEvent*) (this=0x2ccb740, Event=0x7fff683e7070)
    at schematic.cpp:526
#3  0x00007f4a7f82aa2c in Q3ScrollView::viewportMouseMoveEvent(QMouseEvent*) ()
    at /usr/lib64/qt/lib/libQt3Support.so.4
```

felix-salfelder:
On Fri, Jul 20, 2018 at 09:34:29AM -0700, in3otd wrote:
> thanks, it works better but by drawing some wires and moving them around I can get it to crash still.
> `gdb` says

thanks. since the schematic code did not change, the bug must be in the
compat header.

i will have a look...

felix-salfelder:
On Fri, Jul 20, 2018 at 09:34:29AM -0700, in3otd wrote:
> thanks, it works better but by drawing some wires and moving them around I can get it to crash still.
> `gdb` says

this turns out to be a can of worms.

look at componentdialog.cpp+431

`for(Property *p = Comp->Props.at(Comp->Props.find(pp)+1); p != 0; p = Comp->Props.next())`

it's not just ridiculous, but also Comp->Props.find(pp) is not even
defined.

according to the manual, Q3Ptrlist::find uses Q3Ptrlist::compareItems,
and compareItems uses "==". but "==" is not implemented for "Property".

any good guess what the intent of the code in componentdialog could be?

ideally something that does not use `find`...

felix-salfelder:
On Fri, Jul 20, 2018 at 11:27:35AM -0700, in3otd wrote:
> well, the `componentdialog.cpp` is *a bit* messy, yes.
> But the comparison is not about pointers, `Property *` ??

it is possible. but they also have `findRef` for this. the documentation
says that `findRef` is *faster* because it "does not use a compareItem"
function.

> As the comment below tries to explain, I think that for simulation
> "components" a few properties are shown in a separate tab ("Sweep"),
> then the first property to put on the next tap is pointed by `pp`.
> Maybe, if that is not a simulation component `pp` is null so all the
> properties are then put on the usual properties (single) tab. Just a
> wild guess.

thanks. i will have a look, expect more confusion `;)`

... if you plan to test more: --enable-debug now enables testing, that
makes it easier to track down crashes.

felix-salfelder:
part of refactor branch

PR #817, opened 2018-07-18
remove qt3support dependency WIP

felix-salfelder:
this removes the dependency on qt3, but lacks some mouse functionality


felix-salfelder:
obsolete

PR #820, opened 2018-07-18
Qt5-3+q3ptr

felix-salfelder:
actually, merge q3ptr and resolve conflicts

this should fix some of the segfaults in container.h (now qt3_compat.h)


felix-salfelder:
.. yes it does. and i found another bug in Component()::Component().

felix-salfelder:
cmake woes on travis. i tried not to touch cmake files though..

PR #827, opened 2018-07-21
Mouse refactor

felix-salfelder:
protect coordinates in Element. and fix some type issues.

this is excessively used in mouseactions. doing it first, so it can be tested/bisected


felix-salfelder:
ok, hold on. this messes up some of the coordinates.

the previous code had 3 sets of coordinates per mouse event. something broke when i removed the extra arguments float fX, fY as in qt5-* (totally makes sense) and use Event->pos instead.

felix-salfelder:
part of some qt5 branch

PR #832, opened 2018-07-25
configure in qucs

felix-salfelder:
- Romano Giannetti found a whitespace issue and proposed a fix.
- list pkg-config as a dependency.

### DIFF
--- a/qucs/README
+++ b/qucs/README
@@ -36,6 +36,8 @@ Requirements
 
 Qucs needs Qt(R) by Digia(R) <http://www.digia.com>.
 This current version is known to work with Qt 4.6 and later.
+For simplicity, the Qt configuration is currently obtained through pkg-config,
+which you need to install in addition to Qt.
 
 
 Installation
--- a/qucs/README
+++ b/qucs/README
@@ -34,10 +34,9 @@ http://qucs.sourceforge.net
 Requirements
 ============
 
-Qucs needs Qt(R) by Digia(R) <http://www.digia.com>.
-This current version is known to work with Qt 4.6 and later.
-For simplicity, the Qt configuration is currently obtained through pkg-config,
-which you need to install in addition to Qt.
+Qucs needs Qt version 4.6 or later. For simplicity, the Qt configuration is
+currently obtained through pkg-config, which you need to install in addition
+to Qt.
 
 
 Installation
--- a/qucs/configure.ac
+++ b/qucs/configure.ac
@@ -236,7 +236,7 @@ case $host_os in
 
     dnl Enable C++11 support
     CXXFLAGS="$CXXFLAGS -std=c++0x"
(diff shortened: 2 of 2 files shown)


qucs/README
```diff
@@ -36,6 +36,8 @@ Requirements
 
 Qucs needs Qt(R) by Digia(R) <http://www.digia.com>.
```
in3otd: not related but Qt development is now done by the Qt Group Plc, https://www.qt.io :grin:

guitorri:
No need to update qucs-test.

felix-salfelder:
On Wed, Jul 25, 2018 at 11:58:32AM -0700, Guilherme Brondani Torri wrote:
> No need to update qucs-test.

thanks, fixed.

(still don't understand how this always happens since recently. i never
touched any of this...)

qucs/configure.ac
```diff
@@ -236,7 +236,7 @@ case $host_os in
 
     dnl Enable C++11 support
     CXXFLAGS="$CXXFLAGS -std=c++0x"
-    PKG_CHECK_MODULES([QTest], [${qt_name}Test ])
+    PKG_CHECK_MODULES([QTest], [${qt_name}Test])
```
in3otd: by the way, any idea why this works with and without the space on my system and on Travis too?
Here everything also works fine also if I remove this line, so in which case is it needed??

Trying to understand what `PKG_CHECK_MODULES` does, I saw that "[this guy](http://tirania.org/blog/archive/2012/Oct-20.html)" and others discourage its usage; FWIW, I knowing nothing about it.

guitorri:
If you don' t `git update submodule` and `git add -a` you also pick up the old `qucs-test` hash.

The QtTest is needed if you run `make check`.

felix-salfelder:
On Wed, Jul 25, 2018 at 07:35:42PM +0000, in3otd wrote:
>      dnl Enable C++11 support
>      CXXFLAGS="$CXXFLAGS -std=c++0x"
> -    PKG_CHECK_MODULES([QTest], [${qt_name}Test ])
> +    PKG_CHECK_MODULES([QTest], [${qt_name}Test])
> 
> by the way, any idea why this works with and without the space on my
> system and on Travis too?

don't know, perhaps it's actually a bug in an older pkg.m4. I wouldn't
have put the space in, if it had not worked for me..

> Trying to understand what `PKG_CHECK_MODULES` does, I saw that "[this
> guy](http://tirania.org/blog/archive/2012/Oct-20.html)" and others
> discourage its usage; FWIW, I knowing nothing about it.

interesting article, I never gave it a thought. otoh, if
PKG_CHECK_MODULES only induces trouble once every 10+ years, and only due
to stupid whitespace, then why bother? it's just as easy to get a shell
script wrong in a subtle way, YMMV.

PR #837, opened 2018-07-26
Correct closeAllFiles() return value 

### LINKED ISSUE #834: Change Qucs Home lead to empty schematic tab, unusable
Steps to reproduce:

- Build `release-0.0.20` branch and run Qucs
- Open one project/schematic from current Qucs Home location
- Change Qucs Home
  - File > App Settings > Locations (tab) > [Browse to another Qucs Home] > OK
- Problems 
  - schematic tab becomes empty, no Tab widget on view. It is not possible to open projects from the new Qucs Home.
  - app cannot be closed (File > Quit)

Tested on macOS 10.12.6

in3otd:
Fixes #834

A bug was introduced in #708 (62ff9b0f417285a49f4c4dfb0913394d59b8216f) as the `closeAllFiles()` function return value when there were no tabs open was accidentally changed from `true` to `false`.

### DIFF
--- a/qucs/qucs/dialogs/qucssettingsdialog.cpp
+++ b/qucs/qucs/dialogs/qucssettingsdialog.cpp
@@ -656,8 +656,9 @@ void QucsSettingsDialog::slotApply()
     // if QucsHome is changed, refresh projects tree
     // do this after updating the other paths
     if (homeDirChanged) {;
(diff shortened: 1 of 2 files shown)


guitorri:
Ideally bugfixes should go on the release branch (if there is one), and then merged back into develop.
I guess I will just cherry-pick. Thank you for the quick fix.

PR #838, opened 2018-07-26
Correct closeAllFiles() return value

guitorri:
when there are no open tabs.
This is a corner case that currently happens only when
changing the Qucs Home directory.

### DIFF
--- a/qucs/qucs/dialogs/qucssettingsdialog.cpp
+++ b/qucs/qucs/dialogs/qucssettingsdialog.cpp
@@ -656,8 +656,9 @@ void QucsSettingsDialog::slotApply()
     // if QucsHome is changed, refresh projects tree
     // do this after updating the other paths
     if (homeDirChanged) {;
-      // files were actuallt closed above, this will refresh the projects tree
+      // files were actually closed above, this will refresh the projects tree
       // and create an empty schematic
+      // (this could likely be simplified once the automatic Project Content Tab refresh is implemented)
       App->slotMenuProjClose();
       changed = true;
     }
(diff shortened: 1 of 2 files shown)


felix-salfelder:
if closeTabsRange returns a kind of success state, then it must be "true" on empty ranges.

so, the bug is there, not in closeAllFiles (where you fixed it).

in3otd:
well, at the beginning I did exactly what you suggest, changing the return value from `false` to `true` [here](https://github.com/Qucs/qucs/blob/develop/qucs/qucs/qucs.cpp#L1712). While it seemed to work it didn't feel right to me; besides that I must have chosen that return value for a reason, which currently I don't remember, `closeAllFiles()`  called `closeTabsRange(0, -1)` and I thought that in this case it's right that this latter return `false` since no tabs were really closed. So I did not see the `(0,-1)` as an empty range but more as an as an invalid range. OTOH it seemed correct to me that `closeAllFiles()` returns `true` (success) when no files are actually open. This is also what I wanted to explain with the added comments on the return values but I now see that they are not that clear.

in3otd:
the difference is only in how I considered the range: "close from tabs 0 to tabs -1" sounded like an invalid request, while asking to "close all files" while there is none open seems a valid thing to do to me.

felix-salfelder:
On Fri, Jul 27, 2018 at 11:17:46AM -0700, in3otd wrote:
> the difference is only in how I considered the range: "close from tabs
> 0 to tabs -1" sounded like an invalid request.

that's a different issue.

in the rest of the C/C++ world the range (x,y) denotes the items x, x+1,
x+2 ... y-1. then, (0,0) would be a possible empty range, and does not
sound invalid at all. an easy fix, not a reason to get confused.

guitorri:
Feel free to continue this discussion targeting fixes on develop.

PR #845, opened 2018-08-13
Improve qucs attenuator tool

andresmmera:
Originally, the aim of this PR was simply to add some extra space between the components in order to make more readable the resistor values:

image

However, I thought it would be nice to extend the tool capabilities a little bit, so I included a feature to calculate the power dissipated by each resistor:

image

And also added some microwave attenuators:

- Reflection attenuator.

- Quarter wavelength series attenuator (unmatched).

- Quarter wavelength shunt attenuator (unmatched).

Finally, the Qlineedit were converted into QSpinboxes and the attenuator synthesis is triggered whenever the design parameters change. This way, the user can see how the resistor values and the power dissipated vary depending on the attenuation and the input power.

The following workspace contains a few schematics which can be used to test the power dissipation feature:
https://drive.google.com/open?id=1ybKfNxXeOfplV90CycLYmDE4lNhi99zi

Despite the fact that the attenuator docs were updated, I've written some notes which cover the same topics in more detail:

[Notes on resistive attenuator design.pdf](https://github.com/Qucs/qucs/files/2282354/Notes.on.resistive.attenuator.design.pdf)

**References:**
[1] RF design guide. Systems, circuits, and equations. Peter Vizmuller. Artech House, 1995
[2] The PIN diode circuit designer's handbook. W.E. Doherty, Jr., R.D. Joos, Microsemi Corp., 1998


andresmmera:
I'm failing to see why clang c++ is failing... 
Basically, I inspected the [log](https://api.travis-ci.org/v3/job/415366908/log.txt) a few times with focus on the qucs-attenuator stuff and searched for some keywords (like 'error', 'fail', etc.), but without success so far. Do you know a better way to find out where the problem is?

in3otd:
it seems to me that the problem is in building the docs, only the `clang` build on Travis is configured to build them.
The log ends with
```
Transcript written on technical.log.
make[3]: *** [technical.dvi] Error 1
```
and before I see
```
! Undefined control sequence.
l.369 \nointent
                There are two solutions for $R_i$:
[223] <Relative-dissipated-reflection-attenuator.eps> [224]
```
and that **`\nointent`** may be related to the build failure; can you build the docs locally ?

andresmmera:
#&!@#~ Since \nointent is not a latex tag, it may be the responsible for this.
I can compile the docs locally in two steps: It always fails in the first attempt, but when I try a second time, it seems to work and generates the pdf...

andresmmera:
Many thanks @in3otd !
After reviewing 'technical.log' I found two more typos... Now I have a clean installation locally and the Travis build seems to be fine.

PR #848, opened 2018-08-19
deduplicate cursorUpDownLeftRight

felix-salfelder:
qucs_actions needs adjustments for qt5...


felix-salfelder:
fixed somewhere else

PR #850, opened 2018-08-21
workaround showName

felix-salfelder:
implement behaviour in Component, not in parser.

this fixes the GND issue found in #757

### DIFF
--- a/qucs/qucs/components/component.cpp
+++ b/qucs/qucs/components/component.cpp
@@ -798,13 +798,10 @@ Component* Schematic::loadComponent(const QString& _s, Component* c) const
   }
   s = s.mid(1, s.length()-2);   // cut off start and end character
 
-  QString n;
-  if(c->name() == "*"){
-    c->obsolete_name_override_hack("");
-  }else{
-    c->obsolete_name_override_hack(s.section(' ',1,1));
-  }
+  QString label=s.section(' ',1,1);
+  c->setName(label);
 
+  QString n;
   n  = s.section(' ',2,2);      // isActive
   tmp = n.toInt(&ok);
   if(!ok){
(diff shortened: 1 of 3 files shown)


in3otd:
the GND appearance in the GUI is fine now but the generated netlist when copying/pasting a GND is still somewhat different:
before a typical netlist line for GNDs was
``<GND * 5 400 460 0 0 0 0>``
``<GND * 5 420 460 0 0 0 0>``
while now they still have a sort of incremental reference attached
``<GND *1 5 260 520 0 0 0 0>``
``<GND *2 5 280 520 0 0 0 0>``
``<GND *3 5 300 520 0 0 0 0>``
it does not seem to cause any issue, though.

felix-salfelder:
On Wed, Aug 22, 2018 at 05:07:48AM -0700, in3otd wrote:
> the GND appearance in the GUI is fine now but the generated netlist when copying/pasting a GND is still somewhat different:
> before a typical netlist line for GNDs was
> ``<GND * 5 400 460 0 0 0 0>``
> ``<GND * 5 420 460 0 0 0 0>``
> while now they still have a sort of incremental reference attached
> ``<GND *1 5 260 520 0 0 0 0>``
> ``<GND *2 5 280 520 0 0 0 0>``
> ``<GND *3 5 300 520 0 0 0 0>``

thanks. yes i have anticipated some change. i think this one is to the
better.

> it does not seem to cause any issue, though.

both having unique labels, and allowing multiple components with the
same label has pros and cons. I can't see any benefit in doing it
randomly, based on the component type or name.

generally, all components should be treated the same, independent of
their names (even if we are not quite there yet). what should happen to
non-GND components with "*" in their names when you copy them? i think
my patch affects (and fixes) this as well.

andresmmera:
I think  b32659e is needed in release-0.0.20 branch, the gnd bug is still there.

See https://drive.google.com/open?id=1cvcQoFr5XWwanE0upGfC83Jjgbt_bM--

Should we cherry-pick & open PR to release-0.0.20?

in3otd:
yep, we should have likely merged this in `release-0.0.20` first

Issue #860, opened 2018-09-04
install with cmake is broken


in3otd:
when running `make install` after running `cmake`
- it fails because it does not find the file `qucs-powercombining/qucspowercombining.1`
@andresmmera did you forget to upload the `man` page for `qucs-powercombining`? :grin:
Or should we remove that reference (the autotools build doesn't try to install that `man` page)
- after fixing the point above, `qucs` does not run because it does not find the `libqucsschematic.so` installed in `<install_prefix>/lib`

On Travis we do a build using `cmake` also but we don't do a `make install` there so these errors do not occur.

andresmmera:
> it fails because it does not find the file qucs-powercombining/qucspowercombining.1
@andresmmera did you forget to upload the man page for qucs-powercombining? grin
Or should we remove that reference (the autotools build doesn't try to install that man page)

Honestly, I didn't write any man page... but I can do it soon. Gimme a few days :-)

andresmmera:
Ok, I was taking a look to the man page files of the other tools and I think this text is outdated:

> The latest version of Qucs can always be obtained from
www.sourceforge.net\ or www.freshmeat.net

I've browsed www.freshmeat.net and it seems there's nothing related to Qucs there... Shall we remove that and mention that the source code is available on Github?

felix-salfelder:
On Wed, Sep 05, 2018 at 11:05:01AM -0700, Andrés Martínez-Mera wrote:
> Ok, I was taking a look to the man page files of the other tools and I think this text is outdated:
> 
> > The latest version of Qucs can always be obtained from
> www.sourceforge.net\ or www.freshmeat.net
> 
> I've browsed www.freshmeat.net and it seems there's nothing related to
> Qucs there... Shall we remove that and mention that the source code is
> available on Github?

i'd prefer no link over an outdated one. the more links we have, the
less likely they will be maintained.

(how about @QUCS_URL@ everywhere and a top level URL file,
similar to VERSION, see [1]. but i am a bit biased.)

[1] https://en.wikipedia.org/wiki/Don%27t_repeat_yourself

andresmmera:
> i'd prefer no link over an outdated one. the more links we have, the less likely they will be maintained.

I see your point, but if we include a section called 'Availability' in the manpage, we should provide some links to the readers.

> (how about @QUCS_URL@ everywhere and a top level URL file, similar to VERSION, see [1]. but i am a bit biased.)

[...]

Issue #780, opened 2018-02-15
optional things


felix-salfelder:
there are issues with optional things that are entangled with the main project. one is the documentation, the other is the cmake build system. there may be more in the future.

i propose to adopt the "anything that *can* be optional, *must* be optional" paradigm. this will improve maintainabilty, and enable progress. it will also make contributions possible at a later stage.

c.f. #773, #757, #742

guitorri:
Cmake build is not on par with Autotools (yet), but it is certainly not optional. 
It significantly lowers the barrier for new contributors that are familiar with IDE environments such as QtCreator. 

For the rest, documentation, adms, ..., I am ok with making them optional. As long as we keep one integration branch/repository to make sure things still work together.

felix-salfelder:
> It significantly lowers the barrier for new contributors

(maybe) true. currently it mainly makes refactoring impossible. refactoring requires work on the plugin system, which somebody needs to implement in a way that works on all platforms, plus incremental changes to make sure it does not break. somebody does not know how to deal with cmake.

in the far end, contributions will be possible as plugins, plugins can bring their own build systems (even scons, if you'd like).

are we in a deadlock?

i have no problem with cmake, i just can't maintain it. i am thinking of a model similar to gnucap. gnucap provides an autotools build system, which does *not* interfere with upstream development, while anybody can still use it, at will. (what's wrong with that?)

felix-salfelder:
cmake will be dealt with in #1007. other stuff will become optional during the refactoring, either way,

PR #862, opened 2018-09-08
Add manpage for qucspowercombining

andresmmera:
As mentioned in #860, the recently added qucspowercombining tool was
missing the man page file. This caused the cmake installation to crash.

Additionally, the content of the "Availabilty" and "Reporting bugs"
sections was put in the following file: /qucs/man/qucs_availability_bugs.1

### DIFF
--- a/qucs-core/CMakeLists.txt
+++ b/qucs-core/CMakeLists.txt
@@ -97,6 +97,12 @@ SET(PROJECT_COPYRIGHT_YEAR "2013")
 SET(PROJECT_DOMAIN_FIRST "qucs")
 SET(PROJECT_DOMAIN_SECOND "org")
 
+SET(QUCS_URL "https://sf.net/p/qucs")
+SET(QUCS_BUGREPORT "qucs-bugs@lists.sourceforge.net")
+
+configure_file(./doc/qucsconv.1.cmake.in ./doc/qucsconv.1)
+configure_file(./doc/qucsator.1.cmake.in ./doc/qucsator.1)
+
 # use last git commit hash along the version
 set(GIT unknown)
 IF(EXISTS ${CMAKE_SOURCE_DIR}/../.git )
--- a/qucs-core/CMakeLists.txt
+++ b/qucs-core/CMakeLists.txt
@@ -100,8 +100,8 @@ SET(PROJECT_DOMAIN_SECOND "org")
 SET(QUCS_URL "https://sf.net/p/qucs")
 SET(QUCS_BUGREPORT "qucs-bugs@lists.sourceforge.net")
 
-configure_file(./doc/qucsconv.1.cmake.in ./doc/qucsconv.1)
-configure_file(./doc/qucsator.1.cmake.in ./doc/qucsator.1)
+configure_file(${CMAKE_CURRENT_SOURCE_DIR}/doc/qucsconv.1.cmake.in ${CMAKE_CURRENT_SOURCE_DIR}/doc/qucsconv.1)
+configure_file(${CMAKE_CURRENT_SOURCE_DIR}/doc/qucsator.1.cmake.in ${CMAKE_CURRENT_SOURCE_DIR}/doc/qucsator.1)
 
 # use last git commit hash along the version
 set(GIT unknown)
--- a/qucs-core/configure.ac
+++ b/qucs-core/configure.ac
@@ -6,7 +6,7 @@ AC_PREREQ(2.64)
 dnl Read version from file
 m4_define([QUCS_VERSION], m4_esyscmd([tr -d '\n' < VERSION]))
 
-AC_INIT([qucs-core],[QUCS_VERSION],[qucs-bugs@lists.sourceforge.net])
+AC_INIT([qucsator], [QUCS_VERSION], [qucs-bugs@lists.sourceforge.net], [], [https://sf.net/p/qucs])
 AC_CONFIG_HEADERS([config.h])
 AC_CONFIG_AUX_DIR([build-aux])
 LT_PREREQ([2.2.2])
@@ -899,6 +899,8 @@ AH_TOP([
 #endif
 ])
 
+
+
 dnl Add here all your Makefiles. These are created by configure.
 AC_CONFIG_FILES([Makefile
 		 qucs_typedefs.h
@@ -913,6 +915,8 @@ AC_CONFIG_FILES([Makefile
 		 src/components/digital/Makefile
 		 src/components/verilog/Makefile
 		 tests/Makefile
+                 doc/qucsator.1
+                 doc/qucsconv.1
 		 ])
 
 
--- a/qucs-core/doc/qucsator.1.cmake.in
+++ b/qucs-core/doc/qucsator.1.cmake.in
@@ -1,0 +1,72 @@
+.TH Qucsator "1" "September 2004" "Debian/GNU Linux" "User Commands"
+.SH NAME
+Qucsator \- An integrated electronic circuit simulator.
+.SH SYNOPSIS
+.B qucsator
+[\fIOPTION\fR] ...
+.SH DESCRIPTION
+
+\fBQucs\fR is an integrated circuit simulator which means you are able
+to setup a circuit with a graphical user interface (GUI) and simulate
+the large-signal, small-signal and noise behaviour of the circuit.
+After that simulation has finished you can view the simulation results
+on a presentation page or window.
+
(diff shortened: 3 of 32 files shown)


felix-salfelder:
hmm. maybe only the variables? note that
PACKAGE_BUGREPORT and PACKAGE_URL already exist (these are standard autotools variables).

also make sure to use the configure.ac scripts corresponding to the package. the toplevel configure script is just band-aid (and maybe convenience).

andresmmera:
> hmm. maybe only the variables? note that
PACKAGE_BUGREPORT and PACKAGE_URL already exist (these are standard autotools variables).
 
uh, I don't see your point. Could you please ellaborate a bit more? Finally, I'm using three variables: QUCS_DESCRIPTION, QUCS_AVAILABILITY and QUCS_REPORT_BUGS.

> also make sure to use the configure.ac scripts corresponding to the package. the toplevel configure script is just band-aid (and maybe convenience).

Ok, but that implies duplicating the variables mentioned above. If I understood correctly, you're suggesting that we should define them both in qucs-core/configure.ac and  qucs/configure.ac and leave the top-level configure.ac untouched

On the other hand, the fact that the man pages are autogenerated using ./configure means that this approach won't work for the cmake build, right? I'll investigate if it exists a similar mechanism in cmake...

felix-salfelder:
On Sun, Sep 09, 2018 at 09:29:38AM -0700, Andrés Martínez-Mera wrote:
> > hmm. maybe only the variables? note that
> PACKAGE_BUGREPORT and PACKAGE_URL already exist (these are standard autotools variables).
>  
> uh, I don't see your point. Could you please ellaborate a bit more?
> Finally, I'm using three variables: QUCS_DESCRIPTION,
> QUCS_AVAILABILITY and QUCS_REPORT_BUGS.

i suggest to put something like

The latest version of Qucs can always be obtained from
\fB@PACKAGE_URL@\fR.

into the man page. yes it still duplicates some fragments, but not the
ones that require maintenance.

> > also make sure to use the configure.ac scripts corresponding to the
> > package. the toplevel configure script is just band-aid (and maybe
> > convenience).
> 
> Ok, but that implies duplicating the variables mentioned above. If I
> understood correctly, you're suggesting that we should define them
> both in qucs-core/configure.ac and  qucs/configure.ac and leave the
> top-level configure.ac untouched

yes. that's the price of having individual packages for individual
tasks. but the PACKAGE_* variables are already defined in every
*/configure.ac, so the overhead is quite small here.

> On the other hand, the fact that the man pages are autogenerated using
> ./configure means that this approach won't work for the cmake build,
> right?

yes. cmake is a can of worms on its own. the deal with merging
incomplete stuff: maintain it, or drop it, but i am biased. see #780.

thanks

andresmmera:
I proceeded as you suggested. 
Now I'll see what can we do with the CMake build.

felix-salfelder:
thanks

usually, AC_INIT is used to set these variables. *_BUGREPORT is already set (note that this should be an email address). maybe one URL is sufficient, i'd prefer https://sf.net/p/qucs over just sourceforge.net, YMMV

see https://www.gnu.org/software/autoconf/manual/autoconf-2.67/html_node/Initializing-configure.html

andresmmera:
Please, excuse my ignorance about the autoconf stuff:

I put:
`AC_INIT([qucs],  [PACKAGE_URL], [https://sf.net/p/qucs])`
in configure.ac. However, when running ./bootstrap and ./configure, the latter gets stuck.

I'm trying to see how PACKAGE_NAME, PACKAGE_TARNAME, PACKAGE_BUGREPORT, etc. are set in the generated configure file..., but no success so far.

Specifically, I see that this line:
`AC_INIT([qucs],  [QUCS_VERSION], [qucs-bugs@lists.sourceforge.net])`
(configure.ac, line 9) sets PACKAGE_BUGREPORT='qucs-bugs@lists.sourceforge.net' in configure. But, honestly, I don't know how... Am I missing something trivial?

andresmmera:
@felix-salfelder Thanks a lot!!

felix-salfelder:
looks really good now. but when you rename the qucsator package in `qucs-core/configure.ac`, please rename it to "qucsator", that seems to make most sense to me.

i do not know the intent of calling it "qucs-core" in the first place. i can't see a core there, and i believe it happened in error, YMMV.

guitorri:
I will merge #863. When ready, please change the base branch of this PR and rebase on `release-0.0.20`.

guitorri:
Please revert 2110066 to get the manpages installed also with cmake.

andresmmera:
I have got 6 commits from felix and in3otd after the rebase against `release-0.0.20`, is this ok?

image

andresmmera:
Thanks, Felix, I'd forgotten such valuable command!!

Now, the commit history looks clean

image

in3otd:
not directly related but I see we keep forgetting to update the copyright notes - I'll open a separate issue for this.
Not that I consider this the utmost important thing but stating that [qucsactivefilter](https://github.com/Qucs/qucs/blob/develop/qucs/qucs-activefilter/qucsactivefilter.1#L36) was done 13 years ago by someone who never worked on it it simply plain wrong.

andresmmera:
I've pushed a fix for cmake...

guitorri:
Hum...
```
CMake Error at qucs/man/cmake_install.cmake:36 (file):
  file INSTALL cannot find "/home/travis/build/Qucs/qucs/qucs/man/qucs.1".
```

andresmmera:
Yes, I saw that... I can't take a look right now, maybe tomorrow or so...

andresmmera:
The cmake build passed :tada: :tada:

On the other hand, the clang build is failing. (I would swear it was green before...) 
After reviewing the build log, I don't see why it fails.

https://api.travis-ci.org/v3/job/430674247/log.txt

At the end, it says:

> The job exceeded the maximum log length, and has been terminated.

So I guess it's failing after that.

in3otd:
yep, I [also](https://github.com/Qucs/qucs/pull/865#issuecomment-422269246) saw this, don't know what changed in Travis.

guitorri:
Perhaps Travis bumped the clang version?
Try to disable the warning with `-Wno-error=unused-command-line-argument`
https://stackoverflow.com/questions/47094843/clang-warning-argument-unused-during-compilation-stdlib-libc

in3otd:
um, seems not
- this PR [clang version](https://travis-ci.org/Qucs/qucs/jobs/430674247#L1054)
- previous passing build [clang version](https://travis-ci.org/Qucs/qucs/jobs/428222754#L1044)

still `clang version 5.0.0 (tags/RELEASE_500/final)`

Maybe they reduced the max log length; those warnings are also in the previous passing build.

andresmmera:
I've made a diff of both logs. They are very similar indeed. The bad one stops at:

> mv -f .deps/comp_1bit.Tpo .deps/comp_1bit.Plo

whereas the good one continues to work...

bad_vs_good

andresmmera:
I noticed that the bad one is a bit longer because, for example:

When it says:

> Cloning into 'Qucs/qucs'

 it shows the status of the download, which adds 398 extra lines.
extra lines


The same happens when it says:

> git fetch origin +refs/pull/862/merge:

which ads 326 extra lines...
extra lines2

My point is that the good one finished the qucs-core build by line 49501 (and starts with qucs-doc) whereas the bad one didn't finish that by the time it is forced to terminate (line 52595)

Can't we extend the maximum log length?

in3otd:
ops, modified the wrong files... 40ea1ee84939733804d2110305c1970d147f005a will need to be removed later

in3otd:
works now, there are still a bunch of warnings when compiling `adms` and `qucs-core`, some of which look pretty scary, like `warning: array index 2 is past the end of the array`, some coming also from autogenerated code but I have no idea if they are meaningful or not.

I've also noticed that the [test in configure.ac](https://github.com/Qucs/qucs/blob/d52358ca022ee2ae34b0f6886715b9b50b4eda54/qucs/configure.ac#L216) for the Qucs GUI does not work, as the same `CFLAGS` are applied for both [`gcc`](https://travis-ci.org/Qucs/qucs/jobs/431693826#L2753) and [`clang`](https://travis-ci.org/Qucs/qucs/jobs/431693825#L2086).
Maybe all this here should be put into a separate PR.

guitorri:
Please remove the last few Travis commits and rebase once more.
Hope it will be fine.

andresmmera:
Ok. Removed Travis-related commits and rebased.

image

PR #865, opened 2018-09-15
Refix optimization

in3otd:
Patched ASCO to allow passing the full path of simulator required by Qucs.

Removed the stuff about the ASCO "general" mode since it does not contain a post-simulation step to be executed, which we need to update the Qucs result file.
Adding a pre- and post-execution steps to ASCO would have been more general but I just wanted to fix the usage with Qucs here.

Works on Linux here, I'll test on Windows later.


in3otd:
> Works on Linux here, I'll test on Windows later.

I've done more testing on both Linux and Windows and it seems to work as expected with every circuit I tried.

felix-salfelder:
rebase?

felix-salfelder:
On Mon, Dec 10, 2018 at 12:26:53PM -0800, in3otd wrote:
> ok, likely tomorrow.
> Being a bugfix, should this be also in `release` (or it's already there?)

no rush. i was just thinking the CI fail was due to earlier generic
build issues, and not due to problems with your patch.

release.. i think everything made it into develop. i wouldn't object to
call it 0.0.20 any time now.

thanks

felix-salfelder:
@in3otd please rebase on top of #888, or i will do it (with your permission).

.. this is the last one holding back 0.0.20-rc1 (as it seems).

in3otd:
um, Travis is still not happy. I can take a closer look later this evening or feel free to work on this in the meantime if you wish.

in3otd:
Travis `cmake` build had strange errors, restarted two times and now it's passing.

felix-salfelder:
On Thu, Jan 03, 2019 at 10:06:53AM -0800, in3otd wrote:
> Travis `cmake` build had strange errors, restarted two times and now it's passing.

thanks.

will restart develop accordingly, as needed.

felix-salfelder:
travis/osx was happy with on 3e15e312c1a90a. now it fails on develop (0b1453bedac195a) after the merge... "No output has been received in the last 10m0s".

any clue? either way, i think it is about time for a qucs-0.0.20-rc1 tag...

in3otd:
very strange, it should just terminate at that point instead of hanging there. I tried to restart it but still behaves the same. I've sent a message to the Travis support about this.

PR #868, opened 2018-09-22
Internationalize "description" string and show tooltip only is showFrame is set

andresmmera:
According to https://github.com/Qucs/qucs/pull/764#pullrequestreview-155714159, it is needed to have the `description` string internationalized.
Otherwise Qucs will show the tooltip for all languages except English even when it is not expected to be shown.

In addition to this, now the tooltip is only displayed if the `showFrame` is set.


andresmmera:
This was already merged into release-0.0.20 too, so the target branch must be release-0.0.20.
Changing the base branch...

PR #871, opened 2018-09-23
Fix the previous Project navigation modification merge

in3otd:
the call to slotUpdateTreeview() is actually not needed and
this function was removed in commit adcacee97bcebbeba38203555a75bb549a75f9e1.


in3otd:
need #870 in `develop` also, otherwise the Travis `clang` build will always fail

in3otd:
included in #877

PR #873, opened 2018-09-26
Merge release fixes into develop

guitorri:
Apply `release-` fixes to `develop`.

PR #877, opened 2018-10-13
Cherry pick 870 to develop

andresmmera:
This PR is a cherry pick of ecae7fa and d30dc6d targeted to develop.

As @in3otd said [here](https://github.com/Qucs/qucs/pull/873#issuecomment-425209290), these commits need to be merge into develop. Then, we need to rebase and merge #871. Otherwise develop and release-0.0.20 will remain broken.

### DIFF
--- a/qucs/qucs/qucs.cpp
+++ b/qucs/qucs/qucs.cpp
@@ -3209,3 +3209,4 @@ bool QucsSortFilterProxyModel::lessThan(const QModelIndex &left, const QModelInd
 
   return QSortFilterProxyModel::lessThan(left, right);
 }
+// vim:ts=8:sw=2:et
--- a/qucs/qucs/qucs.cpp
+++ b/qucs/qucs/qucs.cpp
@@ -1226,8 +1226,7 @@ void QucsApp::slotListProjOpen(const QModelIndex &idx)
     // change projects directory to the selected one
     QucsSettings.projsDir = QucsSettings.projsDir.filePath(dName);
     readProjects();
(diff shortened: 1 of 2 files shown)


qucs/qucs/qucs.cpp
```diff
@@ -1226,7 +1226,8 @@ void QucsApp::slotListProjOpen(const QModelIndex &idx)
     // change projects directory to the selected one
     QucsSettings.projsDir = QucsSettings.projsDir.filePath(dName);
     readProjects();
-    slotUpdateTreeview();
+    // BUG?: it's no longer there.
```
in3otd: could as well have cherry-pick #871 here: `slotUpdateTreeview()` is no longer needed since the Project View now updates by itself whenever there's a change in the directory being watched. That function was removed in #770 but later #767 was merged and brought this call to `slotUpdateTreeview()` back; this because #767 should have been rebased before merging or merged before #770.
Note that I also commented out the call to `repaint()` since I did not see why it's needed and everything appears to work fine without.

andresmmera: You are right.
I didn't do that because I thought we will merge 9afdd76 and 3b56529  (even if the build failed) and then merge #871

andresmmera:
I had to remove the last commit from @felix-salfelder to cherry-pick 933415e6a171bacd1bf8de79b4491772a9f33d35 here, otherwise, an additional commit for a merge was needed... and basically, both commits do the same thing

Issue #892, opened 2019-01-09
Coupled lines filters size units


gvee-s:
Hello,

Using the filter synthesis tool I noticed that the size units are incomplete:

![screenshot_2019-01-09_21-05-54](https://user-images.githubusercontent.com/8284406/50925438-9d461380-1452-11e9-8aec-e5fdce7f1e4c.png)
![screenshot_2019-01-09_21-06-43](https://user-images.githubusercontent.com/8284406/50925439-9ddeaa00-1452-11e9-9811-9fdd03b95f71.png)
![screenshot_2019-01-09_21-07-31](https://user-images.githubusercontent.com/8284406/50925440-9ddeaa00-1452-11e9-8b0f-001009158ab8.png)

I understand that the units are referred to meters, so **u** should be **um** and **m** should be **mm**.

From what I see here
https://electronics.stackexchange.com/questions/197455/coupled-line-filter-design-in-qucs 
in version 0.0.18 units are printed correctly.

Regards,
g.

andresmmera:
It's highly likely that's my fault. I'm writing a fix for that.
@guitorri , @felix-salfelder : Should I target the PR to "release-0.0.20-2"?

guitorri:
Yes, target `release-0.0.20-2`.
I was about to tag and release `-rc1`. I wait for your fix.

in3otd:
um, we accidentally opened a can of worms here...
As well know to the many Qucs users that were already bitten by this, when a length is specified as "1 m" in Qucs it's interpreted as *one **milli**meter*, **not** *one meter* (it's also [documented](http://qucs.sourceforge.net/docs/tutorial/getstarted.pdf), page 16).
So when lengths happen to be expressed in meters no unit should be added.
We already have a [function](https://github.com/Qucs/qucs/blob/0ef64c307d72fa9c3d8075d6a69b79146f060e40/qucs/qucs/misc.cpp#L203) for this in the Qucs GUI, maybe you can link to that.
So you are now back to the initial solution - sorry.

And BTW, spaces between number and prefix/units should be used for all the units, not only *ohm*.

andresmmera:
And BTW, spaces between number and prefix/units should be used for all the units, not only ohm

Yes, that was done by adding the space at num2str()

felix-salfelder:
On Fri, Jan 11, 2019 at 12:35:55AM -0800, in3otd wrote:
> um, we accidentally opened a can of worms here...

[...]

PR #895, opened 2019-01-10
Fix issue #892

andresmmera:
See https://github.com/Qucs/qucs/issues/892

### DIFF
--- a/qucs/qucs-filter/filter.cpp
+++ b/qucs/qucs-filter/filter.cpp
@@ -89,10 +89,12 @@ double Filter::getQuadraticNormValues(int No, tFilter *theFilter, double &b)
 }
 
 // -----------------------------------------------------------------------
-QString Filter::num2str(double Num)
+// COPIED FROM QUCS-GUI, misc.h
+QString Filter::num2str(double Num, int Precision, QString unit)
 {
-  char c = 0;
-  double cal = std::abs(Num);
+  QString c;
+  QString Str;
+  double cal = fabs(Num);
   if(cal > 1e-20) {
     cal = log10(cal) / 3.0;
     if(cal < -0.2)  cal -= 0.98;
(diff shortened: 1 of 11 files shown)


guitorri:
Can you put changes in formatting a separate commit?
Right now it is hard to find and review the actual change/fix.

qucs/qucs-filter/filter.cpp
```diff
@@ -47,195 +44,262 @@ double Filter::getE6value(double value)
...
   }
 
   QMessageBox::critical(0, "Error", "Filter type not supported.");
   return 2e30;
 }
 
 // -----------------------------------------------------------------------
-QString Filter::num2str(double Num)
-{
+QString Filter::num2str(double Num, units Unit) {
```
in3otd: I guess this is the function actually changed. But in the commit comment it's referenced with a different name :grin:

andresmmera:
I added an extra argument to num2str() to handle unit. In my opinion, that seems more correct.
...It is simpler to keep num2str() untouched and simply add the meter units "m" in those lines where a TLIN is added...

I'm gonna revert these commits...

qucs/qucs-filter/filter.cpp
in3otd: Why did you choose to fix this by passing a *Unit* to `str2num`? Some filter types already have the correct units after the value (mostly the lumped elements ones, I think) while some other don't (the ones with transmission lines mostly, I think).
The type of element used is known in advance so there's no need to pass an additional parameter, just adding the correct  unit after the prefix, like already done for the lumped elements filters, should be enough.
And while you are at it, please re-add the space between the number and the prefix (see e.g. [here](https://www.bipm.org/utils/common/pdf/si_brochure_8_en.pdf), chapter 5.3.3. And, yes, it says also that *Ohm* should actually be written *ohm* but here we will continue with the Qucs tradition for the moment...)

in3otd:
> I'm gonna revert these commits...

just restart from the base branch and force-push, it's less work :grin:

andresmmera:
Yes, I was doing that :-)

in3otd:
(moving back the discussion to the PR, I commented on the issue by mistake)

>>And BTW, spaces between number and prefix/units should be used for all the units, not only ohm

> Yes, that was done by adding the space at num2str()

ah, right, I missed that and the fact that impedances do not go thru `num2str()`

PR #896, opened 2019-01-12
Fix PDF bounding box

### LINKED ISSUE #882: PDF bounding box wrong
Hi, 
I have (I do not know how) managed to create a `.sch` file which has the bounding box of the exported PDF wrong. Exporting in `png` and `eps` works ok, but the PDF is quite strange: 

image

I am using: `Qucs 0.0.20 (a7597048f)` (compiled here, Ubuntu 18.04). 

Files: /home/romano/lib/qucs/sallenkey_prj/BW3k-ok.sch
[test.zip](https://github.com/Qucs/qucs/files/2629198/test.zip)

in3otd:
Fixes #882. 

When creating a PDF file using *File->Export as image...* the resulting PDF size was often not correct.
Now the bouding box is specified in the intermediate `svg` file and not left to `inkscape` to guess/compute.

I've tested it on a few examples and seems to work fine but please double check.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -21,7 +21,6 @@ Qucs
 
 * [NEW] Wattmeter probe. [(PR#682)]
   -  Calculate apparent power, real power, reactive power and power factor.
-  -  Internal resistance can be set for the voltage and current ports.update news
 
 * [NEW] Add context menu to document tabs ([PR#708])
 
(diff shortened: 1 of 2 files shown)


andresmmera:
Hi,

I extended the test to the whole "Examples" corpus. I didn't notice any issue like #882 nor any other problem. :+1: 
Of course, I also tested Romano's schematic and other ones from mine.
[PR896-TEST.zip](https://github.com/Qucs/qucs/files/2752212/PR896-TEST.zip)

andresmmera:
I think it would convenient to tag this PR as "Milestone 0.0.20"

andresmmera:
> ah, did you use the qucs-test printing feature for this?

No, I did that manually, that is, using the File->Export Menu. I considered doing something like you mention, but I wanted to reproduce exactly what a user does to export a schematic as PDF.

andresmmera:
I've noticed that AppVeyor test didn't run. Is that ok?

in3otd:
uh, great, thanks - you have much more patience than me ! :grin:

andresmmera:
Well, I spent 15 minutes. For sure i would spend more time doing that programmatically...

in3otd:
> AppVeyor test didn't run

strange - there is not instruction to skip it in the commits...

in3otd:
I've restarted AppVeyor manually, apparently it did not run again after the force-push.

PR #905, opened 2019-01-29
Ignore swap files for vi/vim

Rmano:
Just found that `vim` swap files are not ignored...


Rmano:
I've put them just after the lock files for emacs, but yes, does not matter a lot...

Rmano:
I have narrowed down the ignore pattern; now it's just after the corresponding `emacs` write lock ignore...

PR #907, opened 2019-02-03
Adding a new tutorial

Rmano:
The linblocks tutorials teach how to use QUCS to simulate the frequency
response of linear blocks, and how to define them using the subcircuit
feature.


felix-salfelder:
there are two pdfs included, but the sources (in the preferred editable format) are missing. in one case that would perhaps be the .sch file, the other case, i don't know.
the screenshots are also binary, and do not carry any license. I suggest a separate text file just next to them, with an explicit license statement. it should list the pngs the license should apply to.

Rmano:
Ok, will fix. I imagine that GPL doc license would be OK, or do you prefer some kind of CC license for the documentation?
The source of one of the PDF would be the `.sch`, yes, I'll add it. The other diagram is made with `xcircuit`; the editable file is an encapsulated postscript that I'll add, too.
Thanks for the review.

felix-salfelder:
On Tue, Feb 05, 2019 at 10:06:12AM -0800, Romano Giannetti wrote:
> Ok, will fix. I imagine that GPL doc license would be OK, or do you
> prefer some kind of CC license for the documentation?

Currently, the project seems to use GPLv2+, also for qucs-doc. I have no
particular preference, but please do not omit the "+"...

> The source of one of the PDF would be the `.sch`, yes, I'll add it.
> The other diagram is made with `xcircuit`; the editable file is an
> encapsulated postscript that I'll add, too.

does includegraphics not support eps anymore? perhaps a make rule eps ->
pdf is preferrable over two copies.

thanks

qucs-doc/tutorial/linblocks/content.tex
```diff
@@ -0,0 +1,215 @@
...
+  \includegraphics[width=0.475\textwidth]{#2}\par\smallskip}
+
+% \begin{document}
+% \title{A gentle introduction to QUCS:\\ simulating a linear circuit}
+% \author{Romano Giannetti, \fname{romano.giannetti@gmail.com}}
+% \maketitle
+
+\tutsection{Presenting the Problem}
+
+This is a step by step tutorial on how to use QUCS to simulate multi-stage amplifier (with very simple and basic approach).
```
in3otd: should it be QUCS or Qucs?
In `qucs-doc` *QUCS* seems to be used about 30 times while *Qucs* about 800. We should likely use just one or the other.

Rmano:
I think I have solved the problems with copyright and source files. About the name style, I don't know ;-)... but I can change it if needed.

Rmano:
...and really I do not know why it's failing the build... no time now, will look at it tomorrow.

felix-salfelder:
On Tue, Feb 05, 2019 at 12:13:44PM -0800, Romano Giannetti wrote:
> ...and really I do not know why it's failing the build... no time now, will look at it tomorrow. 

eps.pdf: is in qucstex.mk. you need to include it.

*** No rule to make target `theproblem.pdf', needed by `distdir'.  Stop.

theproblem.pdf is still in EXTRA_DIST. should be changed to .eps.

there are two more warnings,

Makefile:726: warning: overriding commands for target `.tex.dvi'
Makefile:714: warning: ignoring old commands for target `.tex.dvi'

perhaps you should delete the rules that are already in qucstex.mk (are
they actually used or do we use pdflatex nowadays?).

QUCS vs Qucs: perhaps Qucs, everywhere. imo that's another patch.

Rmano:
Next try. I did a `make distclean; ./bootstrap; ./configure; make` on my machine and it works...

qucs-doc/tutorial/linblocks/content.tex
Rmano: Giving that there where more things to fix, I did that too. Thanks!

felix-salfelder:
On Wed, Feb 06, 2019 at 02:51:08AM -0800, Romano Giannetti wrote:
> Next try. I did a `make distclean; ./bootstrap; ./configure; make` on my machine and it works... 

thanks

FWIW: CI does "make distcheck". it tests the full
tarball->unpack->build->install->clean cycle. it should work
individually for all subdirectories (such as qucs-doc).

Rmano:
So now the error is
```
!!! Error: Cannot open pics/theproblem.eps: No such file or directory
```
...but it's there, both in my tree and in my patch... Puzzled.

Rmano:
There must be something pesky with the Makefile I made. First of all, if I touch `contents.tex` a rebuild is not triggered (why? I listed it in the `SOURCES` variable... next, a `make distcheck` fails with the above error, but if I, after that, do a `make` in the directory (after deleting the produced pdf) it succeeds. 

Any hint?

felix-salfelder:
On Wed, Feb 06, 2019 at 03:47:33AM -0800, Romano Giannetti wrote:
> There must be something pesky with the Makefile I made. First of all, if I touch `contents.tex` a rebuild is not triggered (why? I listed it in the `SOURCES` variable... next, a `make distcheck` fails with the above error, but if I, after that, do a `make` in the directory (after deleting the produced pdf) it succeeds. 

a bit easier maybe: get rid of the extra pics directory.

(you still have PDFS=somefile.eps. depending on how PDFS is used, this
might not work..)

felix-salfelder:
> Any hint? 

BUILT_SOURCES = theproblem.pdf

might help, see
https://www.gnu.org/software/automake/manual/html_node/Built-Sources-Example.html

Rmano:
```
! LaTeX Error: File `theproblem' not found.

 ...                                              
                                                  
l.54 ...egraphics[width=0.7\textwidth]{theproblem}
```

and this is really puzzling.  The graphics library does not require an explicit extension since, well, forever... 
:-(

Next try

felix-salfelder:
On Wed, Feb 06, 2019 at 11:45:24AM -0800, Romano Giannetti wrote:
> and this is really puzzling.  The graphics library does not require an explicit extension since, well, forever... 
> :-(
> 
> Next try

need both, distribute the .eps and build the .pdf from it. the former
you do with EXTRA_DIST, the latter you could do with BUILT_SOURCES.

(i am not sure if BUILT_SOURCES is automatically appended to
CLEANFILES, but you need to clean up what you build.)

Rmano:
Ah! BUILT_SOURCES did the trick. Thanks for help (I am not 100% sure I got it, but well...)

felix-salfelder:
On Wed, Feb 06, 2019 at 10:30:01PM +0000, Romano Giannetti wrote:
> Ah! BUILT_SOURCES did the trick. Thanks for help (I am not 100% sure I got it, but well...)

good. it's green now. two more things.

please check if the name(s) in the headers (Copyright/Authors) are
correct.

I found mentions of GFDL. GPL and FFDL are incopatible [1]. I don't know
what is going on, but GPL'd pics might not be the solution. how is this
done in the other subdirectories?

thanks!

[1] https://en.wikipedia.org/wiki/License_compatibility#GFDL_and_GPL

Rmano:
Ok, I'll check it --- probably after the weekend. I am a bit busy now... but this is not urgent. 
I left the copyright notices I found in the `../dcstatic/` tutorial, at least I think... but I do not mind putting everything under GPLv2+.

Rmano:
Ok. I removed the file with the GFDL and used just my file, released with GPLv2+. I added the copyright notice to the .eps and checked all the other files; please tell me if I have to change something more... and 1000 thanks for the review.

in3otd:
> I found mentions of GFDL. GPL and FFDL are incopatible [1]. I don't know what is going on, but GPL'd pics might not be the solution. how is this done in the other subdirectories? thanks!
 [1] https://en.wikipedia.org/wiki/License_compatibility#GFDL_and_GPL

ah, the usual mess about licenses...
The problem seems mainly that you cannot copy any text from a GFDL-licensed text into a GPL-licensed SW - unless if you are the author of that text, of course.

As far as I can see practically all the docs in Qucs are currently licensed under the GFDL with no Invariant Sections which, although not preferred, is ok also for [Debian](https://wiki.debian.org/DFSGLicenses#GNU_Free_Documentation_License_.28GFDL.29), AFAIU.

guitorri:
Tested locally. Trying to "Rebase and merge" directly on github.

PR #923, opened 2019-05-16
Port qucsfilter to qt5

Kezii:
I'm really failing to understand what is going on with qucs development, I see a ton of pull requests with huge diffs, and years-old discussions about this

So take this PR like an enhanced bug report, i'm not sure if I'm allowed to do this, but if the little programs can be updated to qt5, even without the main application, that'd be great.

It looks and works so nicer, at least on Linux (when a text field is inactive, it shows as that, the qt4 version was almost unusable)

immagine


in3otd:
> I'm really failing to understand what is going on with qucs development

you're not alone...

> it looks and works so nicer, at least on Linux

good but it's strange you had issues with the inactive text inputs with Qt4, they always worked as expected here (i.e. grayed out when disabled)

BTW, your PR build has some errors
https://travis-ci.org/Qucs/qucs/jobs/533564694#L1878

and besides cmake, also the autotools build should be updated...

.travis.yml
```diff
@@ -16,7 +16,7 @@
 
 language: cpp
 dist: trusty
-sudo: false
+sudo: required
```
in3otd: I missed that now Travis has deprecated the  container-based Trusty infrastructure, which required `sudo: false`.
Anyway, you should be able to install the needed Qt5 stuff as Felix did [here](https://github.com/Qucs/qucs/pull/898/files#diff-354f30a63fb0907d4ad57269548329e3R51) in his PR.

Kezii: That seems like a better idea, thanks

guitorri:
This is already done in the `qt5-6-merge-rc1` and `develop+qt5-10` branches.
In both branches can build with either Qt4 or Qt5 be selected.
The issue being that Qucs-Schematic is not yet fully functional.

These are two different stretches to fix different things on the schematic code

Releasing a mix of Qt4 and Qt5 binaries is not an option. 

Yes, the diffs are big because a few things need to be changed.

There are mainly 2 people working on these branches. 
Since this is really boring work, things tend to stale for long periods of time.

All that said your PR cannot be merged in develop at this time.
You are welcome to work on the other branches. Please get in touch to discuss directions before starting.

Kezii:
> This is already done in the qt5-6-merge-rc1 and develop+qt5-10 branches.

Yeah that's probably true, I should have asked before making the pr but my main objective was fixing my own bug and sparking some public/github discussion about the qt5 port, since the last one was mid 2018

 I wasn't expecting to have this merged

> Releasing a mix of Qt4 and Qt5 binaries is not an option.

May I ask why?

> Since this is really boring work, things tend to stale for long periods of time.

Yeah, for that reason I was asking if it was possible to have at least the small binaries ported, that's still a big chunk out of the way, even if it's not coherent, it's better broken




Thanks @guitorri  and @in3otd for the replies

in3otd:
yep, as mentioned before I'm also confused about the current Qucs development status; I should have known/remembered that the tools were already ported but didn't.

To help people know the current status of the Qt5 branches a few lines in the respective `README.md` files telling what has been done, what's not, where help is needed and where a newcomer could start helping would be a good thing.

PR #930, opened 2019-05-28
Fix Graph loading when Dataset names contain spaces

### LINKED ISSUE #929: Diagram can not load from dataset having white spaces
see [discussion](https://sourceforge.net/p/qucs/mailman/message/36678681/) on `qucs-help`;
If a Diagram has a Graph which takes the data from a Dataset with spaces in the name, the Graph loading fails.
Issue is due to [this line](https://github.com/Qucs/qucs/blob/develop/qucs/qucs/diagrams/graph.cpp#L144) and following lines: spaces are considered fields delimiters, so the string after the first space is assumed to be the Graph color but since it's not a valid color encoding loading fails.

in3otd:
Fix #929. Works for me. Please check.
(should this go to `master` as a hotfix too?)

If a Diagram contained a Graph which referenced a variable in a Dataset with a name containing spaces, the Graph loading failed as the simple parsing code was confused by the spaces in the Dataset name.

### DIFF
--- a/qucs/qucs/diagrams/graph.cpp
+++ b/qucs/qucs/diagrams/graph.cpp
@@ -139,9 +139,13 @@ bool Graph::load(const QString& _s)
   s = s.mid(1, s.length()-2);   // cut off start and end character
 
   Var = s.section('"',1,1);  // Var
+  // Var can include a Dataset name, which can contain spaces
+  // remove the Var string so subsequent parsing of the other fields does not fail in this case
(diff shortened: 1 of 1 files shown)


guitorri:
Please branch from `master` a `hotfix-#` and PR against both, `master` and `develop`.

in3otd:
ok, I've done two new PRs (#931, #932)

PR #932, opened 2019-05-29
Fix Graph loading when Dataset names contain spaces (-> develop)

### LINKED ISSUE #929: Diagram can not load from dataset having white spaces

in3otd:
Fix #929. Works for me. Please check.

If a Diagram contained a Graph which referenced a variable in a Dataset with a name containing spaces, the Graph loading failed as the simple parsing code was confused by the spaces in the Dataset name.

(supersedes #930)

### DIFF
--- a/qucs/qucs/diagrams/graph.cpp
+++ b/qucs/qucs/diagrams/graph.cpp
@@ -139,6 +139,9 @@ bool Graph::load(const QString& _s)
   s = s.mid(1, s.length()-2);   // cut off start and end character
 
   Var = s.section('"',1,1);  // Var
+  // Var can include a Dataset name, which can contain spaces
+  // remove the Var string so subsequent parsing of the other fields does not fail in this case
+  s = s.section('"', 2); // keep everything after the closing quotes
 
   QString n;
   n  = s.section(' ',1,1);    // Color


in3otd:
I have rebased this on the latest `develop` since I think it should also go to `0.0.20`.

By the way, `ccache` on OSX seems to work now, even if the hit rate is rather low
```
$ ccache --show-stats
[snip]
stats updated                       Sun Feb 16 09:34:54 2020
stats zeroed                        Sun Feb 16 09:22:31 2020
cache hit (direct)                   370
cache hit (preprocessed)              18
cache miss                           584
cache hit rate                     39.92 %
[snip]
files in cache                      3017
cache size                         630.0 MB
max cache size                       5.0 GB
```

they are not better for Linux `gcc`+`autotools` build, actually
```
$ ccache --show-stats
cache directory                     /home/travis/.ccache
cache hit (direct)                   278
cache hit (preprocessed)             306
cache miss                           585
[snip]
files in cache                      2033
cache size                         468.4 Mbytes
max cache size                     512.0 Mbytes
```

while for the Linux `gcc`+`cmake` build things are much better
```
$ ccache --show-stats
cache directory                     /home/travis/.ccache
cache hit (direct)                   784
cache hit (preprocessed)               7
cache miss                             1
[snip]
cache size                         236.8 Mbytes
max cache size                     512.0 Mbytes
```

similar for the Linux `clang`+`autotools` build
```
$ ccache --show-stats
cache directory                     /home/travis/.ccache
cache hit (direct)                  1150
cache hit (preprocessed)              74
cache miss                             1
[snip]
files in cache                      4940
cache size                         150.0 Mbytes
max cache size                     512.0 Mbytes
```

Not sure where the differences between builds come from.

PR #946, opened 2020-01-04
[Not yet finished] Fix wattmeter probe

### LINKED ISSUE #47: Power probe AC results are wrong
The power probe (`wprobe`) gives wrong results when used with AC simulations:

see this circuit simulation (circuit taken from [here](http://www.ibiblio.org/kuphaldt/electricCircuits/AC/AC_11.html)):
image
note that the overall power values are correct but that the resistor is apparently *generating* power and has a significant reactive power component...

The issue is due to a wrong basic calculation of the power in the `wprobe` component, as in practice it computes the complex power as <tt>S = V I</tt> and flips the sign of the reactive power instead of using <tt>S = V I<sup>*</sup></tt>. So when using the probe to compute the power provided by a source, so measuring the voltage across the source, the current calculation works, as the voltage from the source is purely real. But when measuring across a component, as shown above, the voltage has a phase different from zero and the results are wrong.

After correcting the formula used the results are fine:
image

I'll open a PR with the fix soon...

### LINKED ISSUE #945: Power probe AC results are wrong
The power probe (`wprobe`) gives wrong results when used with AC simulations:

see this circuit simulation (circuit taken from [here](http://www.ibiblio.org/kuphaldt/electricCircuits/AC/AC_11.html)):
image
note that the overall power values are correct but that the resistor is apparently *generating* power and has a significant reactive power component...

The issue is due to a wrong basic calculation of the power in the `wprobe` component, as in practice it computes the complex power as <tt>S = V I</tt> and flips the sign of the reactive power instead of using <tt>S = V I<sup>*</sup></tt>. So when using the probe to compute the power provided by a source, so measuring the voltage across the source, the current calculation works, as the voltage from the source is purely real. But when measuring across a component, as shown above, the voltage has a phase different from zero and the results are wrong.

After correcting the formula used the results are fine:
image

I'll open a PR with the fix soon...

in3otd:
Fix Qucs/qucsator#47 .

There where two main issues:
- the formula used for computing the complex power was wrong:
  this was easy, just use the correct formula for the power
- the voltage noise values from were wrong:
  this required changing the nodes order for the component in the netlist: the probes are a bit of a hack and currently require the voltage sensing to be on the first two nodes, while the wattmeter had the current sensing there.

I also couldn't resist changing a little the probe values output handling to make it feel a little less "hackish" (IMHO).

This goes together with the `qucs-test` update at Qucs/qucs-test#35 .

### DIFF
--- a/qucs-core/src/components/wprobe.cpp
+++ b/qucs-core/src/components/wprobe.cpp
@@ -71,8 +71,6 @@ void wprobe::saveOperatingPoints (void) {
   nr_double_t Q = imag (Sw);
   setOperatingPoint ("P", P);
   setOperatingPoint ("Q", Q);
-//Power Factor calculation
-  setOperatingPoint ("PF", P/std::sqrt(P*P+Q*Q));
 }
 
 void wprobe::initTR (void) {
--- a/qucs-core/src/components/wprobe.cpp
+++ b/qucs-core/src/components/wprobe.cpp
@@ -39,14 +39,16 @@ wprobe::wprobe () : circuit (4) {
   setVoltageSources (1);
 }
 
-//NODE 1 --> I+
-//NODE 2 --> I-
-//NODE 3 --> V+
-//NODE 4 --> V-
+// voltmeter nodes must be at nodes 1 and 2 as so it's assumed
+//   in saveNoiseResults() for noise calculations
+//NODE 1 --> V+
+//NODE 2 --> V-
+//NODE 3 --> I+
+//NODE 4 --> I-
 
 void wprobe::initDC (void) {
   allocMatrixMNA ();
-  voltageSource (VSRC_1, NODE_1, NODE_2);
+  voltageSource (VSRC_1, NODE_3, NODE_4);
 }
 
 void wprobe::initAC (void) {
@@ -54,8 +56,9 @@ void wprobe::initAC (void) {
 }
 
 void wprobe::saveOperatingPoints (void) {
-  nr_double_t Vr = real (getV (NODE_3) - getV (NODE_4));
-  nr_double_t Vi = imag (getV (NODE_3) - getV (NODE_4));
+  nr_double_t Vr = real (getV (NODE_1) - getV (NODE_2));
+  nr_double_t Vi = imag (getV (NODE_1) - getV (NODE_2));
+  // operating points as for the vprobe
   setOperatingPoint ("Vr", Vr);
   setOperatingPoint ("Vi", Vi); //This section works just like a voltmeter
 }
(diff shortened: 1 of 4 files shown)


in3otd:
ah, the joys of CI...
- on AppVeyor, `pacman` [doesn't find Qt4](https://ci.appveyor.com/project/qucs/qucs/builds/29895180/job/g6yskbfax4944c5u#L480)
- on Travis, there is [some issues with Homebrew](https://travis-ci.org/Qucs/qucs/jobs/632696723#L168) and it complains about a missing `ccache`

any idea is appreciated...

in3otd:
uh, there's still something wrong: as [described in the Qucs docs](http://qucs.sourceforge.net/docs/tutorial/getstarted.pdf), "*all voltages and currents are peak values*" (see pag 13) so the AC power results should be divided by 2. But for DC simulations not.

So the results shown in Qucs/qucsator#47 are wrong, since the 120 V of the source is the *peak* value and not the RMS as intended in the linked document calculations.

guitorri:
Concerning CI. MSYS2 seem to have dropped Qt4. One option is to install the official Qt4 package (I used to to it long time ago, it should be on the .appveyor history)... another is to finish the port to Qt5 :). On Travis/OSX you can try to either disable ccache or add it to the brew install list.

in3otd:
> MSYS2 seem to have dropped Qt4

uh, the world is coming to an end :grin:
Finishing the Qt5 port would of course be the best solution but, alas, I cannot help much there.

The OSX build log says 
```
Tapped 0 formulae (189 files, 264KB)
Error: Your Homebrew is outdated (need >=2.1.0, found 1.3.5 at /usr/local).
Please run `brew update`.
Error: Kernel.exit
```
so I guess it's something more than `ccache` but I know nothing about OSX

guitorri:
I will try to get Qt5 back on track.
In the mean time, can you rebase?

in3otd:
Thanks - Qt5 will obviously be the best solution.

As mentioned above, the fix here is not yet finished - I need to separate the DC and AC cases due to the fact that AC uses peak values. I'll rebase anyway in the meantime.

felix-salfelder:
This is listed as a blocker for 0.0.20. it changes the wattmeter symbol and looks like that will introduce a backwards incompatibility.

better defer to 0.0.21?

felix-salfelder:
I made the 0.0.20 "release" long time ago, with the intent to archive previous work. 0.0.20 does not include this work, and it might make sense to recheck.

The discussion above got stuck at Qt, which is now sorted in develop. It also seems that some of the changes are to qucsator, which is now in a separate repo ...

PR #954, opened 2020-02-02
Fix Qt4 removal from MSYS2 packages

in3otd:
...trying a "fix" from [here](https://github.com/slowphil/texmacs-win-builder/issues/3).

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -34,7 +34,16 @@ install:
   # Switch from SF to msys2.org (default, much faster)
   - bash -lc "pacman --noconfirm --sync pacman-mirrors"
   - bash -lc "pacman --noconfirm -S autoconf automake bison flex"
-  - bash -lc "pacman --noconfirm -S mingw-w64-$MARCH-qt4"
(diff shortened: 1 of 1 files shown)


in3otd:
Just noticed that the [Appveyor docs](https://www.appveyor.com/docs/build-cache/) say that "*By default, saving cache is disabled in Pull Request builds.*", so it seems we need to have this merged before having a valid cache.

in3otd:
ok, it seems to work.
Someone please press the *Merge* button - or I'll do that later :grin:.
Not sure this should go to `master` now, as there are other small things that should be merged in `develop`, so maybe we can merge `develop` into `master` later.

guitorri:
Only releases branches should merge back into master. 
I will give another try on finishing off release 20.

in3otd:
thanks.
yes, but I wonder(ed) if also #953 and #951 (and #946, assuming I'll be able to finish it soon) should be included in 0.0.20.

I was also taking a look at how to do the release, cross-compiling for Windows in Linux - but I'll need quite some time to set up everything and I'm not sure I'll be able to make it work.

guitorri:
Sure, tag them for 20. After they are in, we release it.

PR #960, opened 2020-02-11
Travis check01

oxmon-2500:
test01

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -133,6 +133,8 @@ install:
       export PATH="/usr/local/opt/bison/bin:$PATH";
       pip2 install matplotlib;
       export PATH="/usr/local/opt/ccache/libexec:$PATH";
+      ls -la /usr/local/opt/ccache/libexec
+      find /usr/local -name ccache -ls
     fi
 
 before_script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -138,10 +138,8 @@ install:
     fi
 
 before_script:
-  - brew ccache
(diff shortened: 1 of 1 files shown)


in3otd:
looking e.g. [here](https://github.com/scipy/scipy/blob/master/.travis.yml) I think the right command is `brew install ccache` and it should be run only for OSX.

Anyway, I saw other had issues with outated `homebrew` on Travis and it seems that the easier solution is to use a newer Xcode, see https://travis-ci.community/t/xcode-8-3-homebrew-outdated-error/3798/10 . I don't know if this is an option here. @guitorri ?

oxmon-2500:
> looking e.g. [here](https://github.com/scipy/scipy/blob/master/.travis.yml) I think the right command is `brew install ccache` and it should be run only for OSX.
> 
> Anyway, I saw other had issues with outated `homebrew` on Travis and it seems that the easier solution is to use a newer Xcode, see https://travis-ci.community/t/xcode-8-3-homebrew-outdated-error/3798/10 . I don't know if this is an option here. @guitorri ?

Hi, thank you for the suggestions. I am new to Github and Travis and I learn a lot. At the moment I do experiments with travis and I am optimistic in the next days to eradicate the problems with travis.

in3otd:
feel free to experiment, it's one of the best way to learn :grin: .

Note that you can rewrite your PR branch completely if you want to try a different solution, just reset it  or rebase, ignoring unwanted commits, and then `push --force`.

PR #962, opened 2020-02-12
Fix cross-compiling to Win32

in3otd:
I have done some modifications to be able to cross compile to Win32 using Wine again.
Extensive rearrangement/modifications of `configure.ac` section were needed since the related part was butchered when we started to use `pkg-config` and in subsequent modifications, hope everything still works.

### DIFF
--- a/contrib/windows/qucs-build-win32.sh
+++ b/contrib/windows/qucs-build-win32.sh
@@ -1,8 +1,8 @@
 #!/bin/bash
 
 #
-# This script cross compiles Qucs from OSX to Windows 32.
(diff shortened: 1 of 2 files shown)


guitorri:
This looks good. pkg-config logic was broken also with Ubuntu 18.04. With this fix the way of building with mingw32 might work again. Thanks!

in3otd:
yep, I see that all the builds using `pkg-config` still work, which I was not sure with all the modifications done. 
I wanted to add also a Wine-based build to Travis, but this will be for much later.

PR #964, opened 2020-02-13
Qucs test python3

guitorri:


guitorri:
It is still mixing python2 and python3 somewhere:
```
cd qucs-test && python run_equations.py --prefix /home/travis/build/Qucs/qucs/qucs-0.0.20/_build/qucs-core/src/
  File "run_equations.py", line 636
    print('   ', line, end=' ')
```

PR #966, opened 2020-02-14
travisCheck03

oxmon-2500:


.travis.yml
```diff
@@ -112,27 +112,27 @@ before_install:
...
 
   # Setup OSX
   # Installed already: autoconf automake libtool pkg-config
   - if [[ $OSX ]]; then
       export PATH="/usr/local/opt/flex/bin:$PATH";
       export PATH="/usr/local/opt/bison/bin:$PATH";
-      pip2 install matplotlib;
+      pip3 install matplotlib;
       export PATH="/usr/local/opt/ccache/libexec:$PATH";
+      brew install ccache
```
guitorri: Travis is not running, you need to add semicolon (;) to the end of this line

.travis.yml
```diff
@@ -112,27 +112,27 @@ before_install:
...
+      sudo find /usr /opt -type f -name ccache -ls #
```
guitorri: add ; here also

.travis.yml
oxmon-2500: Thx, I don't have any messages from travis in such a case

guitorri:
It seems that on OSX the library path is not being passed to the linker. 
Could be related to the recent pkg-config changes.
I cannot debug this atm.
The mac computer I have around is on a newer OSX version and qt@4 needs to be built from source as there are no binaries from cartr/qt4.

oxmon-2500:
I will analyse this over weekend.
Brew has an option --build-from-sources I will try this on travis

guitorri:
Travis is installing qt@4 just fine. For the macOS versions on Travis (High Sierra, Mojave) we can still use the binaries bottles available from cartr/qt4.

It is the linker that cannot find the libraries.

Try adding :
```
    dnl Append linker flags
    QT_LIBS="${QT_LIBS} ${QT_LDFLAGS}"
```

After this line: https://github.com/Qucs/qucs/blob/develop/qucs/configure.ac#L416

in3otd:
oops, I tried not to modify the OSX part but I have missed that, indeed the
`QT_LIBS="${QT_LIBS} ${QT_LDFLAGS}"` was there before - I removed it because it was no longer needed for Windows, sorry.

oxmon-2500:
Thank you, the next try started

oxmon-2500:
Hooray, we made it together!

oxmon-2500:
Thanks to all

guitorri:
Nice job.
Can you cleanup the commit history? Or do you need help this time?

guitorri:
I squashed and pushed your commits to #964.

oxmon-2500:
Thank you, did squashing mean rebase? I am still not acquainted with all steps in git.

guitorri:
Squash is to merge related commits into a single commit. One way to do that is during a interactive rebase.

in3otd:
@oxmon-2500 there are some (brief) explanations abot squashing/rebasing/etc. also in our "[contributing](https://github.com/Qucs/qucs/wiki/Contribution)" guide.

PR #969, opened 2020-02-15
Silence warnings

guitorri:


in3otd:
FYI, I saw the ` 'WDestructiveClose' is not a member of 'Qt'` error when fixing the Win32 build and I had the impression it was related to the libraries/macro order. Anyway, the right fix is to use the Qt4 equivalent, as you did :grin:.

I noticed that `ccache` seems not actually used for OSX, not sure why. Travis [docs](https://docs.travis-ci.com/user/caching/#ccache-cache) say to also modify the `PATH` to use it in OSX.

guitorri:
Yep, these are old and new patches staring in May 2019 when I saw the issues in Win32 builds...
Not sure about ccache. I never used it myself. Why do you say it is not used? The build time?
It doesn't hurt to follow the docs and add the path. Do you want to try this?
I will merge this one.

in3otd:
yes, the build time and the `ccache` [stats](https://travis-ci.org/Qucs/qucs/jobs/650918298#L4691).
I'll try to fix that later.

PR #970, opened 2020-02-16
Fix ccache path for OSX

in3otd:
as ccache wrappers for the compilers were not seen


in3otd:
ok, `ccache` [stats](https://travis-ci.org/Qucs/qucs/jobs/651038267#L4693) are now non-zero, so at least the wrappers are used as expected. I guess we need to wait for the next PR to see if the `ccache` cache is effectively restored by Travis.

guitorri:
Looks right.

PR #971, opened 2020-02-17
qucs#936 bug corrected, tested with python3

oxmon-2500:
python_01
python_03

### DIFF
--- a/qucs/qucs/python/parse_result.py
+++ b/qucs/qucs/python/parse_result.py
@@ -65,7 +65,8 @@ def parse_file(name):
     data['variables'] = variables
     # reverse the shape variable in order to get the reshape operation (see below)
     # correct
-    shape = shape[::-1]
+    shape = shape[::-1].astype('int')
+    shape = shape.squeeze()
 
     # here comes the clever trick :-)
     # if a dependent variable depends on N > 1 (independent) variables,
@@ -75,6 +76,6 @@ def parse_file(name):
         temp = data['variables'][key]
         if temp == 'dep':
             temp_data = data[key]
-            data[key] = np.reshape(temp_data, shape)
+            data[key] = np.reshape(temp_data, shape).squeeze()
 
     return data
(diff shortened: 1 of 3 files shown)


in3otd:
could you please check if this is related to [this post](https://sourceforge.net/p/qucs/mailman/message/36737979/);  - see fix at the [end of the thread](https://sourceforge.net/p/qucs/mailman/message/36738182/) . I forgot about opening an issue/PR about that...
IIRC, there the issue was that `shape` should be an `int` or a list of `int` but it was actually an `ndarray`.

oxmon-2500:
> could you please check if this is related to [this post](https://sourceforge.net/p/qucs/mailman/message/36737979/); - see fix at the [end of the thread](https://sourceforge.net/p/qucs/mailman/message/36738182/) . I forgot about opening an issue/PR about that...
> IIRC, there the issue was that `shape` should be an `int` or a list of `int` but it was actually an `ndarray`.

The fix at the end of the thread was already implemented. I inserted the fix proposed in qucs#936 and make python3 migration. I tested it only with the files from the same directory which was OK.

in3otd:
um, it seems to me that the fix in that thread is not in here, line [31](https://github.com/Qucs/qucs/blob/a04439250c1d9165365efcde8faa3a43e43d308b/qucs/qucs/python/parse_result.py#L31) was not changed.
The fix done here in the end is similar, as it makes sure that `shape` contains only `int` but I don't see why it has to be a numpy array instead of a list.

I see that after restarting Travis is passing now, no idea where the problem seen before came from.

oxmon-2500:
I thought the prefix np. is the change, it was to quick conclusion - tnx for accurate review. But I tried 

     shape.append(int(g[2]))

and it produces on other place:

    shape = shape[::-1].astype('int')
    AttributeError: 'list' object has no attribute 'astype'

In the current version it works fine with the python3 and produces a nice chart.
Probably it lies on the newly version of numpy .

Regarding Travis, its probably somewhere a non-multithread-able procedure and it may come again

PR #993, opened 2020-03-23
qucs#992 corrections in NMOSFETs PMOSFETs Transistors Z-Diodes libs

oxmon-2500:
Value out of range - deactivated by commenting it out
NMOSFETs    **IRF3709ZCS_L**
NMOSFETs    **IRF3709Z_S_L**
NMOSFETs    **IRF3805S-7P**
NMOSFETs    **IRF7821**
PMOSFETs    **IRF7220**
PMOSFETs    **IRF7233**
PMOSFETs    **IRF7410**
Transistors **MMST4401**

Typos - corrected
NMOSFETs **IRF7465**                 missing \<Model\> tag
NMOSFETs **IRFB4229PBF**       Component Name and .Def Name equalized
NMOSFETs **IRFR024N**      Component Name and .Def Name equalized
NMOSFETs **IRFR110**      Component Name and .Def Name equalized
NMOSFETs **IRFR1205**      Component Name and .Def Name equalized
NMOSFETs **IRFR120N**      Component Name and .Def Name equalized
NMOSFETs **IRFR220**      Component Name and .Def Name equalized
NMOSFETs **IRFR320**      Component Name and .Def Name equalized
NMOSFETs **IRFR3910**      Component Name and .Def Name equalized
NMOSFETs **IRFR4105**      Component Name and .Def Name equalized
Z-Diodes **1N5342B**              'Vj' occurred 2x
Z-Diodes **1N5344B**              'Vj' occurred 2x
Z-Diodes **1N5346B**              'Vj' occurred 2x
Z-Diodes **1N5377B**              .Def:Z-Diodes_1N5377B -> .Def:Z_Diodes_1N5377B

### DIFF
--- a/qucs/qucs-lib/library/NMOSFETs.lib
+++ b/qucs/qucs-lib/library/NMOSFETs.lib
@@ -4053,73 +4053,75 @@
   </Model>
 </Component>
 
-<Component IRF3709ZCS_L>
-  <Description>
-   Enhancement n-channel power MOSFET
-   30 V, 87 A, 6.3 mOhm, package: D2-Pak
-   Manufacturer: International Rectifier
-   added by Gunther Kraut <gn.kraut@online.de>
-  </Description>
-  <Model>
-  .Def:NMOSFETs_IRF3709ZCS_L _net2 _net1 _net3
-    MOSFET:M1 _net7 _net9 _net8 _net8 Type="nfet" L="100u" W="100u" Is="1e-32" Vt0="2.31318" Lambda="0" Kp="226.971" Cgso="1.95329e-05" Cgdo="1e-11" N="1" Gamma="0" Phi="0.6"
-    # .MODEL:MM  NMOS (LEVEL=1 IS=1E-32 VTO=2.31318 LAMBDA=0 KP=226.971 CGSO=1.95329E-05 CGDO=1E-11) 
-    R:RS _net8 _net3 R="0.00411316"
-    Diode:D1 _net1 _net3 Is="1.99276e-10" Rs="0.0029252" N="1.17266" Bv="30" Ibv="0.00025" Eg="1.2" Xti="3.10172" Tt="1e-07" Cj0="9.86355e-10" Vj="0.5" M="0.403132" Fc="0.5"
-    # .MODEL:MD  D (IS=1.99276E-10 RS=0.0029252 N=1.17266 BV=30 IBV=0.00025 EG=1.2 XTI=3.10172 TT=1E-07 CJO=9.86355E-10 VJ=0.5 M=0.403132 FC=0.5) 
-    R:RDS _net3 _net1 R="1e+06"
-    R:RD _net9 _net1 R="0.0001"
-    R:RG _net2 _net7 R="1.48797"
-    Diode:D2 _net5 _net4 Is="1e-32" N="50" Cj0="2.98433e-10" Vj="20.8385" M="0.560566" Fc="1e-08"
-    # .MODEL:MD1  D (IS=1E-32 N=50 CJO=2.98433E-10 VJ=20.8385 M=0.560566 FC=1E-08) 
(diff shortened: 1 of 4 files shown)


felix-salfelder:
On Mon, Mar 23, 2020 at 02:17:20AM -0700, Szymon Blachuta wrote:
> Value out of range - deactivated by commenting it out
> [..]

I can't see the issue with these subcircuit models.

If a value is out of range, there are three possibilities
- the value is not supported by the device and there is a warning.
- the range check is wrong, and it is actually within range.
- the value is a typo, should be a different one, possibly within range.

what will commenting out fix?

> Typos - corrected
> [..]

changes the name of a several components. on what basis?
what happens if somebody uses these components?

oxmon-2500:
>what will commenting out fix?

Using this components results in aborting the simulation:

checking netlist...
Errors occurred during simulation on Mon 23. Mar 2020 at 15:49:49:190
**Aborted.**
Errors and Warnings:
checker error, value of 'Vj' (11.8882) is out of range ']0,10]' in 'Diode:D2'

If the component is unusable then it is better not to offer it to build it into a circuit.
Maybe the range check is wrong, thus it remains unusable. The range should be other issue.
It seems to me it is a typo, because there are many more Vj values within the range.
The components are still in the library as comments. Hopefully someone will take a look at it, correct the value and uncomment it


> Typos - corrected [..]
> changes the name of a several components. on what basis? what happens if somebody uses these components?

In all component definitions: 
<Component **BSP125**>
...
  \<Model\>
  .Def:NMOSFETs_**BSP125** _net1 _net3 _net2

the component name and the name in .Def (after underscore) are the same. Only in those erroneous the second name slightly differs and looks like typo. I wanted to ask the author Gunther Kraut, but his eMail \<gn.kraut@online.de\> is no more up to date.

felix-salfelder:
On Mon, Mar 23, 2020 at 09:23:18AM -0700, Szymon Blachuta wrote:
> >what will commenting out fix?
> 
> Using this components results in aborting the simulation:
> Errors and Warnings:
> checker error, value of 'Vj' (11.8882) is out of range ']0,10]' in 'Diode:D2'

Is this not correct behaviour?

> If the component is unusable then it is better not to offer it to
> build it into a circuit.

It is unusable with qucsator. Does not imply that the subcircuit model
is wrong, and there are other simulators out there.  If the
Warning/Error is not good enough (I think it is), maybe a comment should
explain the issue, until the devices are implemented or changed...?
(the latter is a qucsator issue, not qucslib).

> In all component definitions: 
> <Component **BSP125**>
> ...
>   \<Model\>
>   .Def:NMOSFETs_**BSP125** _net1 _net3 _net2
> 
> the component name and the name in .Def (after underscore) are the
> same. Only in those erroneous the second name slightly differs and
> looks like typo. I wanted to ask the author Gunther Kraut, but his
> eMail \<gn.kraut@online.de\> is no more up to date.

OK i see. Was not obvious from the patch.

(There was something strange going on in how these subcircuit
definitions are used, but I'm afraid I do not remember the details).

Maybe the name after .Def is the one used in the netlist, ideally
unique, but otherwise arbitrary.  (Can you confirm this?). With this, I
think your change is an improvement. No need to ask the author in that
case.

thanks
felix

oxmon-2500:
>there are other simulators out there

I see, I will try this components with qucs-s

felix-salfelder:
On Mon, Mar 23, 2020 at 10:50:19AM -0700, Szymon Blachuta wrote:
> >there are other simulators out there
> I see, I will try this components with qucs-s

Qucs runs any (suitable) simulator you choose, (set the QUCSATOR
environment variable.) With gnucsator, you may choose from about half a
dozen Diode implementations, all with potentially different ranges (have
not checked). Add your own if needed.

More simulators or diodes may come or not. The point is, that you can't
possibly know if a modelname (+ parameters) is understood and supported
before you run the simulator. The current error handler implementation
reflects this. ("asking" the simulator without actually running it is
not implemented. that would be overkill, ymmv).

(qucs-s is a different project, don't know what they do, but they might
also import your patches in the end.)

Issue #979, opened 2020-03-01
transistor model problems


boomboompsh:
the BFR92A is an NPN transistor, not PNP. 
datasheet: [BFR92A](https://www.nxp.com/docs/en/data-sheet/BFR92A_N.pdf)
I don't understand the transistor models yet, so it could be as easy as switching it from "pnp" to "npn". The X5_500 example uses this part, so it needs to be changed to use the BFT92A, the pnp equivalent.

The MMBTH10 and MMBTH24 are not characterized above 1GHz on their respective datasheets, but the component library claims frequencies of 650GHz and 400GHz respectively. It's probably a typo and it was supposed to be MHz. I'm not sure exactly what the frequencies listed on the model description are supposed to mean.
[transistor library](https://github.com/Qucs/qucs/blob/0ef64c307d72fa9c3d8075d6a69b79146f060e40/qucs/qucs-lib/library/Transistors.lib)
[X5_500 schematic](https://github.com/Qucs/qucs/blob/0ef64c307d72fa9c3d8075d6a69b79146f060e40/qucs/examples/X5_500.sch)

boomboompsh:
The BUZ272 has an on resistance of 0.3 Ohm, not 0.3 mOhm as claimed in the component library.
[BUZ272 datasheet](http://pdf.datasheetcatalog.com/datasheet/infineon/1-buz272.pdf)

oxmon-2500:
Thank you for contribution, PR #996 issued

felix-salfelder:
closing, as this is covered by 996.

PR #996, opened 2020-03-30
Library typos corrections

oxmon-2500:
corrections:
BFR92A  pnp -> npn  https://www.nxp.com/docs/en/data-sheet/BFR92A_N.pdf
MMBTH10 650GHz -> 650MHz https://www.diodes.com/assets/Datasheets/ds31031.pdf
MMBTH24 400GHz -> 400MHz https://www.diodes.com/assets/Datasheets/ds31034.pdf
BUZ272  0.3 mOhm -> 0.3 Ohm http://pdf.datasheetcatalog.com/datasheet/infineon/1-buz272.pdf
Issue https://github.com/Qucs/qucs/issues/979 #979

### DIFF
--- a/qucs/qucs-lib/library/PMOSFETs.lib
+++ b/qucs/qucs-lib/library/PMOSFETs.lib
@@ -351,7 +351,7 @@
 <Component BUZ272>
   <Description>
    Enhancement p-channel power MOSFET
-   -100 V, -15 A, 0.3mOhm
+   -100 V, -15 A, 0.3 Ohm
    Manufacturer: Infineon Technologies
    added by Gunther Kraut <gn.kraut@online.de>
   </Description>
--- a/qucs/qucs-lib/library/Transistors.lib
+++ b/qucs/qucs-lib/library/Transistors.lib
@@ -2606,11 +2606,11 @@ Added by Frans Schreuder <fransschreuder@users.sourceforge.net>
 
 <Component BFR92A>
   <Description>
-    Si  pnp 300mW 20V 25mA 5GHz pkg:SOT-23 2,1,3
+    Si  npn 300mW 20V 25mA 5GHz pkg:SOT-23 2,1,3
     added by Frans Schreuder
   </Description>
   <Model>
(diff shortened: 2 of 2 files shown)


felix-salfelder:
On Sun, Mar 29, 2020 at 10:35:38PM -0700, Szymon Blachuta wrote:
> BFR92A  pnp -&gt; npn  https://www.nxp.com/docs/en/data-sheet/BFR92A_N.pdf

not sure about this one.

   <Description>
-    Si  pnp 300mW 20V 25mA 5GHz pkg:SOT-23 2,1,3
+    Si  npn 300mW 20V 25mA 5GHz pkg:SOT-23 2,1,3
     added by Frans Schreuder
   </Description>

but (3 lines further down, unchanged)

<_BJT BFR92A_ 1 0 0 8 -26 0 0 "pnp" [..] >
                              ^^^^

Is it possible that this is a BFR92A_P, while the pdf describes the _N variant?
In that case, maybe add BFR92A_N...

cheers

oxmon-2500:
>Is it possible that this is a BFR92A_P, while the pdf describes the _N variant?

I think the document is named unfortunately with the suffix "_N" , the component is named BFR92A without a suffix inside the document. A web search did not bring results with a mention of pnp variant.

>but (3 lines further down, unchanged)

oops! corrected

boomboompsh:
The X5_500 example uses the BFR92A and should probably be changed to use an equivalent pnp transistor. Unfortunately, the proper pnp equivalent, the BFT92A, causes an error in the transient simulation when I just drop it in as a replacement.

felix-salfelder:
On Mon, Mar 30, 2020 at 07:49:10PM -0700, boomboompsh wrote:
> The X5_500 example uses the BFR92A and should probably be changed to
> use an equivalent pnp transistor. Unfortunately, the proper pnp
> equivalent, the BFT92A, causes an error in the transient simulation
> when I just drop it in as a replacement.

ouch.

Perhaps we should keep the "wrong" BFR92A, with a comment.
then introduce a fixed BFR92A-1.

This is approximately what gEDA folks do. I cannot think of a better or
simpler way right now.

oxmon-2500:
>Perhaps we should keep the "wrong" BFR92A, with a comment.
then introduce a fixed BFR92A-1.

I think BF**R**92A is after correction (pnp->npn) OK. BF**T**92A is a pnp complement to BF**R**9A (as we sought after BF**R**9A_P). I tried modified X5_500 using BF**T**92A and it resulted in an endles loop (program hangs). This is another problem of qucsator using transient simulation.

felix-salfelder:
On Tue, Mar 31, 2020 at 05:12:37AM -0700, Szymon Blachuta wrote:
> >Perhaps we should keep the "wrong" BFR92A, with a comment.
> then introduce a fixed BFR92A-1.
> 
> I think BF**R**92A is after correction (pnp->npn) OK. BF**T**92A is a
> pnp complement to BF**R**9A (as we sought after BF**R**9A_P). I tried
> modified X5_500 using BF**T**92A and it resulted in an endles loop
> (program hangs). This is another problem of qucsator using transient
> simulation.

Thanks for pointing out T vs R.

We can now leave it there, breaking backwards compatibility (with a
remark in the release notes). Or we do the -1 thing.

I don't have a strong opinion on this particular one. But we might want
to find out how this was dealt with in the past, and how we'd like to
deal with it in the future.

PR #1007, opened 2020-05-08
Common lists of files for all build systems

felix-salfelder:
Plan: Just store lists of files and keep the build systems separate.
... Most build systems can read files.


felix-salfelder:
not going to happen in 0.0.20, done in refactor branch.

PR #1011, opened 2020-06-14
Fix qucs.desktop so that we can deploy AppImages

Rmano:
This fix is needed because otherways making AppImages you get this error:

```
[appimage/stdout] /home/romano/software/qucs/AppImage/qucs-qucs-0.0.20-rc2/build/AppDir/qucs.desktop: warning: key "Encoding" in group "Desktop Entry" is deprecated
[appimage/stdout] /home/romano/software/qucs/AppImage/qucs-qucs-0.0.20-rc2/build/AppDir/qucs.desktop: error: value "Education;Electronics" for string list key "Categories" in group "Desktop Entry" does not have a semicolon (';') as trailing character
```

...and if you notice, the other `.desktop` in the repo (contrib/ubuntu-debian/qucs/debian/qucs.desktop) seems correct.

### DIFF
--- a/qucs/qucs/qucs.desktop
+++ b/qucs/qucs/qucs.desktop
@@ -1,5 +1,4 @@
 [Desktop Entry]
-Encoding=UTF-8
 Name=Qucs
 Comment=A Universal Circuit Simulator
 Comment[es]=Un simulador universal de circuitos
@@ -13,5 +12,5 @@ Icon=qucs
 Terminal=false
 
 Type=Application
-Categories=Education;Electronics
+Categories=Education;Electronics;
 StartupNotify=false


Rmano:
Hi, I think this one is quite straightforward and should be merged in --- I am trying to create an automatic AppImage release for those systems that have no QT4 and this is necessary...

thomaslepoix (approved):

felix-salfelder:
fine with me.

Rmano:
The other possibility is to use directly the `contrib/ubuntu-debian/qucs/debian/qucs.desktop`, which has a lot more translations. Whatever works better...

felix-salfelder:
whatever works. I am just sifting through the patches...

ideally there should be only one qucs.desktop.

Rmano:
I didn't try with the other one, but I can give it a spin.

Rmano:
...compiling now on https://github.com/Rmano/qucsAppImagesBuild/actions

Rmano:
No, the other `.desktop` does not work because now it doesn't find a correct icon. So I think that it's better to merge this one for now, we can sort out the "double `.desktop` issue" later.

PR #1013, opened 2020-08-18
docs: fixed non UTF-8 files

yarda:
Modern systems use UTF-8 by default and compilation (docs generation with latex) fails on non UTF-8 characters.


felix-salfelder:
thank you. i tried something different, as qucs is also being used on non-modern systems.

does f8a2e185525f work for you?

yarda:
Yes, it seems it works. I switched to it in Fedora, thanks.

Strange, that you have to support non UTF-8 capable systems - fortunately, I haven't seen any for decades :)

felix-salfelder:
merged. thanks for testing.

PR #1021, opened 2020-12-06
rename qucs folder to src to not have that much qucs folders

Murmele:


felix-salfelder:
done. i did not delete any files in qucs/python/test though.

PR #1022, opened 2020-12-09
move filenames to MakeList files

Murmele:
@felix-salfelder can you test if it is still working for you?


felix-salfelder:
On Wed, Dec 09, 2020 at 09:02:35AM -0800, Murmele wrote:
> @felix-salfelder can you test if it is still working for you?
> You can view, comment on, or merge this pull request online at:

there is a rule of thumb: untested code never works.

anyway, I can see we are making some progress.

ideally, the MakeLists all use the same variable names, so the directory
determines the name of the plugin, and the build system could use the
same boilerplate for all of them (this is on my list, don't worry. i'll
try not to interfere with your stuff).

PR #1025, opened 2020-12-28
Cmake and Flatpak packages

Murmele:
Add CMake and flatpak build support

- Until now only autotools was supported. Get CMake support which was already once for an older Qucs version available work again with Qt5.
- Add flatpak support for qucs, qucsator is not shown yet, has to be checked out why
- Initialize all pointers with a nullptr, so they have a defined state


felix-salfelder:
On Mon, Dec 28, 2020 at 02:18:53AM -0800, Murmele wrote:
> CMake and Flatpak package.

please rebase to the newest revision (-20?), and squash into one commit.
avoid white space and format changes.

Murmele:
I merged, this is the reason, why it looks like I'm behind. Why do I have to squash into one commit? Then the history is gone

felix-salfelder:
On Mon, Dec 28, 2020 at 04:14:52AM -0800, Murmele wrote:
> I merged, this is the reason, why it looks like I'm behind. Why do I
> have to squash into one commit? Then the history is gone

when you squash, you can concatenate the commit messages into a bullet
point list. kick out the unneeded ones. also, have a look at [1].

[1] https://chris.beams.io/posts/git-commit/

Murmele:
I create a new one with only one commit

PR #1028, opened 2020-12-29
Instead of fixed numbers, use an enum. So the code gets more readable

Murmele:


felix-salfelder:
I think I commented on the preprocessor macro somewhere turn it into a
static function.

this

m_model->item(FileTypes_T::Datasets, 0)->appendRow(columnData);

is known as the "train wreck" anti pattern. pointers must be checked
before dereferencing (e.g. in function).

Murmele:
Ok good to know I will change it

Murmele:
@felix-salfelder now I'm checking if the item is available or not.

PR #1029, opened 2020-12-29
Add cmake support

Murmele:


felix-salfelder:
On Tue, Dec 29, 2020 at 09:55:37AM -0800, Murmele wrote:
> -- File Changes --
> 
>     A cmake/main/compileAndInstallPlugins.cmake (55)

should be cmake/plugins.cmake (or so). not part of main.

>     M legacy/CMakeLists.txt (24)
>     M legacy/components/CMakeLists.txt (43)
>     A legacy/libfiles/CMakeLists.txt (27)
>     M legacy/paintings/CMakeLists.txt (40)
>     A legacy/qucsator/CMakeLists.txt (29)

All cmake files must move to cmake. I don't want to break stuff by
cleaning up eventually.

Murmele:
> On Tue, Dec 29, 2020 at 09:55:37AM -0800, Murmele wrote: -- File Changes -- A cmake/main/compileAndInstallPlugins.cmake (55)
> should be cmake/plugins.cmake (or so). not part of main.
> M legacy/CMakeLists.txt (24) M legacy/components/CMakeLists.txt (43) A legacy/libfiles/CMakeLists.txt (27) M legacy/paintings/CMakeLists.txt (40) A legacy/qucsator/CMakeLists.txt (29)
> All cmake files must move to cmake. I don't want to break stuff by cleaning up eventually.

Ok I will move them, but what is the reason for? It is just one more file in each folder? As you can se in the image in the file tree in QtCreator there is this unneccesary item <OtherLocations> this is, because the files are not located in a subfolder of the cmakelists file (add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/../src src))
Another benefit is that when a plugin will be removed, only the folder and the add_subdirectory instruction must be removed and not a CMakeLists.txt file somewhere

qucsCmake

PR #1034, opened 2021-01-08
do not use else, because when all documents are closed, a new one sho…

Murmele:
A new document should be created also when all files are closed successfully

### DIFF
--- a/src/qucs.cpp
+++ b/src/qucs.cpp
@@ -1115,12 +1115,14 @@ void QucsApp::openProject(const QString& Path)
     return; // BUG
   }
 
-  // try to close files and ask for saving them
-  if(closeAllFiles()){ untested();
-    // "all files were succesfully closed"
-  }else{ untested();
-    // not sure why we need an empty schematic
-    DocumentTab->createEmptySchematic("");
+  if(!closeAllFiles()){
+    untested();
+    // close files and ask for saving them
+  }
+    // As soon as all files are closed, open a new file.
+  // TODO: maybe it is better to open no project or?
+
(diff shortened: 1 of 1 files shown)


felix-salfelder:
On Fri, Jan 08, 2021 at 08:04:46AM -0800, Murmele wrote:
> A new document should be created also when all files are closed successfully
> You can view, comment on, or merge this pull request online at:

thanks. can you say *why* it should be created? is this
opinion/taste or is it because 0.0.20 does it?

what is sensible? how do other programs do it? e.g.

firefox: close all tabs -> also closes main app
gimp: close all images -> nothing else happens

which one makes more sense? Why do we have to care at this stage?

Murmele:
It is when a project will be opened. The problem is, if the file gets not created and you drop a component a segmentation fault occurs.

felix-salfelder:
On Fri, Jan 08, 2021 at 10:16:37AM -0800, Murmele wrote:
> It is when a project will be opened. The problem is, if the file gets
> not created and you drop a component a segmentation fault occurs.

drop as in drag&drop from the left menu? does this work for you at all?
I think just selecting a symbol and clicking somewhere on the scene is
working much better (how does 0.0.20 do it?).

do we want/need drag&drop? some code needs debugging or removal, either
way.

Murmele:
Both is working for me. Hmm for me also selecting is also fine. But I found it just in a second step. I will show a video

Murmele:
The ground was dragged, the opamp was selected without dragging
https://user-images.githubusercontent.com/10099533/104089217-57a0c900-526d-11eb-8b02-c312e7270979.mp4

PR #1038, opened 2021-02-25
[WIP] Trying to build for windows

Patarimi:
I try to build qucs for windows using msys. The list below summaries what works.
- [x] autotools
- [x] man
- [x] src
- [x] qucs-lib
- [x] tools
- [x] translations
- [x] examples
- [x] include
- [ ] main
build successful. However, in debug mode, it crashes with the following message
```
incomplete:../../src/settings.cpp:14:tQucsSettings
error 3 bad internal command: attach
error 3 bad internal command: attach
error 3 bad internal command: attach
error 3 bad internal command: attach
error 3 bad internal command: attach
error 3 bad internal command: loadlegacylib
error 3 bad internal command: attach
error 3 bad internal command: attach
error 5 no language qucsator
```
- [x] cli
- [x] contrib
note : need to include qcustomplot.h et qcustomplot.cpp in msys64\mingw64\include (to be added in the appveyor.yml script ?)
- [x] tests
- [x] legacy
- [x] plugins

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,68 @@
+image: Visual Studio 2019
+environment:
+  matrix:
+    - MSYSTEM : MINGW64
+      MBITS: 64
+      MARCH: x86_64
+
+skip_commits:
+  files:
+    - .travis.yml
+    - '*.md'
+
+cache:
+  # cache pacman packages
+  # max cache size is 1 GB, which should be enough
+  # but in case we could cache only mingw-w64-*-qt4-*
+  - C:\msys64\var\cache\pacman\pkg\
+
+install:
+  #Setting path to find pacman
+  - set "PATH=C:\msys64\usr\bin;%PATH%"
+  # Take a look at the environment
+  - set
+  #correction to pgk-config path (needed for Qt5)
+  - set QT5_PATH=/mingw%MBITS%/lib/pkgconfig
+
+  - bash -lc ""
+  - bash -lc "pacman --version"
+  - bash -lc "pacman -Q"
+  # Switch from SF to msys2.org (default, much faster)
+  - bash -lc "pacman --noconfirm --needed --sync pacman-mirrors"
+  - bash -lc "pacman --noconfirm --needed -S autoconf automake bison flex"
+  - bash -lc "pacman --noconfirm --needed -S mingw-w64-$MARCH-qt5"
+  
+  #- bash -lc "cd /var/cache/pacman/pkg && wget -nc http://repo.msys2.org/mingw/$MARCH/mingw-w64-$MARCH-qt4-4.8.7-4-any.pkg.tar.xz && pacman --noconfirm --needed -U mingw-w64-$MARCH-qt4-4.8.7-4-any.pkg.tar.xz"
+
+  # Set compiler (64bit)
+  - set "CC=/c/msys64/usr/bin/gcc.exe"
+  - set "CXX=/c/msys64/usr/bin/g++.exe"
+  - set "CXXFLAGS=-std=gnu++14 -pipe -fno-check-new"
+
+  - bash -lc "awk --version | head -n1"
+
+build_script:
+  ## Notes
+  #  * The "exec 0</dev/null" opens a dummy file descriptor to fix error: ./configure: line 560: 0: Bad file descriptor
+
+  ## Build Qucs-GUI applications and install
+  - bash -lc "exec 0</dev/null && cd $APPVEYOR_BUILD_FOLDER && ./bootstrap"
+  - bash -lc "cd $APPVEYOR_BUILD_FOLDER && mkdir build"
+  - bash -lc "exec 0</dev/null && cd $APPVEYOR_BUILD_FOLDER/build && export PKG_CONFIG_PATH+=:$QT5_PATH && ../configure --prefix=/c/qucs-win$MBITS --disable-dependency-tracking"
+  - bash -lc "exec 0</dev/null && cd $APPVEYOR_BUILD_FOLDER/build && make -e"
+  - bash -lc "exec 0</dev/null && cd /c/qucs-win$MBITS/bin && ./qucs -v"
+
+  ## Check
+  - bash -lc "exec 0</dev/null && cd $APPVEYOR_BUILD_FOLDER/qucs && make check || :"
+
+after_build:
+   - 7z a %APPVEYOR_BUILD_FOLDER%\qucs-win%MBITS%.zip %QUCSDIR%
+
+artifacts:
+   # variables here must use the PowerShell syntax
+   - path: qucs-win$(MBITS).zip
+     name: qucs-win$(MBITS).zip
+
+on_failure:
+  - bash -lc "exec 0</dev/null && cat $APPVEYOR_BUILD_FOLDER/config.log"
+  - ps: Get-ChildItem .\*.log -Recurse | % { Push-AppveyorArtifact $_.FullName -FileName $_.Name }
--- a/autotools/components.am
+++ b/autotools/components.am
@@ -28,7 +28,7 @@ include ../../autotools/common.mk
 
 inst_LTLIBRARIES = components.la
 
-components_la_LDFLAGS = -shared -module -avoid-version
+components_la_LDFLAGS = -no-undefined -shared -module -avoid-version
 
 components_la_SOURCES = $(DEFAULT_COMPONENTS)
 components_la_LIBADD = $(QT_LIBS)
--- a/autotools/hidpi.am
+++ b/autotools/hidpi.am
@@ -28,8 +28,10 @@ AM_DEFAULT_SOURCE_EXT = .cpp
 
 inst_LTLIBRARIES = hidpi.la
 
-hidpi_la_LDFLAGS = -shared -module -avoid-version
+hidpi_la_LDFLAGS = -no-undefined -shared -module -avoid-version
 hidpi_la_CXXFLAGS = ${AM_CXXFLAGS} -fPIC
 hidpi_la_CPPFLAGS = ${AM_CPPFLAGS} ${QT_CFLAGS}
+
+hidpi_la_LIBADD = $(X11_LIBS) $(QT_LIBS)
 
 CLEANFILES = *~ Makefile.in *.rej *.orig
--- a/autotools/vadim.am
+++ b/autotools/vadim.am
@@ -24,9 +24,11 @@ include ../../autotools/common.mk
 
 inst_LTLIBRARIES = vadim.la
 
-vadim_la_LDFLAGS = -shared -module -avoid-version
+vadim_la_LDFLAGS = -no-undefined -shared -module -avoid-version
 vadim_la_SOURCES = \
 	DIODE_SPICE.cpp
+
+vadim_la_LIBADD = $(X11_LIBS) $(QT_LIBS)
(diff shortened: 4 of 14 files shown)


felix-salfelder:
Thanks for your comments and patches.

On Thu, Feb 25, 2021 at 12:34:58AM -0800, Patarimi wrote:
> I try to build qucs for windows using msys. The list below summaries what works.
> - [x] autotools
> - [x] man
> - [x] src
> - [x] qucs-lib
> - [ ] tools
> _lst_ variables detected as *int* by compiler. Rename lst in list and it builds.

Does no harm. merged.

> - [x] translations
> - [x] examples
> - [ ] include
> `../../include/output.h:153:3: error: &#39;axis&#39; declared as a &#39;virtual&#39; field`
> function DataX from include/output.h need to be removed.

I think on your platform "axis" is a macro (?). The line is staged for
removal anyway but removing it breaks compilation in another place. I
need to check.

> - [ ] main
> build successful, in debug crash with the following message `assertion failed` in main/main.cpp: line 190
> the line.png file is not find.

The line.png line is a sanity check, I don't know exactly where QFile is
looking for things, but you won't see any icons if this does not work. I
wouldn't mind to drop the "resource files" for some more sensible
approach...

> - [ ] cli
> `../../cli/main.cpp:69:3: error: &#39;sigjmp_buf&#39; does not name a type; did you mean &#39;jmp_buf&#39;?`

This should be handled by platform.h. maybe I broke it. maybe it has to
do with SIGSETJMP_IS_BROKEN. see line 138. try -DSIGSETJMP_IS_BROKEN.

> - [ ] contrib
> note : need to include qcustomplot.h et qcustomplot.cpp in msys64\mingw64\include (to be added in the appveyor.yml script ?)

This plugin is optional. install qcustomplot headers, or just skip it.
(there will be changes to how plugins will be built, better dont waste
your time on configure.ac)

> `/bin/sh ../../libtool  --tag=CXX   --mode=link g++ -g -O2 -g -O0 -pipe -fno-check-new -W -Wall -std=c++14 -O0 -fPIC -fPIC -g -O2 -g -O0 -pipe -fno-check-new -W -Wall -std=c++14 -O0 -shared -module -avoid-version -mwindows -o vadim.la -rpath /c/qucs-win64/lib/qucs//contrib/vadim/.. vadim_la-DIODE_SPICE.lo
> libtool:   error: can&#39;t build x86_64-w64-mingw32 shared library unless -no-undefined is specified`

Do you see this for all plugins? you need to link them against libqucs
(on some non-POSIX platforms.). plugins should be linked against
QUCS_LIBS (e.g. through qucs-conf, but currently autotools...), it needs
to be set accordingly. Perhaps you also need to link qt libraries,
sometimes but not sure.

> - [x] tests
> - [ ] legacy
> `test -L _ || cp -pR .libs/action.so _
> cp: cannot stat &#39;.libs/action.so&#39;: No such file or directory`

strange. will have a look.

Patarimi:
Thanks a lot for your responses.  It was very helpfull. Every things compile now !

> - [ ] main build successful, in debug crash with the following message `assertion failed` in main/main.cpp: line 190 the line.png file is not find.
> The line.png line is a sanity check, I don't know exactly where QFile is looking for things, but you won't see any icons if this does not work. I wouldn't mind to drop the "resource files" for some more sensible approach...

The program still crash after anyway, but all the .png are missing. Maybe there is a problem in the src folder ?

> - [x] cli `../../cli/main.cpp:69:3: error: &#39;sigjmp_buf&#39; does not name a type; did you mean &#39;jmp_buf&#39;?`
> This should be handled by platform.h. maybe I broke it. maybe it has to do with SIGSETJMP_IS_BROKEN. see line 138. try -DSIGSETJMP_IS_BROKEN.

Fix with -DSIGSETJMP_IS_BROKEN.

> - [ ] contrib note : need to include qcustomplot.h et qcustomplot.cpp in msys64\mingw64\include (to be added in the appveyor.yml script ?)
> This plugin is optional. install qcustomplot headers, or just skip it. (there will be changes to how plugins will be built, better dont waste your time on configure.ac)
> `/bin/sh ../../libtool --tag=CXX --mode=link g++ -g -O2 -g -O0 -pipe -fno-check-new -W -Wall -std=c++14 -O0 -fPIC -fPIC -g -O2 -g -O0 -pipe -fno-check-new -W -Wall -std=c++14 -O0 -shared -module -avoid-version -mwindows -o vadim.la -rpath /c/qucs-win64/lib/qucs//contrib/vadim/.. vadim_la-DIODE_SPICE.lo libtool: error: can&#39;t build x86_64-w64-mingw32 shared library unless -no-undefined is specified`
> Do you see this for all plugins? you need to link them against libqucs (on some non-POSIX platforms.). plugins should be linked against QUCS_LIBS (e.g. through qucs-conf, but currently autotools...), it needs to be set accordingly. Perhaps you also need to link qt libraries, sometimes but not sure.

I add -no-undefined in LD_FLAGS variable and it fix it. It seems to be mandatory when the -shared flag is used.

> - [x] tests - [x] legacy `test -L _ || cp -pR .libs/action.so _ cp: cannot stat &#39;.libs/action.so&#39;: No such file or directory`
> strange. will have a look.

This error disappeared.

felix-salfelder:
On Fri, Feb 26, 2021 at 12:14:35AM -0800, Patarimi wrote:
> Thanks a lot for your responses.  It was very helpfull. Every things compile now !

Thanks for your feedback. Please rebase now and then -- I tend to fix
things one by one, sometimes in a different way.

> The program still crash after anyway, but all the .png are missing. Maybe there is a problem in the src folder ?

I simply don't know. It does not happen here.

> Fix with -DSIGSETJMP_IS_BROKEN.

see platform.h

> I add -no-undefined in LD_FLAGS variable and it fix it. It seems to be mandatory when the -shared flag is used.

It must not be hard coded. Find a way to make it optional. Ideas

- look at common.mk
- look at qucs-conf
- introduce EXTRA_PLUGIN_LD_FLAGS (?) and build with
  $ make EXTRA_PLUGIN_LD_FLAGS=-no-undefined

Same issue with additional libraries. This might as well require changes
to MakeLists... only MakeList contents and qucs-conf will inform
compilation later on. (Will get back to this.)

Patarimi:
I think I fix the issue with the missing file (see [6212ba6](https://github.com/Qucs/qucs/pull/1038/commits/6212ba676a3303d1edb6f20654de6a01797d05f8)).
For the lib_LD_FLAGS, want do you mean by hard-coded ? For example, is the fix in [0ae75e2](https://github.com/Qucs/qucs/pull/1038/commits/0ae75e2272cc2928bd76de2748e2c39131a9a181) correct ?
The program still don't start perhaps because the qucsator program is not accessible.

felix-salfelder:
On Wed, Mar 03, 2021 at 12:06:09AM -0800, Patarimi wrote:
> I think I fix the issue with the missing file (see
> [6212ba6](https://github.com/Qucs/qucs/pull/1038/commits/6212ba676a3303d1edb6f20654de6a01797d05f8)).

Good, this is where it is looking for the bitmaps. It solves the issue
half-way... the programs are supposed to run without the need for
installation (alias "make install".) In particular the tests will still
fail after this "fix". There is more to it.

> For the lib_LD_FLAGS, want do you mean by hard-coded ? For example, is
> the fix in
> [0ae75e2](https://github.com/Qucs/qucs/pull/1038/commits/0ae75e2272cc2928bd76de2748e2c39131a9a181)
> correct ?

Never thought about it, but -no-undefined is probably wrong in general.
What does it do exactly?

> hidpi_la_LIBADD = $(X11_LIBS) $(QT_LIBS)

This may be needed on non-POSIX, but it is wrong in general and it has
side effects.

> The program still don't start perhaps because the qucsator program is not accessible.

Do you have an error message? qucsator is not needed for sure.

Patarimi:
> Never thought about it, but -no-undefined is probably wrong in general. What does it do exactly?

According to the [documentation ](https://www.gnu.org/software/libtool/manual/libtool.html):

> _-no-undefined_
Declare that output-file does not depend on any libraries other than the ones listed on the command line, i.e., after linking, it will not have unresolved symbols. Some platforms require all symbols in shared libraries to be resolved at library creation (see Inter-library dependencies), and using this parameter allows libtool to assume that this will not happen.

This is probably specific to windows. This may also explain why the declaration below is mandatory.

> hidpi_la_LIBADD = $(X11_LIBS) $(QT_LIBS)
> This may be needed on non-POSIX, but it is wrong in general and it has side effects.


> The program still don't start perhaps because the qucsator program is not accessible.
> Do you have an error message? qucsator is not needed for sure.

When run from a terminal, it output the following:
```
incomplete:../../src/settings.cpp:14:tQucsSettings
error 3 bad internal command: attach
error 3 bad internal command: attach
error 3 bad internal command: attach
error 3 bad internal command: attach
error 3 bad internal command: attach
error 3 bad internal command: loadlegacylib
error 3 bad internal command: attach
error 3 bad internal command: attach
error 5 no language qucsator
```

felix-salfelder:
On Wed, Mar 03, 2021 at 05:41:55AM -0800, Patarimi wrote:
> When run from a terminal, it output the following:
> ```
> incomplete:../../src/settings.cpp:14:tQucsSettings
> error 3 bad internal command: attach

"attach" is in src/attach.cc it can't be not there...  The dispatcher
stuff was added years ago, when CI was still working. I have never seen
this on non-POSIX since, something stupid must be going on...

If you can't figure it out, please send the configure & build logs and
the full output, just in case. cross fingers.

thanks

Patarimi:
I have not found the solution yet. I'll try to look for it tomorrow.
Here are the config.log and the make output:
[make.log](https://github.com/Qucs/qucs/files/6077546/make.log)
[config.log](https://github.com/Qucs/qucs/files/6077548/config.log)

felix-salfelder:
On Wed, Mar 03, 2021 at 08:37:57AM -0800, Patarimi wrote:
> I have not found the solution yet. I'll try to look for it tomorrow.
> Here are the config.log and the make output:
> [make.log](https://github.com/Qucs/qucs/files/6077546/make.log)
> [config.log](https://github.com/Qucs/qucs/files/6077548/config.log)

this is what I found.

> ../../include/e_base.h:29: warning: "INTERFACE" redefined
>   29 | #define INTERFACE

All such definitions must be in platform.h (this is for me to fix.) I
don't really know what it does (not needed on POSIX), but it might be
worth checking in the long run.

redefinition of FALSE, also for me to fix.

> /bin/sh ../libtool  --tag=CXX   --mode=link g++ -g -O2 -pipe -fno-check-new -std=c++14 -fPIC -g -O2 -pipe -fno-check-new -std=c++14 -mwindows -o libqucs.la [..]
> [..] libtool: warning: undefined symbols not allowed in x86_64-w64-mingw32 shared libraries;
> [..] building static only [..]

I don't know if static builds will work. They introduce problems, e.g.
[1]. I suggest to dig into the the libtool warning. static builds are
for platforms without a dynamic loader (rare).

> libtool:   error: can't build x86_64-w64-mingw32 shared library unless -no-undefined is specified make[3]:
> *** [Makefile:501: hidpi.la] Error 1

hidpi.la is not mission critical, still you need to extend LDFLAGS and
LDADD in a few more more places. (The build is not complete, so
something else might be missing.)

[1] https://stackoverflow.com/questions/1202494/why-doesnt-attribute-constructor-work-in-a-static-library

PR #1040, opened 2021-03-15
Modular fix cmake

guitorri:
Qucs build and runs. Tested both cmake and autotools.
I am afraid the issues I had were due to cmake and its automoc feature. 
Somehow it includes all the code to be moc'ed. This created collision of definitions (F, p1,...).
I guess all the code end up on the same translation unit.
Taking a step back the .cc headers should have worked. The .hqt don't work with cmake. I picked .hxx for a change...
Up to you if you want to revert.


felix-salfelder:
On Mon, Mar 15, 2021 at 03:23:24PM -0700, Guilherme Brondani Torri wrote:
> Taking a step back the .cc headers should have worked. The .hqt don&#39;t work with cmake. I picked .hxx for a change...
> Up to you if you want to revert.

Thanks for figuring out the compilation units with cmake. It seems we
are getting there.

- .cc files do not work with moc. It has to be .h*. Symptoms are it wont
  print '#include "file.cc"'. Regardless of "autoinclude".
- QT_PLUGIN_SRC = ..cc was a temporary hack with side effects. Must
  always list the source files.
- autotools treats headers differently. a .h, .hxx (etc) file
  listed in *_SOURCE will be silently dropped.

I'd really like to stick to .hqt files now if possible, as these align
with the plan, and common use of file extensions. Plugins will be built
through attach.cc (at least where this is desired). For this, extensions
should map to rules, with little magic. I.e. "attach my_stuff.hqt" will
run moc, "attach other_stuff.cc" must not. Will have to rely on
extensions anyway, "attach more.py" etc.

With Make, .hqt -> .cc -> .so chains are trivial. Do you know why
QT5_WRAP_CPP does not like .hqt? I would hope this can be fixed or
implemented some other way with cmake (create symlinks?).

felix-salfelder:
cmake does not seem to mind .hqt. i picked the two other commits. This leaves minor issues only. \o/

PR #1069, opened 2023-01-27
Fix build with GCC 13

yselkowitz:
`uintptr_t` requires an explicit include with GCC 13, as described in https://gcc.gnu.org/gcc-13/porting_to.html

Signed-off-by: Yaakov Selkowitz <yselkowi@redhat.com>


felix-salfelder:
Thanks, applied to develop_filtered. c.f. https://aur.archlinux.org/packages/qucsator#comment-889978 for context.

PR #1092, opened 2024-09-24
Fixes to allow compilation on Windows using MSYS2 MINGW64 toolchain

EmilKleijn:
This patch fixes the following issues:

- git raising an error on 1u not being an integer during bootstrap / autoconf
- Name clash of lst2 and lst3 with defines in dlgs.h
- Name clash of Arc with function definition in wingdi.h
- toAscii deprecation, replaced with toLatin1
- Missing QtScript in configure.ac caused compilation errors

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -250,7 +250,7 @@ if test "x$PKG_CONFIG" != "x"; then
 
       PKG_CHECK_MODULES([QTest], [${qt_name}Test])
 
-      PKG_CHECK_MODULES([QT], [${qt_name}Core ${qt_name}Gui ${qt_name}Svg ${qt_name}Xml], [
+      PKG_CHECK_MODULES([QT], [${qt_name}Core ${qt_name}Gui ${qt_name}Script  ${qt_name}Svg ${qt_name}Xml], [
         QT_CFLAGS="$QT_CFLAGS $QT_DEF -DQT_DEPRECATED_WARNINGS -D_REENTRANT"
         case $host_os in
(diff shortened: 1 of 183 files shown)


felix-salfelder:
On Tue, Sep 24, 2024 at 12:01:09PM -0700, EmilKleijn wrote:
> This patch fixes the following issues:

> - git raising an error on 1u not being an integer during bootstrap / autoconf

Interesting. Don't know where the u came from. Works without for me.

> - Name clash of lst2 and lst3 with defines in dlgs.h
> - Name clash of Arc with function definition in wingdi.h
> - toAscii deprecation, replaced with toLatin1

I agree, except for the "using namespace" in header files. It makes code
even harder to read. Please consider removing it.

I don't think it's as much of a problem in .cpp files, but even there, I
would prefer to pick one at a time.

using qucs::Arc; [..]

Not everything qucs has should show up everywhere. Fully qualified names
are also an option.

> - Missing QtScript in configure.ac caused compilation errors

You are right. I thought it's no longer needed with Qt6, but theres more
work to do.

Thanks.

EmilKleijn:
I'll remove the using namespace qucs from the header files and use the fully qualified name everywhere. It will take some time to refactor though, as there are many files to change.

felix-salfelder:
On Wed, Sep 25, 2024 at 04:40:36AM -0700, EmilKleijn wrote:
> I'll remove the using namespace qucs from the header files and use the
> fully qualified name everywhere. It will take some time to refactor
> though, as there are many files to change.

Thanks!

PR #1098, opened 2025-01-07
Fixing broken symbol issue

leviathanch:
This solves the issue with broken symbols after the last mega patch

    Fixing broken symbol issue
    
    This patch ended being the same Felix also did to fix
    building qrc with main.cpp using automake but it also
    fixes the build using CMake


felix-salfelder:
how does it relate to my earlier fix, 274123284dea? please rebase.

